PR #13, opened 2021-09-14
Rewrite InputFileParameter to FileSystemPathParameter

imikejackson:
This allow use of InputFile, InputPath, OutputFile, OutputPath


src/complex/Core/Parameters/FileSystemPathParameter.cpp
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  if(!fs::is_regular_file(path))
+  {
+    return {nonstd::make_unexpected(std::vector<Error>{{-1, fmt::format("Path \"{}\" is not a file", path.string())}})};
+  }
+  return {};
+}
+
+Result<> _validateInputPath(const FileSystemPathParameter::ValueType& value)
```
JDuffeyBQ, line 97: I'd move these functions into an anonymous namespace and drop the leading underscore from the name.

JDuffeyBQ (changes requested):
Rather than changing the input type, I'd say we make the path type part of the parameter definition itself.

imikejackson:
@JDuffeyBQ @mmarineBlueQuartz I don't know why the test is failing. There is a runtime issue with a std::any cast. Could one of you look at it?

JDuffeyBQ:
https://github.com/BlueQuartzSoftware/complex/blob/bef6f55cb10f92818cd59430ee48f938a4db3839/src/complex/Core/Filters/ImportTextFilter.cpp#L157

The value type you have for `FileSystemPathParameter` is a struct that contains a `std::filesystem::path`, but here it's cast to the path not the struct. I thought we were moving the path type into the parameter and then the value type would just be a `std::filesystem::path`.

imikejackson:
That isn't going to work. The 2 values (Path and Type) need to stay together so that the actual filter in it's execution knows what the intent is? Or I guess I could store a local value in the filter itself stating what kind of path (file/folder, input/output) but other filters might not have that luxury of doing that..

src/complex/Core/Parameters/FileSystemPathParameter.cpp
imikejackson: Done

JDuffeyBQ:
I meant storing the type in the parameter. So when the user constructs the parameter in the filter they'll specify whether it's for a file or directory. Then when deciding what to do with the value, the parameter will check what type it is using `std::filesystem` and fail if it's not the right type.

imikejackson:
But what happens if in the edge case where the Parameter does not do enough checks for the filter and the filter itself needs to know the type? Then the filter will need to store the type and now the type is in 2 locations.

JDuffeyBQ:
The filter creates the parameter so it already knows what type it should be.

imikejackson:
It typically creates it inline in the "setupFilterParameters()" method and does not store it in the Filter itself.

JDuffeyBQ (approved):

PR #14, opened 2021-09-16
[DRAFT] Add parameter for a long multi-line string data.

imikejackson:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ set(CoreParameters
   NumericTypeParameter
   StringParameter
   VectorParameter
+  ParagraphParameter
 )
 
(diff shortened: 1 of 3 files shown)


JDuffeyBQ (commented):
Does this parameter place any restrictions on the input text? If not then maybe it would make sense to add an option to StringParameter instead that controls whether the input is expected to be multi-line or not. Especially if this is only relevant for displaying in the GUI.

imikejackson:
There are not any restrictions on the input. An option to the StringParameter to make it multi-line would be an option. While the Paragraph is most meant for a UI to display, having it in a filter makes it available in a pipeline file which could be read and displayed by almost anything.

PR #18, opened 2021-09-23
[WIP] Refactor DatArray and DataStore to use std::vector<size_t> for Tuple and Comp dimensions

imikejackson:
Removed TupleCount and TupleSize
Use std::vector<size_t> for the dimensions of Tuple and Components
Updated all uses of API
Fixed reader/writing DataStructure to HDF5 files
Fixed writing of ImageGeom to HDF5 file


mmarineBlueQuartz (approved):

mmarineBlueQuartz (changes requested):
If implementation constants are only used in the source file, they do not belong in the header file. 

This happened a number of times in the data structure classes where calling Find All References points purely to the cpp file and a single reference in the header where the constant is declared. In this situation, if you really do want to put a single class's implementation constants in the header, mark them extern instead of inline const. The actual value does not matter to other classes, and changing the value should not force a recompile for every source file that directly or indirectly includes the header.

src/complex/Utilities/Parsing/HDF5/H5.hpp
```diff
@@ -12,8 +26,20 @@ using IdType = int64_t;
...
+static constexpr StringLiteral k_ComponentShapeTag = "ComponentShape";
+static constexpr StringLiteral k_DataObjectIdTag = "DataObjectId";
+static constexpr StringLiteral k_DataArrayTag = "DataArray";
+
+static constexpr StringLiteral k_ObjectTypeTag = "ObjectType";
+static constexpr StringLiteral k_DataStructureTag = "DataStructure";
+
+static constexpr StringLiteral k_VertexListIdTag = "VertexListId";
+static constexpr StringLiteral k_SharedVertexListTag = "SharedVertexList";
+static constexpr StringLiteral k_VertexSizesIdTag = "VertexSizesId";
```
JDuffeyBQ, line 42: These should be `inline` not `static`.

Also maybe these should be in their own HDF5 constants header.
```suggestion
inline constexpr StringLiteral k_DataTypeTag = "DataType";

inline constexpr StringLiteral k_DataStoreTag = "DataStore";
inline constexpr StringLiteral k_TupleShapeTag = "TupleShape";
inline constexpr StringLiteral k_ComponentShapeTag = "ComponentShape";
inline constexpr StringLiteral k_DataObjectIdTag = "DataObjectId";
inline constexpr StringLiteral k_DataArrayTag = "DataArray";

inline constexpr StringLiteral k_ObjectTypeTag = "ObjectType";
inline constexpr StringLiteral k_DataStructureTag = "DataStructure";

inline constexpr StringLiteral k_VertexListIdTag = "VertexListId";
inline constexpr StringLiteral k_SharedVertexListTag = "SharedVertexList";
inline constexpr StringLiteral k_VertexSizesIdTag = "VertexSizesId";
```

src/complex/Utilities/Parsing/HDF5/H5.hpp
```diff
@@ -1,8 +1,22 @@
...
+#include <hdf5.h>
+
+#ifdef H5Support_USE_MUTEX
+#include <mutex>
+#define H5SUPPORT_MUTEX_LOCK()                                                                                                                                                                         \
+  std::mutex mutex;                                                                                                                                                                                    \
+  std::lock_guard<std::mutex> lock(mutex);
+#else
+#define H5SUPPORT_MUTEX_LOCK()
+#endif
```
JDuffeyBQ, line 16: Do we want HDF5 headers included here? I believe this file is analogous to the one in H5Support that we made to avoid recompiling everything when we made small HDF5 changes.

JDuffeyBQ (changes requested):
Aside from the 2 comments, I think it's ok to merge once the conflicts are sorted out.

src/complex/Utilities/Parsing/HDF5/H5.hpp
imikejackson: Yes. We need the HDF5 headers in here because we are setting an HDF5 Macro to a specific value. If you get the headers out of order compilation fails in with very strange errors. By forcing developers to just #include this file we ensure that we get it correct.

Thinking more through the problem we might want to actually have the HDF5 macro that defines that we are using the 1.10 API as an actual compilation argument -D than sitting in a header file. This will guarantee the problem expansion of macros during compilation.

mmarineBlueQuartz: It should be pointed out that this macro does not do what I believe you think it does. If this is added to a function, each call gets its own mutex, making it impossible for two calls to the same function to lock the same mutex.

No actual HDF5 code is supposed to be in this file. Otherwise, you might as well delete H5.hpp and merge it all into H5Support.hpp. If you need HDF5 includes, include H5Support. If all you need are aliases, include H5.hpp.

And if you want this macro to be capable of locking across calls, change it to `static std::mutex mutex;` and it will lock when the same function is called. Move `std::mutex mutex` out of `H5SUPPORT_MUTEX_LOCK` and rename it to something unique if you want it to lock across any function call that contains this macro.

src/complex/Utilities/Parsing/Text/CsvParser.hpp
```diff
@@ -0,0 +1,69 @@
...
+    {
+      continue;
+    }
+    std::vector<std::string> tokens = complex::StringUtilities::split(buf, delim);
+    float value = std::atof(tokens[0].c_str());
+    data.push_back(value);
+    value = std::atof(tokens[1].c_str());
+    data.push_back(value);
+    value = std::atof(tokens[2].c_str());
+    data.push_back(value);
```
mmarineBlueQuartz, line 62:
```suggestion
    for(size_t i = 0; i < 3; i++)
    {
      float value = std::atof(tokens[i].c_str());
      data.push_back(value);
    }
```
I think that would make it more readable. Agreed?

src/complex/Utilities/Parsing/Text/CsvParser.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <vector>
+
+namespace complex
+{
+namespace CsvParser
+{
+
+std::vector<float> ParseVertices(const std::string& inputFile, const std::string& delimiter, bool headerLine)
+{
+  ;
```
mmarineBlueQuartz, line 19: The extra semicolon should probably be deleted.

src/complex/Utilities/StringUtilities.hpp
```diff
@@ -0,0 +1,224 @@
...
+    out.erase(back + 1);
+  }
+  else
+  {
+    out.clear();
+  }
+  return out;
+}
+
+inline std::string trimmed(const std::string& s)
```
mmarineBlueQuartz, line 158: Is there a reason this does not just call `ltrim` and `rtrim` and be done with it? With these being inlined, it shouldn't impose a penalty when compiling, and it's a lot easier to read that way than figuring out how it's different or keeping everything updated should changes be made.

src/complex/Utilities/StringUtilities.hpp
```diff
@@ -0,0 +1,224 @@
...
+{
+  return s.substr(0, s.size() - numElements);
+}
+
+template <typename T>
+inline std::string number(T arg)
+{
+  std::stringstream ss;
+  ss << arg;
+  return ss.str();
```
mmarineBlueQuartz, line 197:
```suggestion
  return std::to_string(arg);
```

PR #21, opened 2021-09-30
Add DataGroupSelection Filter parameter

imikejackson:

### DIFF
--- a/src/complex/Core/Parameters/DataGroupSelectionParameter.cpp
+++ b/src/complex/Core/Parameters/DataGroupSelectionParameter.cpp
@@ -1,5 +1,7 @@
 #include "DataGroupSelectionParameter.hpp"
 
+#include "complex/DataStructure/DataGroup.hpp"
+
 #include <fmt/core.h>
 
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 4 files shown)


src/complex/Core/Parameters/DataGroupSelectionParameter.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+Result<> DataGroupSelectionParameter::validatePath(const DataStructure& dataStructure, const DataPath& value) const
+{
+  const std::string prefix("FilterParameter 'DataGroupSelectionParameter' Validation Error: ");
+
+  if(value.empty())
+  {
+    return complex::MakeErrorResult(complex::FilterParameter::Constants::k_Validate_Empty_Value, fmt::format("{}DataPath cannot be empty", prefix));
+  }
+  const DataObject* object = dataStructure.getData(value);
```
JDuffeyBQ, line 87: We should probably also do a cast here to check if it's a `DataGroup`.

JDuffeyBQ (changes requested):

mmarineBlueQuartz (changes requested):
With a type name of _DataGroup_ SelectionParameter, I expect the target DataObject should be of type _DataGroup_.

With no documentation to the contrary and either very bad naming or an error in logic, this does not do what it sounds like it should. Either change the name or update the logic in `validatePath(const DataStructure& dataStructure, const DataPath& value)`

imikejackson:
Added a dynamic_cast<> with a check that the cast works otherwise return an error.

PR #24, opened 2021-09-30
DataStructure: Add makePath() method.

imikejackson:
This method allows a convenient method to generate a nested set of data groups

### DIFF
--- a/src/complex/DataStructure/DataStructure.cpp
+++ b/src/complex/DataStructure/DataStructure.cpp
@@ -1,8 +1,5 @@
 #include "DataStructure.hpp"
 
-#include <numeric>
-#include <stdexcept>
-
 #include "complex/DataStructure/BaseGroup.hpp"
 #include "complex/DataStructure/DataGroup.hpp"
 #include "complex/DataStructure/LinkedPath.hpp"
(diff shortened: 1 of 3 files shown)


src/complex/DataStructure/DataStructure.cpp
```diff
@@ -88,6 +88,40 @@ LinkedPath DataStructure::getLinkedPath(const DataPath& path) const
   }
 }
 
+LinkedPath DataStructure::makePath(const DataPath& path)
```
JDuffeyBQ, line 91: Since we want to report errors in the filter parameter, we should change this to a `Result<LinkedPath>` or at the very least `std::optional<LinkedPath>`.

JDuffeyBQ (changes requested):

src/complex/DataStructure/DataStructure.cpp
imikejackson: Is there an example of doing this in the code?

imikejackson: using Result<LinkedPath> just gives a compiler error. Not sure that API exists to be able to do that. Suggest we just merge this PR.

JDuffeyBQ:
```cpp
Result<LinkedPath> DataStructure::makePath(const DataPath& path)
{
  // actually make the path
  if(!sucess)
  {
     return MakeErrorResult<LinkedPath>(-1, "Unable to create path");
  }

  return {linkedPath};
}
```
Something like this?

Also what was the compile error?

And there's always `std::optional<LinkedPath>`.

imikejackson: Done

JDuffeyBQ: Looks like we don't have copy/move assignment for LinkedPath and that's causing the compile issue. Though since we have copy/move constructors I'd say there's no reason we shouldn't.

JDuffeyBQ (approved):

PR #25, opened 2021-09-30
Parameters: Add method to return comma delimited list of accepted keys

imikejackson:

### DIFF
--- a/src/complex/Filter/Parameters.cpp
+++ b/src/complex/Filter/Parameters.cpp
@@ -52,20 +52,14 @@ void Parameters::insert(IParameter::UniquePointer parameter)
   m_Params.insert({std::move(name), std::move(parameter)});
 }
 
-std::string Parameters::getAcceptedKeys() const
+std::vector<std::string> Parameters::getKeys() const
 {
-  std::stringstream acceptedKeys;
-  size_t count = 0;
+  std::vector<std::string> acceptedKeys;
   for(const auto& param : m_Params)
   {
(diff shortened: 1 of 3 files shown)


src/complex/Filter/Parameters.cpp
```diff
@@ -49,4 +51,21 @@ void Parameters::insert(IParameter::UniquePointer parameter)
   std::string name = parameter->name();
   m_Params.insert({std::move(name), std::move(parameter)});
 }
+
+std::string Parameters::getAcceptedKeys() const
```
JDuffeyBQ, line 55: Instead of `std::string getAcceptedKeys() const`, maybe we could have `std::vector<std::string> getKeys() const`. That could be reused in more places. Then in `IFilter` we can use `fmt::join(params.getKeys(), ", ")` instead.

imikejackson: Done. Returns std::vector<std::string>

src/complex/Filter/Parameters.cpp
```diff
@@ -1,5 +1,7 @@
 #include "Parameters.hpp"
 
+#include <sstream>
+
```
JDuffeyBQ, line 4:
```suggestion
```

JDuffeyBQ (changes requested):
Aside from the leftover header looks good.

src/complex/Filter/Parameters.cpp
imikejackson: Done

PR #28, opened 2021-10-01
Add Parameter to allow selection of a DataArray

imikejackson:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,7 @@ set(CoreFilters
 
 set(CoreParameters
   ArrayCreationParameter
+  ArraySelectionParameter
   BoolParameter
   ChoicesParameter
   DataGroupCreationParameter
--- a/src/complex/Core/Parameters/ArraySelectionParameter.cpp
+++ b/src/complex/Core/Parameters/ArraySelectionParameter.cpp
@@ -1,0 +1,96 @@
+#include "ArraySelectionParameter.hpp"
+
+#include <fmt/core.h>
+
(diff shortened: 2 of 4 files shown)


mmarineBlueQuartz (approved):
I lean towards method and class documentation should be filled in or not included, but there is nothing wrong with the code.

JDuffeyBQ (commented):
Should we add a type option for limiting the which DataArray the user can select? Or do we want to add that later and get this in now?

Otherwise looks good.

JDuffeyBQ (commented):
Also maybe we need an API to check if a `DataObject` is a `DataArray`? So we can check that when validating.

imikejackson:
> Also maybe we need an API to check if a `DataObject` is a `DataArray`? So we can check that when validating.

Added a check to make sure it isn't at least a DataGroup. Trying to check for DataArray can't work because DataArray is templated. If DataArray subclassed from something between DataObject and DataArray, like IDataArray, then you could check that. Is there an enumeration that we could maybe check instead? Otherwise merge this in.

JDuffeyBQ:
> 
> 
> > Also maybe we need an API to check if a `DataObject` is a `DataArray`? So we can check that when validating.
> 
> Added a check to make sure it isn't at least a DataGroup. Trying to check for DataArray can't work because DataArray is templated. If DataArray subclassed from something between DataObject and DataArray, like IDataArray, then you could check that. Is there an enumeration that we could maybe check instead? Otherwise merge this in.

I'm not aware of any, but Matthew would know for sure.

PR #29, opened 2021-10-04
[WIP] Test adding filters that are auto generated from current SIMPL

imikejackson:
ConditionalSetValue
CreateDataContainer
CreateGridMontage
CreateStringArray
CropVertexGeometry
DataContainerWriter
ExecuteProcess
FeatureCountDecision
FeatureDataCSVWriter
FindDerivatives
GenerateTiltSeries
GenerateVertexCoordinates
MaskCountDecision
PostSlackMessage
RawBinaryReader
RenameAttributeArray
RenameAttributeMatrix
RenameDataContainer
ReplaceValueInArray
RequiredZThickness
SplitAttributeArray
WriteTriangleGeometry

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,44 @@ set(CoreFilters
   ImportTextFilter
   TestFilter1
   TestFilter2
+        CombineAttributeArrays
+        CombineAttributeMatrices
+        ConditionalSetValue
+        ConvertColorToGrayScale
+        ConvertData
+        CopyFeatureArrayToElementArray
+        CopyObject
+        CreateDataContainer
+        CreateFeatureArrayFromElementArray
+        CreateGridMontage
+        CreateStringArray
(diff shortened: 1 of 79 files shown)


imikejackson:
DO NOT MERGE THIS. Using the PR to test auto generated code to ensure it will compile on other platforms.

imikejackson:
@JDuffeyBQ Take a look at the filters and see if you agree with what is being generated. If you have seen one filter you have seen them all. Let me know if anything needs to be change.

The generated code was from the perspective of someone outside of the project that is looking at their transformed code for the first time which is why I chose to use explicit types instead of 'auto' in some locations.

JDuffeyBQ (commented):
Looks good to me. The only addition that might be nice would be to take the comments from the `IFilter` functions and add them here.

imikejackson:
Done.

PR #30, opened 2021-10-08
Add convenience methods in FilterList and AbstractPlugin

imikejackson:
Add convenience functions in FilterList and AbstractPlugin

### DIFF
--- a/src/complex/Core/Application.cpp
+++ b/src/complex/Core/Application.cpp
@@ -97,6 +97,10 @@ Application::~Application()
 
 Application* Application::Instance()
 {
+  if(nullptr == s_Instance)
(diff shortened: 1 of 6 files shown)


imikejackson:
Interesting. Reading up on QApplication I was not aware that instance() can actually return a nullptr. Every singleton that I had researched on the Design patterns, they would all implement a static Instance() method that would initialize the static point object if needed, and then return that pointer. So this is a bit different in its use. I will adjust how I am using our "Application" object and push some changes to this repo.

imikejackson:
this is being replaced by https://github.com/BlueQuartzSoftware/complex/pull/33

PR #32, opened 2021-10-12
FILTER: Add ConditionalSetValue Filter

imikejackson:


imikejackson:
@JDuffeyBQ @mmarineBlueQuartz 

Please take a look at this filter as a starting point. I had to implement a few missing constants and a few more pieces of API to DataArray and DataStore. Feedback is welcome.

This filter also makes changes "in place" to an existing array. We have never fully figured out if this was going to be allowed  or not. Please advise.

There is a huge macro switch statement which might be better done as a constexpr but I have no idea how to write such a thing. Please advise.

imikejackson:
I think the filter could just take a string parameter and do the testing/validation in preflight that the value will convert to a numeric value.

This has been updated with working unit test.

PR #33, opened 2021-10-13
Update Plugin Architecture, Move core filters to CorePlugin

imikejackson:

### DIFF
--- a/src/complex/Filter/FilterList.cpp
+++ b/src/complex/Filter/FilterList.cpp
@@ -59,7 +59,7 @@ IFilter::UniquePointer FilterList::createFilter(const FilterHandle& handle) cons
 
 IFilter::UniquePointer FilterList::createFilter(const Uuid& uuid) const
 {
-  auto iter = std::find_if(m_PluginMap.cbegin(), m_PluginMap.cend(), [uuid](const decltype(*m_PluginMap.cbegin())& item) { return item.second->getPlugin()->containsFilterId(uuid); });
+  auto iter = std::find_if(m_PluginMap.cbegin(), m_PluginMap.cend(), [uuid](decltype(*m_PluginMap.cbegin())& item) { return item.second->getPlugin()->containsFilterId(uuid); });
 
   if(iter == m_PluginMap.cend())
   {
@@ -81,18 +81,6 @@ AbstractPlugin* FilterList::getPlugin(const FilterHandle& handle) const
     if(iter.second->getPlugin()->getId() == handle.getPluginId())
     {
       return iter.second->getPlugin();
-    }
-  }
-  return nullptr;
-}
-
-AbstractPlugin* FilterList::getPlugin(const std::string& humanOrClassName) const
-{
-  for(const auto& handle : getFilterHandles())
-  {
-    if(handle.getFilterName() == humanOrClassName || handle.getClassName() == humanOrClassName)
-    {
-      return getPlugin(handle);
     }
   }
   return nullptr;
--- a/src/complex/Filter/FilterList.hpp
+++ b/src/complex/Filter/FilterList.hpp
@@ -89,13 +89,6 @@ public:
   AbstractPlugin* getPlugin(const FilterHandle& handle) const;
 
   /**
-   * @brief
-   * @param humanOrClassName
-   * @return
-   */
-  AbstractPlugin* getPlugin(const std::string& humanOrClassName) const;
-
-  /**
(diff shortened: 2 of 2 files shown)


imikejackson:
@JDuffeyBQ @mmarineBlueQuartz There is a lot going on here and I will try to summarize:
+ Moved all the Core/Filters into their own plugin "ComplexCore" (updated CMake to reflect this)
+ Move complex/Core/Parameters to complex/Parameters
+ Moved remaining classes out of complex/Core/ into other places that made more sense
+ Added API to Application, FilterHandle and FilterList to allow easier instantiation of filters via their human or class name.
+ Updated unit tests so that they all pass. Their is one test for the Plugins that will fail when we start compiling more plugins because it assumes that there are 3 plugins which will change depending on how you compile.

## CMake Updates ##

You can set the `COMPLEX_EXTRA_PLUGINS` just like in DREAM3D to a semi-colon separated list of Plugins that you would like to compile. By default I have CMake looking at complex/../complex_plugins directory.

You can set the `COMPLEX_PLUGIN_SEARCH_DIRS` cmake variable to any number of directories and cmake will iterate on those directories looking for the various plugins. If nothing is found in those directories it will fall back to the complex/../complex_plugins directory.

There are CMake variables to ENABLE/DISABLE each plugin: `COMPLEX_PLUGIN_ENABLE_`.

This goes together with the [https://www.github.com/bluequartzsoftware/complex_plugins](https://www.github.com/bluequartzsoftware/complex_plugins) repository.

imikejackson:
@mmarineBlueQuartz @JDuffeyBQ 

Based on feedback from @JDuffeyBQ, Moved most of the CMake code for each plugin into a central CMake function located in cmake/Plugin.cmake. This really shortens up the cmake that the plugin author needs to write.

src/complex/Filter/FilterList.cpp
```diff
@@ -45,12 +62,24 @@ AbstractPlugin* FilterList::getPluginById(const FilterHandle::PluginIdType& id)
...
+{
+  for(const auto& handle : getFilterHandles())
+  {
+    if(handle.getFilterName() == humanOrClassName || handle.getClassName() == humanOrClassName)
+    {
+      return createFilter(handle);
+    }
+  }
+  return nullptr;
+}
```
mmarineBlueQuartz, line 75: I believe it is fundamentally a bad design to use only the class or human name. Having two perfectly valid plugins should not destroy the ability to instantiate a filter for no other reason than they both declared a filter with the same name.

If it was impossible for anyone outside of BlueQuartz to create plugins at any point, it would purely be our fault if this method failed. If we added a filter that shared a name with a filter another organization developed for their own internal workflow, when, if ever, would a user be notified that the incorrect filter was being instantiate? Would their plugin be allowed to be imported at all? Is it their job to rename their classes whenever we release new filters? What if that is not in their budget? What if a user / developer had pre-compiled plugins from multiple organizations with a naming conflict they couldn't change?

Do NOT use names. There are too many ways it can fail miserably. If you have to, use the search method and force the caller to pick a FilterHandle if multiple filters were found. FilterHandles were made to be the sole way to instantiate filters for this reason.

mmarineBlueQuartz (changes requested):
Never support creating filters purely off of their name, be it human or class. Even if it's rare, it isn't worth the risk of a collision, especially since the provided method never checks for one.

src/complex/Filter/FilterList.cpp
```diff
@@ -73,6 +69,18 @@ AbstractPlugin* FilterList::getPlugin(const FilterHandle& handle) const
...
+{
+  for(const auto& handle : getFilterHandles())
+  {
+    if(handle.getFilterName() == humanOrClassName || handle.getClassName() == humanOrClassName)
+    {
+      return getPlugin(handle);
+    }
+  }
+  return nullptr;
+}
```
mmarineBlueQuartz, line 82: I would expect getPlugin to use plugin name, not filter names... There is no method documentation saying that the plugin name has nothing to do with getPlugin(const std::string&), so I can only assume that this is incorrect or at best misleading. Either add method documentation explaining what the method does and why it only uses filter names (still not a good idea without collision checking) or update the method to do what it's name implies as that is the only thing currently explaining what the method does.

There's also far less plugins than FilterHandles, so if you are looking for a plugin by name, it would be better to iterate on those.

imikejackson: I removed this API.

PR #42, opened 2021-11-09
Add unit test to ComplexCore Plugin.

imikejackson:
The idea is to figure out the bare basics and use this as a template for other
plugins to use so we can auto generate those tests.


imikejackson:
The goal with this PR is to gather feedback about how both the actual unit test, but also the cmake involved is written. I just put the basics together but I want to use what comes out of this to auto generate the unit tests for the complex_plugins repository.

PR #73, opened 2022-01-25
Propose using getStatus to determine the toString

craigusr:


imikejackson:
I think this may not be needed once the newer notifications are implemented on the 01_pipeline branch

PR #77, opened 2022-01-27
Add support for multiple valid file extensions in FileSystemPathParameter

jmarquisbq:
Can now explicitly specify a list of all valid file extension types for Input/Output file path types or leave blank to allow any extensions.

### DIFF
--- a/src/complex/Parameters/FileSystemPathParameter.cpp
+++ b/src/complex/Parameters/FileSystemPathParameter.cpp
@@ -70,7 +70,7 @@ Result<> ValidateOutputDir(const FileSystemPathParameter::ValueType& path)
 namespace complex
 {
 //-----------------------------------------------------------------------------
-FileSystemPathParameter::FileSystemPathParameter(const std::string& name, const std::string& humanName, const std::string& helpText, const ValueType& defaultValue, PathType pathType, ExtensionsType extensionsType)
+FileSystemPathParameter::FileSystemPathParameter(const std::string& name, const std::string& humanName, const std::string& helpText, const ValueType& defaultValue, PathType pathType, const ExtensionsType extensionsType)
 : ValueParameter(name, humanName, helpText)
 , m_DefaultValue(defaultValue)
 , m_PathType(pathType)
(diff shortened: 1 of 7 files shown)


src/complex/Parameters/FileSystemPathParameter.hpp
```diff
@@ -25,10 +26,12 @@ class COMPLEX_EXPORT FileSystemPathParameter : public ValueParameter
     OutputDir = 3
   };
 
+  using ExtensionsType = std::unordered_set<std::string>;
+
   using ValueType = std::filesystem::path;
 
   FileSystemPathParameter() = delete;
-  FileSystemPathParameter(const std::string& name, const std::string& humanName, const std::string& helpText, const ValueType& defaultValue, PathType pathType);
+  FileSystemPathParameter(const std::string& name, const std::string& humanName, const std::string& helpText, const ValueType& defaultValue, PathType pathType, ExtensionsType extensionsType={});
```
mmarineBlueQuartz, line 34: Consider making the new parameter a const reference to avoid copying data

src/complex/Parameters/FileSystemPathParameter.cpp
```diff
@@ -70,10 +70,11 @@ Result<> ValidateOutputDir(const FileSystemPathParameter::ValueType& path)
 namespace complex
 {
 //-----------------------------------------------------------------------------
-FileSystemPathParameter::FileSystemPathParameter(const std::string& name, const std::string& humanName, const std::string& helpText, const ValueType& defaultValue, PathType pathType)
+FileSystemPathParameter::FileSystemPathParameter(const std::string& name, const std::string& humanName, const std::string& helpText, const ValueType& defaultValue, PathType pathType, ExtensionsType extensionsType)
```
mmarineBlueQuartz, line 73: Consider making the new parameter a const reference to avoid copying data.

mmarineBlueQuartz (approved):
While it is not a big deal, I marked the new parameter being passed by value instead of const reference, but I do not consider it major enough to force changes before a merge. It does look weird having const ref parameters followed by a single value parameter, however.

Just make sure to clang-format the affected files before merging.

src/complex/Parameters/FileSystemPathParameter.hpp
jmarquisbq: Ah good catch thanks.

jmarquisbq:
@mmarineBlueQuartz Yeah it does feel weird to have the const parameter after a non const parameter.  I can always remove the default value and move the new parameter to be before pathType if that's preferrable.  I don't think there are that many spots where this parameter is used yet.

PR #81, opened 2022-01-31
Lots of bug fixes for filters, pipelines, other classes and pipeline notifications.

imikejackson:
Together these changes allow the first of the example Pipelines to complete with the correct results.
Unit tests for the Threshold Filter, min size, raw binary reader filters all need to be written

+ DataArray
  - Add early return to copyTuple if the source and destination are the source
  - Add getTypeName() which returns more information in the string about the type 'DataArray<uint8_t>'

+ DataStore
  - Remove commennted code

+ IDataArray
  - Added virtual getTypeName() method

+ NeighborList()
  - Add public default constructor
  - Add some DOxygen doc skeleton

+ DataArrayFactory
  - Added method to get the 'typeName' from the actual DataArray<T>
  - Supports reading .dream3d (v8) files because of the change to getTypeName()

+ ArrayThresholdsParmaeter
  - Refactored a few variables
  - Fixed an error message

+ Pipeline
  - Added std::cout for each filter. Assists debugging (probably not needed long term)

+ DataArrayUtilities
  - Updated ReplaceValue() with additional template parameters in support of redesigned Threshold filter

+ DataGroupUtilities
  - New header class for DataGroup related operations
  - Added RemoveInactiveObjects() in support of RemoveMinimumSizeFilter

+ H5DataFactoryManager
  - Added new factories for each of the DataArray Types

+ CalculateFeatureSizes
  - Set tuple dimensions to {1} during preflight
  - Ensure created DataArrays have the proper Tuple Dimensions

+ ConditionalSetValue
  - Ensure we are checking DataArrays are the types expected by the filter.
  - Create constants for error codes

+ CreateDataArray
  - Added DynamicaTableData parameter to get the Tuple Dimensions from the user.

+ FindDifferencesMap
  - ClangTidy misc fixes
  - Ensure output DataArray has the proper TupleShape

+ MultiThresholdObjects
  - Rewrote the threshold classes for huge performance gain. Now as fast as SIMPL based codes
  - These updates need further unit testing

+ RawBinaryReaderFilter
  - Added some help text for the Tuple Dims parameter

+ RemoveMinimunSizeFeaturesFilter
  - Added preflight checks for valid DataArrays
  - Finished implementing the final part of the filter where the feature data arrays are updated

+ ScalarSegmentFeatures | SegmentFeatures
  - Moved most of the code to an 'Algorithms' class
  - ScalarSegmentFeaturesFilter no longer subclasses from SegmentFeaturesFilter
  - Redesigned the ScalarSegmentFeatures algorithm for large performance gain


src/Plugins/ComplexCore/src/ComplexCore/Filters/RemoveMinimumSizeFeaturesFilter.cpp
```diff
@@ -12,54 +12,53 @@
...
+using NeighborsArrayType = Int32Array;
+
 namespace
 {
 constexpr int32 k_BadMinAllowedFeatureSize = -5555;
 constexpr int32 k_BadNumCellsPath = -5556;
 constexpr int32 k_ParentlessPathError = -5557;
 
-void assign_badpoints(DataStructure& data, const DataPath& featureIdsPath, SizeVec3 dimensions, const Int32Array* numCellsPtr)
+void assign_badpoints(DataStructure& dataStructure, const DataPath& featureIdsPath, SizeVec3 dimensions, const NumCellsArrayType* numCellsPtr)
```
JDuffeyBQ, line 31:
```suggestion
void assign_badpoints(DataStructure& dataStructure, const DataPath& featureIdsPath, SizeVec3 dimensions, const NumCellsArrayType& numCellsPtr)
```

src/Plugins/ComplexCore/src/ComplexCore/Filters/RemoveMinimumSizeFeaturesFilter.cpp
```diff
@@ -192,19 +192,24 @@ void assign_badpoints(DataStructure& data, const DataPath& featureIdsPath, SizeV
   }
 }
 
-std::vector<bool> remove_smallfeatures(DataArray<int64>* featureIdsPtr, const DataArray<int32>* numCellsPtr, const DataArray<int32>* featurePhasesPtr, int32 phaseNumber, bool applyToSinglePhase,
+// -----------------------------------------------------------------------------
+std::vector<bool> remove_smallfeatures(FeatureIdsArrayType* featureIdsPtr, const NumCellsArrayType* numCellsPtr, const PhasesArrayType* featurePhasesPtr, int32_t phaseNumber, bool applyToSinglePhase,
```
JDuffeyBQ, line 196:
```suggestion
std::vector<bool> remove_smallfeatures(FeatureIdsArrayType& featureIdsPtr, const NumCellsArrayType& numCellsPtr, const PhasesArrayType& featurePhasesPtr, int32_t phaseNumber, bool applyToSinglePhase,
```

src/Plugins/ComplexCore/src/ComplexCore/Filters/RemoveMinimumSizeFeaturesFilter.cpp
```diff
@@ -373,18 +380,23 @@ Result<> RemoveMinimumSizeFeaturesFilter::executeImpl(DataStructure& data, const
...
+  PhasesArrayType* featurePhasesArray = nullptr;
+  if(applyToSinglePhase)
+  {
+    featurePhasesArray = dataStructure.getDataAs<PhasesArrayType>(featurePhasesPath);
+  }
 
-  auto featureIdsPtr = data.getDataAs<Int64Array>(featureIdsPath);
-  auto featureIds = featureIdsPtr->getDataStore();
+  FeatureIdsArrayType* featureIdsPtr = dataStructure.getDataAs<FeatureIdsArrayType>(featureIdsPath);
+  FeatureIdsArrayType::store_type* featureIds = featureIdsPtr->getDataStore();
```
JDuffeyBQ, line 390:
```suggestion
  FeatureIdsArrayType& featureIdsPtr = dataStructure.getDataRefAs<FeatureIdsArrayType>(featureIdsPath);
  FeatureIdsArrayType::store_type& featureIds = featureIdsPtr.getDataStore();
```

src/Plugins/ComplexCore/src/ComplexCore/Filters/RemoveMinimumSizeFeaturesFilter.cpp
```diff
@@ -373,18 +380,23 @@ Result<> RemoveMinimumSizeFeaturesFilter::executeImpl(DataStructure& data, const
...
 
-  auto numCellsPtr = data.getDataAs<Int32Array>(numCellsPath);
-  auto numCells = numCellsPtr->getDataStore();
+  NumCellsArrayType* numCellsPtr = dataStructure.getDataAs<NumCellsArrayType>(numCellsPath);
+  NumCellsArrayType::store_type* numCells = numCellsPtr->getDataStore();
```
JDuffeyBQ, line 393:
```suggestion
  NumCellsArrayType& numCellsPtr = dataStructure.getDataRefAs<NumCellsArrayType>(numCellsPath);
  NumCellsArrayType::store_type& numCells = numCellsPtr.getDataStore();
```

src/Plugins/ComplexCore/src/ComplexCore/Filters/RemoveMinimumSizeFeaturesFilter.cpp
```diff
@@ -409,12 +421,12 @@ Result<> RemoveMinimumSizeFeaturesFilter::executeImpl(DataStructure& data, const
     return {nonstd::make_unexpected(std::vector<Error>{errorReturn})};
   }
 
-  auto imageGeom = data.getDataAs<ImageGeom>(imageGeomPath);
-  assign_badpoints(data, featureIdsPath, imageGeom->getDimensions(), numCellsPtr);
+  auto imageGeom = dataStructure.getDataAs<ImageGeom>(imageGeomPath);
```
JDuffeyBQ, line 424:
```suggestion
  auto& imageGeom = dataStructure.getDataRefAs<ImageGeom>(imageGeomPath);
```

CMakeLists.txt
```diff
@@ -335,7 +336,7 @@ set(COMPLEX_SRCS
   ${COMPLEX_SOURCE_DIR}/Filter/FilterHandle.cpp
   ${COMPLEX_SOURCE_DIR}/Filter/FilterList.cpp
 
-  ${COMPLEX_SOURCE_DIR}/DataStructure/Factory/DataArrayFactory.cpp
+  # ${COMPLEX_SOURCE_DIR}/DataStructure/Factory/DataArrayFactory.cpp
```
JDuffeyBQ, line 339:
```suggestion
```

JDuffeyBQ (changes requested):

src/complex/Pipeline/Pipeline.cpp
```diff
@@ -230,24 +230,30 @@ bool Pipeline::executeFrom(index_type index, DataStructure& ds)
   setHasWarnings(hasWarningsBeforeIndex(index));
   setHasErrors(false);
   setIsExecuting();
-
+  size_t currentIndex = 0;
   for(auto iter = begin() + index; iter != end(); iter++)
   {
+    std::cout << "[" << currentIndex << "] " << iter->get()->getName() << " Starting Filter Execution" << std::endl;
```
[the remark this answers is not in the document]
mmarineBlueQuartz: A debug output method might be good too. Something that can be enabled for all non-release builds would be easier to manage and find on top of being self-documenting.

src/complex/DataStructure/DataArray.hpp
```diff
@@ -496,6 +500,61 @@ class DataArray : public IDataArray
...
+    }
+    else if constexpr(std::is_same_v<T, double>)
+    {
+      typeName = "DataArray<double>";
+    }
+    else if constexpr(std::is_same_v<T, bool>)
+    {
+      typeName = "DataArray<bool>";
+    }
+    return typeName;
```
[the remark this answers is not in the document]
imikejackson: Why is this needed? int8 and int8_t should resolve to the same thing? This will also have ramifications on reading/writing the dream3d files as the typeName is written as a piece of meta-data that is then used during reading to determine what kind of DataArray to create.

src/complex/DataStructure/NeighborList.hpp
```diff
@@ -25,6 +25,8 @@ class NeighborList : public IDataArray
   using VectorType = std::vector<T>;
   using SharedVectorType = std::shared_ptr<VectorType>;
 
+  NeighborList() = default;
```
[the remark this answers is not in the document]
imikejackson: And why, exactly, can't I make a stand alone DataArray or NeighborList?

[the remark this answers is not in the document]
mmarineBlueQuartz: DataObject uses the DataStructure both directly and indirectly including within its constructor. Most DataObject operations either require either having or having had a DataStructure, which is why all constructors either require a DataStructure reference or copy values from an existing DataObject.

If you need a temporary DataObject, you can create a temporary DataStructure first that operates separately than the one data is being stored / accessed from. That way you avoid the possibility of existing observers being notified to temporary data being created and destroyed while also ensuring that temporary data is cleaned up at the end of the structure's scope without having to delete all temporary data manually like you would if you used the main DataStructure. It's one extra step to avoid a lot of misuse and confusion as to why data created that way isn't working like everything else is.

src/complex/DataStructure/Factory/DataArrayFactory.hpp
```diff
@@ -1,23 +1,45 @@
...
 public:
-  DataArrayFactory();
+  DataArrayFactory()
+  : IDataFactory()
+  {
+    DataStructure ds;
+    using DataArrayType = DataArray<T>;
+    using DataStoreType = DataStore<T>;
+    DataArrayType* temp = DataArrayType::template CreateWithStore<DataStoreType>(ds, "TEMP", {1ULL}, {1ULL});
+    m_TypeName = temp->getTypeName();
```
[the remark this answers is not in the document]
mmarineBlueQuartz: Also, if you are creating temp data, consider putting it in a temp DataStructure. Otherwise, you have to worry about cleaning it up, name conflicts, etc. If another filter forgets to clean up its temp data, that shouldn't prevent you from creating yours. Using RAII to clean up all temp data under all error conditions or branching return statements is a lot easier than manually handling it.

mmarineBlueQuartz (approved):

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportTextFilter.cpp
```diff
@@ -87,6 +87,8 @@ IFilter::PreflightResult ImportTextFilter::preflightImpl(const DataStructure& da
 
 Result<> ImportTextFilter::executeImpl(DataStructure& data, const Arguments& args, const PipelineFilter* pipelineNode, const MessageHandler& messageHandler) const
 {
+  messageHandler.operator()({IFilter::Message::Type::Progress, "Read ASCII Progress", 50});
```
JDuffeyBQ, line 90:
```suggestion
  messageHandler({IFilter::Message::Type::Progress, "Read ASCII Progress", 50});
```

src/complex/Filter/IFilter.hpp
```diff
@@ -35,12 +35,14 @@ class COMPLEX_EXPORT IFilter
...
       Info = 0,
       Debug,
+      Progress,
       Warning,
       Error
     };
 
     Type type = Type::Info;
     std::string message;
+    int32 progress;
```
JDuffeyBQ, line 45: Do we need a dedicated progress message type? And if so do we need to include the progress value for every message type?

JDuffeyBQ (dismissed):

mmarineBlueQuartz (commented):
I don't have much more to add that Jared did not mention, but I do have a few extra comments.

I stand by my previous note in DataArray.hpp that the description on line 48 was incorrectly changed. The DataArray instance takes ownership, not the method call instance.

I also agree with Jared that you should be using the type aliases defined in Types.hpp wherever possible. It's two characters less, but it does look cleaner when reading and it looks bad to mix and match. Either everything should use them (and most things do), or they should be removed. I lean towards the former. I doubt anything will be changed in this PR, but something to keep in mind going forwards.

imikejackson:
@JDuffeyBQ  In reference to the Progress type of message; In SIMPL we did have specific message types for pipeline and filter and for warning, error, message, progress + message. They all subclassed from AbstractMessage. The current design in `complex` is a single struct Message that has an enum to tell the difference. We may want to consider moving the `Message` struct outside of the `IFilter` class so that we can have subclasses of it. I think the `MessageHandler` can stay inside of `IFilter` then we could overload the `operator()()` to take the various kinds of messages. Then we would need to modify the `PipelineFilter::notifyFilterMessage()` method to understand what kind of message is coming in and dispatch that out to the correct method in `AbstractPipelineNode`.

imikejackson:
@mmarineBlueQuartz I fixed up the comment in DataArray.hpp. I think I fat fingered a copy/paste on that one.

RE: Types: the std types have been around and standardized for a long time now. There really isn't a need to go and invent "yet another way to declare an int32". While I could get on board with the integer types the size_t, float and double are a bit mind boggling frankly. The only thing that we have done is create a bit of confusion when developers see "usize" which comes from the Rust language. Doing a cursory search on google for "c++ usize" gets a bunch of code that is bridging Rust to C/C++. The only thing that has been done is to add additional mental workload for developers to have to remember to use "yet another non-standard" naming convention. `float` and `double` are the same way. In our `Types.hpp` file all we have done is create an alias for each one. `Float32` and `Float64` don't actually convey any more information than their original types, float and double. When we first talked about this it was brought up that float and double do not, according to the spec, *have* to mean 32 bit and 64 bit floating point values. Essentially every compiler just does it that way. Also by using `Float32` and `Float64` we have introduced marginally worse readability as Float32 and Float64 now look similar as one is reading through code. Again, more mental load on the developer to differentiate between them. And back to the original argument of 32 and 64 bit, `Float32` and `Float64` are just "comfort" aliases. If the compiler changes `float` to mean 64 bit float at some point in the future, then we have a `Float32` type that is now conveying the wrong information to the developer. This is along the same lines as `long int` and why it is banned in our source codes: Depending on how you compile you will get either a 32 or 64 bit integer. Using `float` and `double` stand out when reading code mush easier than `Float32` and `Float64`. All of these reasons are why I stay away from the Qt specific alias like `qlong`. Same problems.

Mixing an matching styles: If we were dead serious about the mixing and matching of styles (camel case versus snake case) then we wouldn't be using the STL (snake_case) or a bunch of libraries for that matter. So using this argument for int32_t versus int32 sort of falls flat. We should alias <vector> to be Vector and vector.push_back() to be Vector.pushBack().

I don't know how many copy/paste bugs I had to fix in DREAM3D/SIMPL due to int8/int16/int32/int64 because. Not single one that I can remember between `float` and `double`. It was *always* with the integer types. Because they are *easy* to gloss over when writing code or editing within a code dense section where you have a large if-else or switch block.

If just use the value types from the C++ std then it is one less thing that I have to remind incoming C++ developers about when they are using our code base.

I've said my peace.

JDuffeyBQ (approved):

PR #89, opened 2022-02-01
Added Output DataPath rename checks

mmarineBlueQuartz:
* PipelineFilter now checks for renamed outputs between preflights. A signal is emitted when the change in arguments results in a created DataObject having a different DataPath.
* Added IDataCreationAction as a subclass of IDataAction specific for the creation of DataObjects in the DataStructure.
* Pipelines adjust the arguments of PipelineFilter nodes following a renamed DataPath creation.
* Added a renamed output test to PipelineTest.cpp

### DIFF
--- a/src/complex/DataStructure/DataPath.cpp
+++ b/src/complex/DataStructure/DataPath.cpp
@@ -106,6 +106,28 @@ DataPath DataPath::replace(std::string_view symbol, std::string_view targetName)
   return DataPath(std::move(newPath));
 }
 
+bool DataPath::attemptRename(const DataPath& oldPath, const DataPath& newPath)
+{
+  if(oldPath.getLength() > getLength())
+  {
+    return false;
+  }
+
+  for(usize i = 0; i < oldPath.getLength(); i++)
+  {
+    if(oldPath.m_Path[i] != m_Path[i])
+    {
+      return false;
+    }
+  }
+
+  for(usize i = 0; i < oldPath.getLength(); i++)
+  {
+    m_Path[i] = newPath.m_Path[i];
+  }
+  return true;
+}
+
 bool DataPath::operator==(const DataPath& rhs) const
 {
   return m_Path == rhs.m_Path;
--- a/src/complex/Filter/Actions/CreateArrayAction.cpp
+++ b/src/complex/Filter/Actions/CreateArrayAction.cpp
@@ -11,10 +11,10 @@ using namespace complex;
 namespace complex
 {
 CreateArrayAction::CreateArrayAction(NumericType type, const std::vector<usize>& tDims, const std::vector<usize>& cDims, const DataPath& path)
-: m_Type(type)
+: IDataCreationAction(path)
+, m_Type(type)
 , m_Dims(tDims)
 , m_CDims(cDims)
-, m_Path(path)
 {
 }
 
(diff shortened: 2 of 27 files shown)


imikejackson (commented):
I think before this is merged we should have an idea of how NX will use this API to ensure the signals are getting pushed up the chain as needed.

imikejackson (approved):

src/complex/DataStructure/DataPath.cpp
```diff
@@ -106,6 +106,28 @@ DataPath DataPath::replace(std::string_view symbol, std::string_view targetName)
   return DataPath(std::move(newPath));
 }
 
+bool DataPath::attemptRename(const DataPath& oldPath, const DataPath& newPath)
```
JDuffeyBQ, line 109: I'm a little confused on this function. Why does it need to take in `oldPath`?

src/complex/Filter/Actions/CreateArrayAction.cpp
```diff
@@ -77,6 +77,6 @@ const std::vector<usize>& CreateArrayAction::componentDims() const
 
 const DataPath& CreateArrayAction::path() const
 {
-  return m_Path;
+  return getCreatedPath();
```
JDuffeyBQ, line 80: Like the warnings say, this is return a reference to a temporary object.

src/complex/Filter/Output.hpp
```diff
@@ -43,6 +43,41 @@ class IDataAction
...
+public:
+  /**
+   * @brief Constructs an IDataCreationAction that takes a DataPath specifying the path to be created.
+   * @param createdPath
+   */
+  IDataCreationAction(const DataPath& createdPath)
+  : m_CreatedPath(createdPath)
+  {
+  }
+  virtual ~IDataCreationAction() = default;
```
JDuffeyBQ, line 61:
```suggestion
  ~IDataCreationAction() noexcept override = default;
```

src/complex/Pipeline/Messaging/OutputRenamedMessage.hpp
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * @brief Constructs a OutputRenamedMessage specifying which AbstractPipelineNode
+   * was added to the target Pipeline and the index it was added at.
+   * @param pipeline
+   * @param newNode
+   * @param index
+   */
+  OutputRenamedMessage(AbstractPipelineNode* pipeline, const PipelineFilter::RenamedPath& renamedOutput);
+
+  virtual ~OutputRenamedMessage();
```
JDuffeyBQ, line 29:
```suggestion
  ~OutputRenamedMessage() noexcept override;
```

src/complex/Pipeline/AbstractPipelineNode.hpp
```diff
@@ -122,6 +125,16 @@ class COMPLEX_EXPORT AbstractPipelineNode
...
 
+  /**
+   * @brief Attempts to preflight the node using the provided DataStructure.
+   * Returns true if preflighting succeeded. Otherwise, this returns false.
+   * @param data
+   * @param RenamedPaths& renamedPaths = {} Collection of renamed output paths.
+   * Only used for PipelineFilters.
+   * @return bool
+   */
+  virtual bool preflight(DataStructure& data, RenamedPaths& renamedPaths, const std::atomic_bool& shouldCancel) = 0;
```
JDuffeyBQ, line 136: If `renamedPaths` is an output, should it be part of the return type?

src/complex/Pipeline/PipelineFilter.cpp
```diff
@@ -168,6 +248,162 @@ bool PipelineFilter::execute(DataStructure& data, const std::atomic_bool& should
...
+ * @brief Removes the overlap between two groups of DataPaths
+ * @param group1
+ * @param group2
+ */
+void removeOverlap(std::vector<DataPath>& group1, std::vector<DataPath>& group2)
+{
+  auto overlap = getOverlap(group1, group2);
+  for(const auto& value : overlap)
+  {
+    std::remove(group1.begin(), group1.end(), value);
```
JDuffeyBQ, line 308: If `std::remove` is intending to actually remove items from the vector, you will also need a call to `erase`. This is also the source of the `nodiscard` warnings.

JDuffeyBQ (changes requested):

src/complex/DataStructure/DataPath.cpp
mmarineBlueQuartz: This method checks if an attempt to rename is valid before applying it.

Let's say you have DataPath _Foo / Bar / Bazz_. _Foo / Bar_ was renamed to _Foo / Crobar_. _Foo / Bar_ is the old path used to check if the renaming is applicable when renaming. Meanwhile, if _Foo / Random_ was renamed, _Random_ does not match the value in the path and the rename would not be relevant to the current DataPath.

src/complex/Pipeline/AbstractPipelineNode.hpp
mmarineBlueQuartz: The output is the success state. The renamedPaths is a parameter that is shared through the course of preflighting a Pipeline.

src/complex/Pipeline/Messaging/OutputRenamedMessage.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+using namespace complex;
+
+OutputRenamedMessage::OutputRenamedMessage(AbstractPipelineNode* pipeline, const PipelineFilter::RenamedPath& renamedOutput)
+: AbstractPipelineMessage(pipeline)
+, m_RenamedPath(renamedOutput)
+{
+}
+
+OutputRenamedMessage::~OutputRenamedMessage() = default;
```
JDuffeyBQ, line 11:
```suggestion
OutputRenamedMessage::~OutputRenamedMessage() noexcept = default;
```
Fixes exception specification mismatch.

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #91, opened 2022-02-02
Moving Example filters into TestOne plugin, and adding ErrorWarningFilter to TestOne plugin

joeykleingers:
* Combining TestOne and TestTwo plugins from Core into Test plugin.
* Adding ErrorWarningFilter from SIMPL.
* Updating MakeWarningVoidResult to MakeWarningResult so that it can be used in PreflightImpl methods.


joeykleingers:
Updated this PR to do the correct thing (misunderstanding between Mike and I), which is to move the example filters into the TestOne plugin, and then add ErrorWarningFilter to TestOne plugin as well.

joeykleingers:
It appears that warning and error messages are not being received during execution of filters.  You can test it by turning on the Execute Warning and Execute Error checkboxes in the ErrorWarningFilter.

Also, there are duplicate filter warning messages being sent in preflight because both the node and the filter are notifying the issues table of the same warning message.  I don't know if maybe we should not have the node report warning messages and just let the filter do it?

imikejackson:
I guess a unit test would allow us to figure out if the filter works at the 'complex' level. NX will sort itself out soon enough.

joeykleingers:
The unit test passes, so the complex level is fine.

imikejackson (approved):

PR #92, opened 2022-02-03
[WIP] Use IDataArray->getDataType() instead of dynamic_cast<> to figure out code paths

imikejackson:


src/Plugins/ComplexCore/src/ComplexCore/Filters/InterpolatePointCloudToRegularGridFilter.cpp
```diff
@@ -79,45 +79,44 @@ void mapPointCloudDataByKernel(IDataArray* source, IDataArray* dynamic, std::vec
...
-  else if(dynamic_cast<DataArray<float64>*>(source))
-  {
+    break;
+  case complex::DataType::float64:
     mapPointCloudDataByKernel<float64>(source, dynamic, kernelVals, kernel, dims, curX, curY, curZ, vertIdx);
+    break;
+  case complex::DataType::boolean:
+    mapPointCloudDataByKernel<bool>(source, dynamic, kernelVals, kernel, dims, curX, curY, curZ, vertIdx);
+    break;
+  case complex::DataType::error:
```
JDuffeyBQ, line 118: I'd change this to be the default case. Once we change `NeighborList` to inherit directly from `DataObject` instead of `IDataArray`, we should be able to remove `DataType::error`.
```suggestion
  default:
```

PR #94, opened 2022-02-04
Add allowed types to ArraySelectionParameter

JDuffeyBQ:


mmarineBlueQuartz (approved):
One thing that could be convenient is another constructor to ArraySelectionParameter that leaves out allowEmpty (default initializes to false) for cases where only the AllowedTypes needs to be specified. Or perhaps switch the allowedTypes and allowEmpty order since I imagine allowedTypes is specified more often than allowEmpty.

Everything else looks good.

imikejackson:
Do we have the notion of an "optional" parameter in the arguments?

That is essentially what "allowEmpty" is saying. The filter either requires a value in it or it optionally does not because there is some other criteria that isn't being met and that allows the value to be empty.

JDuffeyBQ:
> Do we have the notion of an "optional" parameter in the arguments?
> 
> That is essentially what "allowEmpty" is saying. The filter either requires a value in it or it optionally does not because there is some other criteria that isn't being met and that allows the value to be empty.

I had envisioned this originally for `Parameters`, but never implemented it due to time constraints. It would be my preferred solution as I mentioned when `allowEmpty` was added.

imikejackson:
@JDuffeyBQ What is your idea/design for optional Arguments? would it be something like "params.insertOptional(...)"

JDuffeyBQ:
For parameters that are always optional that would work, but if we want to support dependent parameters we could tie it to the group functionality. It just depends on what kind of dependencies we need to support.

imikejackson (approved):

PR #97, opened 2022-02-08
Fixed CalculateTriangleAreasFilter accepting non triangle geoms

JDuffeyBQ:


imikejackson:
@JDuffeyBQ I think the algorithm should be changed to accept tetrahedral geometries as that is actually made up of triangles.

JDuffeyBQ:
The original filter did not accept any other geometries. As of right now, the filter will fail with an exception if given another geometry type.

mmarineBlueQuartz (approved):

JDuffeyBQ:
I added #100 for adding tetrahedral support. I'll merge this so we don't deal with failures now.

PR #102, opened 2022-02-08
Remove unnecessary RunState data member from AbstractPipeline

jmarquisbq:


imikejackson (approved):
I think this should work correctly. @JDuffeyBQ  What are your thoughts?

JDuffeyBQ (approved):

PR #104, opened 2022-02-09
Allow pipelines to be canceled

JDuffeyBQ:
* Changed IFilter cancel member variable to std::atomic_bool arg


imikejackson:
What is the benefit of sending the atomic_bool into the algorithm classes instead of the IFilter pointer?

JDuffeyBQ:
`IFilter` no longer has cancel as a member variable.

imikejackson (approved):

PR #107, opened 2022-02-10
Fixed issue when trying to insert objects with duplicate names

JDuffeyBQ:
When creating a new `DataObject` and trying to insert into `BaseGroup`, the id is generated before the insertion. Then in `BaseGroup::canInsert(const DataObject*)`, it calls `BaseGroup::contains(const DataObject*)` which checks if the id of the `DataObject` already exists in the group. But since the id was just generated it will never be in the group and `contains` will always return `false`. So objects with the same name will exist in the group which will cause an issue when using a `DataPath` to attempt to access them.

### DIFF
--- a/src/complex/DataStructure/BaseGroup.cpp
+++ b/src/complex/DataStructure/BaseGroup.cpp
@@ -87,7 +87,7 @@ bool BaseGroup::canInsert(const DataObject* obj) const
   {
     return false;
   }
-  if(contains(obj))
+  if(contains(obj) || contains(obj->getName()))
   {
     return false;
   }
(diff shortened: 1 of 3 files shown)


imikejackson (dismissed):

imikejackson:
Could we get a unit test to exercise this?

JDuffeyBQ:
> The failing ConditionalSetValue test is fixed in PR #108. If want to merge that PR first, then rebase this on top of that, then everything should work.
> 
> Would still like to see a unit test covering this use case written

Once it passes, we could do that.

imikejackson (dismissed):

imikejackson:
Other PR fixed and passes.

mmarineBlueQuartz (approved):

PR #113, opened 2022-02-12
Add a 'contains' method to Metadata() class.

imikejackson:
Add a constant for the filter index that something was added


JDuffeyBQ (changes requested):
The other changes should be in a separate PR.

imikejackson:
Changes addressed.

PR #119, opened 2022-02-17
Added AbstractDataStore::fillTuple()

JDuffeyBQ:


imikejackson:
Don't we have that with "initializeTuple()" or initializeValue() or something like that ? maybe at the DataArray level?

JDuffeyBQ:
Hmm somehow I missed that. I'd still say we should have the function on `AbstractDataStore` and then `DataArray` would just call that.

imikejackson:
> Hmm somehow I missed that. I'd still say we should have the function on `AbstractDataStore` and then `DataArray` would just call that.

Sounds good.

mmarineBlueQuartz (approved):

PR #121, opened 2022-02-17
FILTER: InitializeData 

JDuffeyBQ:


mmarineBlueQuartz (dismissed):

JDuffeyBQ:
(min, max) for each x/y/z or (x, y, z) for each min/max? The code example is different from the text.

imikejackson:
I tried xyz Min and xyz max and it looked ok from a visual point of view. I never know which way to go on these. 3 sets of 2 or 2 sets of 3. either way works as we get to label everything in the GUI.

imikejackson (changes requested):
Change the input parameters for the min max and this should be good to go.

imikejackson (approved):

PR #123, opened 2022-02-18
Feature/NeighborList HDF5

mmarineBlueQuartz:
* Updated NeighborLists to be read and written to HDF5.
* Additional HDF5 data object is linked to match NeighborList data links.
* Added INeighborList base class for NeighborList<T>
* Updated HDF5 unit tests

### DIFF
--- a/src/complex/DataStructure/INeighborList.hpp
+++ b/src/complex/DataStructure/INeighborList.hpp
@@ -1,0 +1,67 @@
+#pragma once
+
+#include "complex/DataStructure/IDataArray.hpp"
+
+namespace complex
+{
+/**
+ * @brief Non-templated base class for NeighborList class.
+ */
+class COMPLEX_EXPORT INeighborList : public IDataArray
+{
+public:
+  virtual ~INeighborList();
+
+  /**
+   * @brief Returns typename of the DataObject as a std::string.
+   * @return std::string
+   */
+  std::string getTypeName() const override;
+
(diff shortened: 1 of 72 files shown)


src/complex/DataStructure/INeighborList.hpp
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#include "complex/DataStructure/IDataArray.hpp"
+
+namespace complex
+{
+/**
+ * @brief Non-templated base class for NeighborList class.
+ */
+class COMPLEX_EXPORT INeighborList : public IDataArray
```
JDuffeyBQ, line 10: Do we want `INeighborList` to inherit from `IDataArray`?

mmarineBlueQuartz: For the sake of complexity, that needs to be its own PR. I'm stashing my changes for now to create a new branch with all the required changes. ArraySelectionParameter needs to be split for the two "array" types, filters need to be updated, and various helper methods need their parameters changed.

This PR is about HDF5, so the changes should be kept to that topic even if it lays the groundwork for additional changes. The PR would get way too much bloat otherwise.

imikejackson (approved):

imikejackson:
Is this ready to merge? I have other work that is waiting on this PR to be merged.
@mmarineBlueQuartz @JDuffeyBQ

PR #124, opened 2022-02-18
FILTER: ExtractInternalSurfacesFromTriangleGeometry

mmarineBlueQuartz:
* Added ExtractInternalSurvacesFromTriangleGeom filter
* Added unit test
* Added CreateTriangleGeomAction
* Renamed DataObject::getParentDataPaths to getDataPaths due to misleading name. DataObject::getParentDataPaths did not return DataPaths to known parents. It returned the DataPaths to the object calling it.

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/ExtractInternalSurfacesFromTriangleGeometry.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/ExtractInternalSurfacesFromTriangleGeometry.cpp
@@ -1,6 +1,7 @@
 #include "ExtractInternalSurfacesFromTriangleGeometry.hpp"
 
 #include <string>
+#include <unordered_map>
 
 #include "fmt/format.h"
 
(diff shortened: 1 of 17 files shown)


JDuffeyBQ (commented):
The `DataObject::getParentDataPaths()` changes should probably be in a separate PR. And I thought there was already a change about that function?

imikejackson:
@mmarineBlueQuartz  What is the function `getDataPaths()` *supposed* to return? Because it returns the DataPaths of all the parents of the object. If it is *supposed* to return the children DataPaths then there is a logic error in the function.

src/Plugins/ComplexCore/src/ComplexCore/Filters/ExtractInternalSurfacesFromTriangleGeometry.cpp
```diff
@@ -0,0 +1,375 @@
...
+    {
+      usize hash = std::hash<float32>()(vert[0]);
+      hashCombine(hash, vert[1]);
+      hashCombine(hash, vert[2]);
+      return hash;
+    }
+  };
+
+  typedef std::array<int64, 3> Triangle;
+  typedef std::unordered_map<Vertex, int64, ArrayHasher> VertexMap;
```
JDuffeyBQ, line 224:
```suggestion
  using Triangle = std::array<int64, 3>;
  using VertexMap = std::unordered_map<Vertex, int64, ArrayHasher>;
```

JDuffeyBQ (changes requested):
This just needs the `getDataPaths()` changes extracted to a separate PR with whatever name we ended up deciding on.

imikejackson (approved):

PR #130, opened 2022-02-21
ChoicesParameter update

imikejackson:


src/Plugins/ComplexCore/test/ChangeAngleRepresentationTest.cpp
```diff
@@ -0,0 +1,126 @@
...
+  auto executeResult = filter.execute(dataGraph, args);
+  REQUIRE(executeResult.result.valid());
+
+  // Check the results
+  float d2r = complex::numbers::pi / 180.0f;
+  for(size_t t = 0; t < tupleShape[0]; t++)
+  {
+    for(size_t c = 0; c < componentShape[0]; c++)
+    {
+      REQUIRE((*angles)[t * componentShape[0] + c] == static_cast<float>(t * c) * d2r);
```
JDuffeyBQ, line 76: We should typically keep `REQUIRE` out of loops since it can be much slower. Though since there are only 30 elements in this case, it shouldn't have much of an impact.

src/Plugins/ComplexCore/src/ComplexCore/Filters/ChangeAngleRepresentation.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <cmath>
+#include <fmt/format.h>
+
+using namespace complex;
+
+namespace EulerAngleConversionType
+{
+constexpr uint64 DegreesToRadians = 0;
+constexpr uint64 RadiansToDegrees = 1;
+} // namespace EulerAngleConversionType
```
JDuffeyBQ, line 20:
```suggestion
namespace
{
namespace EulerAngleConversionType
{
constexpr uint64 DegreesToRadians = 0;
constexpr uint64 RadiansToDegrees = 1;
} // namespace EulerAngleConversionType
} // namespace
```

JDuffeyBQ (changes requested):

src/Plugins/ComplexCore/test/ChangeAngleRepresentationTest.cpp
imikejackson: Exactly. I'm taking this case by case which is why I left it in the loop.

JDuffeyBQ (changes requested):
In addition to the other comments from before, the PR title doesn't reference the addition of `ChangeAngleRepresentation` and some changes don't seem related.

JDuffeyBQ:
Actually I just saw the other PR. This one still contains duplicate code.

PR #136, opened 2022-02-23
Added itk as optional vcpkg feature (not required by default)

JDuffeyBQ:
* Primarily used for the plugin ITKImageProcessing

### DIFF
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -13,6 +13,7 @@
         "fmt",
         "h5support",
         "hdf5",
+        "itk",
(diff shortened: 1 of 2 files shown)


imikejackson:
Are there instructions in a ReadMe.md file on how to invoke cmake with the proper arguments to build `complex` both with and without options. There are options so far for:

* Testing. 
* TBB
* ITK

JDuffeyBQ:
No but we can put that in a separate PR.

imikejackson (approved):

PR #152, opened 2022-03-14
BUG FIX: Legacy D3D array dimensions are now read correctly.

joeykleingers:


joeykleingers:
Found a bug where legacy D3D arrays were not being read correctly because the tuple and component dimensions were being read by adding the tuple dimensional sizes together and returning a usize instead of reading them in as a vector of usize.

joeykleingers:
This caused the Read DREAM3D File (v8) filter to fail because the DataStore tuple and component counts were not equal to the actual sizes from the file.

mmarineBlueQuartz (approved):

JDuffeyBQ (approved):

PR #170, opened 2022-03-25
Feature/multi-DREAM.3D import

mmarineBlueQuartz:
* Fixed importing multiple .dream3d files overwriting existing DataObjects.
* Imported DataObjects have their IDs reset and existing references updated to accommodate the updated values.

### DIFF
--- a/src/complex/DataStructure/DataStructure.cpp
+++ b/src/complex/DataStructure/DataStructure.cpp
@@ -517,6 +517,16 @@ DataStructure::ConstIterator DataStructure::end() const
 
 bool DataStructure::insert(const std::shared_ptr<DataObject>& dataObject, const DataPath& dataPath)
 {
+  if(dataObject == nullptr)
+  {
+    return false;
+  }
+
+  if(getData(dataObject->getId()) != nullptr)
+  {
+    dataObject->setId(generateId());
+  }
+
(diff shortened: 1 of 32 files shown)


src/complex/DataStructure/DataStructure.cpp
```diff
@@ -517,6 +517,16 @@ DataStructure::ConstIterator DataStructure::end() const
...
 {
+  if(dataObject == nullptr)
+  {
+    return false;
+  }
+
+  if(getData(dataObject->getId()) != nullptr)
+  {
+    dataObject->setId(generateId());
+  }
```
JDuffeyBQ, line 528: If the object is already in the `DataStructure`, shouldn't it already have an id?

JDuffeyBQ (commented):
It would be good to have a test case for importing multiple DREAM3D imports.

src/complex/DataStructure/DataStructure.cpp
JDuffeyBQ: What about adding an existing `DataObject` to a new path so that it would have multiple parents? It would overwrite the ID here.

imikejackson (approved):

JDuffeyBQ (approved):

PR #180, opened 2022-04-04
Parameters: Add API to retrieve the active value for a linked parameter.

imikejackson:
This allows the gui side to get the specific value of a linked parameter that
would make the target parameter 'active'

Also update ExampleFilter2 to use linked parameters.


JDuffeyBQ (changes requested):
ExampleFilter2 has some extra changes.

JDuffeyBQ (approved):

PR #184, opened 2022-04-07
Removed unused value from DynamicTableParameter

JDuffeyBQ:


imikejackson:
What happens if someone uses a DynamicTableParameter but it is an optional parameter? In fact I have that exact case in ApplyTransformationToGeometry. https://github.com/BlueQuartzSoftware/complex/pull/181  In that filter we have a fixed 4x4 table so in reality it can _never_ be empty.

JDuffeyBQ:
`DynamicTableParameter` didn't use that value at all. And in fact performs no validation of the incoming data at all. So rather than checking empty, the parameter should check the size of the incoming data and if it's supposed to have a fixed size.

Also I think `DynamicTableData` should be split into the description of the table and the actual data. Then the parameter will only accept a vector of vectors.

imikejackson:
If the split of values and description is needed, then go ahead and make the split. Please be sure to validate those changes against DREAM3DNX.

PR #194, opened 2022-04-08
Add numVerts static const to Vertex & Tetrahedral Geometries

jmarquisbq:


src/complex/DataStructure/Geometry/TetrahedralGeom.hpp
```diff
@@ -15,6 +15,8 @@ class COMPLEX_EXPORT TetrahedralGeom : public AbstractGeometry3D
 public:
   friend class DataStructure;
 
+  static const int k_NumVerts = 4;
```
mmarineBlueQuartz, line 18:
```suggestion
  static inline constexpr usize k_NumVerts = 4;
```
I don't think these are supposed to be different data types.

mmarineBlueQuartz (changes requested):

imikejackson (approved):

mmarineBlueQuartz (approved):

PR #195, opened 2022-04-09
DataStructure::getId(..). Check for nullptr before using raw pointer.

imikejackson:
Also added unit test to ensure it stays working

### DIFF
--- a/src/complex/DataStructure/DataStructure.cpp
+++ b/src/complex/DataStructure/DataStructure.cpp
@@ -112,7 +112,12 @@ std::optional<DataObject::IdType> DataStructure::getId(const DataPath& path) con
   {
     return {0};
   }
-  return getData(path)->getId();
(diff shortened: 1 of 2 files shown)


src/complex/DataStructure/DataStructure.cpp
```diff
@@ -112,7 +112,12 @@ std::optional<DataObject::IdType> DataStructure::getId(const DataPath& path) con
   {
     return {0};
   }
-  return getData(path)->getId();
+  const DataObject* dataObject = getData(path);
+  if(nullptr == dataObject)
+  {
+    return {0};
```
mmarineBlueQuartz, line 118:
```suggestion
    return std::nullopt;
```

mmarineBlueQuartz (changes requested):

mmarineBlueQuartz (approved):

PR #215, opened 2022-04-29
Add helper method that returns all data object ids in the data structure.

joeykleingers:


src/complex/DataStructure/DataStructure.cpp
```diff
@@ -201,6 +201,16 @@ std::vector<DataPath> DataStructure::getAllDataPaths() const
   return dataPaths;
 }
 
+std::vector<DataObject::IdType> DataStructure::getAllDataObjectIds() const
+{
+  std::vector<DataObject::IdType> dataIds;
```
JDuffeyBQ, line 206:
```suggestion
  std::vector<DataObject::IdType> dataIds;
  dataIds.reserve(m_DataObjects.size());
```
Since we know the size, we can reserve ahead of time.

JDuffeyBQ (changes requested):

PR #220, opened 2022-05-05
Added tuple and component methods to AbstractDataStore

mmarineBlueQuartz:
* Added methods to set tuple values.
* Added methods to set a specific component value.
* Added method to get a specific component value.

### DIFF
--- a/src/complex/DataStructure/AbstractDataStore.hpp
+++ b/src/complex/DataStructure/AbstractDataStore.hpp
@@ -4,7 +4,7 @@
 #include "complex/DataStructure/IDataStore.hpp"
 #include "complex/Utilities/Parsing/HDF5/H5.hpp"
 
-#include "nonstd/span.hpp"
+#include <nonstd/span.hpp>
 
 #include <algorithm>
 #include <functional>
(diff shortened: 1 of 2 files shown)


src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -4,6 +4,8 @@
 #include "complex/DataStructure/IDataStore.hpp"
 #include "complex/Utilities/Parsing/HDF5/H5.hpp"
 
+#include "nonstd/span.hpp"
```
JDuffeyBQ, line 7:
```suggestion
#include <nonstd/span.hpp>
```

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+   * @param values
+   */
+  void setTuple(index_type tupleIndex, value_type* values)
+  {
+    if(values == nullptr)
+    {
+      return;
+    }
+
+    if(tupleIndex > getNumberOfTuples())
```
imikejackson, line 534:
```suggestion
    if(tupleIndex >= getNumberOfTuples())
```

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+
+  /**
+   * @brief Sets all component values for a tuple using a span of values.
+   *
+   * If the tuple index is out of bounds or the provided span does not match
+   * the number of components, this method does nothing.
+   * @param tupleIndex
+   * @param values
+   */
+  void setTuple(index_type tupleIndex, const nonstd::span<value_type>& values)
```
JDuffeyBQ, line 583:
```suggestion
  void setTuple(index_type tupleIndex, nonstd::span<const value_type> values)
```
We really only need span since a span can be constructed from a pointer, vector, etc.

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+    }
+
+    index_type index = tupleIndex * getNumberOfComponents() + componentIndex;
+    setValue(index, value);
+  }
+
+  /**
+   * @brief Returns the component value at the specified tuple and component index.
+   *
+   * This method returns the default T value if either index is out of bounds.
```
JDuffeyBQ, line 630: I think we should fail somehow instead of returning a default value.

JDuffeyBQ (changes requested):

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+    }
+
+    if(tupleIndex > getNumberOfTuples())
+    {
+      return;
+    }
+
+    index_type numComponents = getNumberOfComponents();
+    index_type offset = tupleIndex * numComponents;
+    for(index_type i = 0; i < numComponents; i++)
```
imikejackson, line 569: maybe std::copy for this one also? or memcpy()

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+   * @param values
+   */
+  void setTuple(index_type tupleIndex, const nonstd::span<value_type>& values)
+  {
+    if(values.size() != getNumberOfComponents())
+    {
+      return;
+    }
+
+    if(tupleIndex > getNumberOfTuples())
```
imikejackson, line 590:
```suggestion
    if(tupleIndex >= getNumberOfTuples())
```

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+   * @return value_type
+   */
+  value_type getComponentValue(index_type tupleIndex, index_type componentIndex) const
+  {
+    if(tupleIndex > getNumberOfTuples())
+    {
+      return {};
+    }
+
+    if(componentIndex > getNumberOfComponents())
```
imikejackson, line 642:
```suggestion
    if(componentIndex >= getNumberOfComponents())
```

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+    {
+      return {};
```
imikejackson, line 644: std::runtime_exception if they are out of range.

imikejackson (changes requested):

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+    if(tupleIndex > getNumberOfTuples())
+    {
+      return;
+    }
+
+    index_type numComponents = getNumberOfComponents();
+    index_type offset = tupleIndex * numComponents;
+    for(index_type i = 0; i < numComponents; i++)
+    {
+      setValue(offset + i, values[i]);
```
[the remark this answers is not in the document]
JDuffeyBQ: Not in `AbstractDataStore`.

src/complex/DataStructure/AbstractDataStore.hpp
```diff
@@ -506,12 +509,145 @@ class AbstractDataStore : public IDataStore
...
+    if(tupleIndex > getNumberOfTuples())
+    {
+      return;
+    }
+
+    index_type numComponents = getNumberOfComponents();
+    index_type offset = tupleIndex * numComponents;
+    for(uint64 i = 0; i < numComponents; i++)
+    {
+      setValue(offset + i, values[i]);
```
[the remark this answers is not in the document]
mmarineBlueQuartz: memcpy is impossible as AbstractDataStore does not have access to the container. std::copy inserts values, which would change the size of the container and offset all proceeding values. The only option for AbstractDataStore is to call virtual methods for the derived class to handle.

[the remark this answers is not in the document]
JDuffeyBQ: I don't think `std::copy` inserts values.

src/complex/DataStructure/AbstractDataStore.hpp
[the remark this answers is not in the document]
imikejackson: true

imikejackson (approved):

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #224, opened 2022-05-09
Ensure file extensions are always compared as lower case strings.

imikejackson:

### DIFF
--- a/src/complex/Parameters/FileSystemPathParameter.cpp
+++ b/src/complex/Parameters/FileSystemPathParameter.cpp
@@ -7,6 +7,7 @@
 #include <nlohmann/json.hpp>
 
 #include "complex/Common/Any.hpp"
+#include "complex/Utilities/StringUtilities.hpp"
 
 namespace fs = std::filesystem;
 
@@ -166,9 +167,27 @@ Result<> FileSystemPathParameter::validatePath(const ValueType& path) const
       return {nonstd::make_unexpected(std::vector<Error>{{-2, "File System Path must include a file extension"}})};
     }
 
-    if(path.has_extension() && !m_AvailableExtensions.empty() && m_AvailableExtensions.find(path.extension().string()) == m_AvailableExtensions.end())
+    if(path.has_extension() && !m_AvailableExtensions.empty())
     {
-      return {nonstd::make_unexpected(std::vector<Error>{{-3, fmt::format("File extension '{}' is not a valid file extension", path.extension().string())}})};
+      bool validExtension = false;
+      for(const auto& ext : m_AvailableExtensions)
+      {
+        std::string inputExt = complex::StringUtilities::toLower(path.extension().string());
+        std::string compareExt = complex::StringUtilities::toLower(ext);
+        if(compareExt.at(0) == '.' && inputExt[0] != '.')
+        {
+          compareExt = complex::StringUtilities::chopr(compareExt, 1);
+        }
+        if(compareExt == inputExt)
+        {
+          validExtension = true;
+          break;
+        }
+      }
+      if(!validExtension)
+      {
+        return {nonstd::make_unexpected(std::vector<Error>{{-3, fmt::format("File extension '{}' is not a valid file extension", path.extension().string())}})};
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/complex/Parameters/FileSystemPathParameter.cpp
```diff
@@ -166,9 +167,27 @@ Result<> FileSystemPathParameter::validatePath(const ValueType& path) const
...
-      return {nonstd::make_unexpected(std::vector<Error>{{-3, fmt::format("File extension '{}' is not a valid file extension", path.extension().string())}})};
+      bool validExtension = false;
+      for(const auto& ext : m_AvailableExtensions)
+      {
+        std::string inputExt = complex::StringUtilities::toLower(path.extension().string());
+        std::string compareExt = complex::StringUtilities::toLower(ext);
+        if(compareExt.at(0) == '.' && inputExt[0] != '.')
+        {
+          compareExt = complex::StringUtilities::chopr(compareExt, 1);
+        }
```
JDuffeyBQ, line 180: `std::filesystem::path::extension()` always includes the `.` so instead we could force the available extension to have it in the constructor as well as making them lowercase to avoid doing it in every `validate` call. Then this loop can be:
```cpp
std::string inputExt = StringUtilities::toLower(path.extension().string());
bool validExtension = std::find(m_AvailableExtensions.cbegin(), m_AvailableExtensions.cend(), inputExt) != m_AvailableExtensions.cend()
```

Additionally while checking the documentation for this, I noticed that it behaves like `QFileInfo::suffix()` not `QFileInfo::completeSuffix()`. I don't know if this makes a difference in any code we have, but I thought it was important to mention just in case.

src/complex/Utilities/StringUtilities.hpp
```diff
@@ -193,5 +199,20 @@ inline std::string simplified(std::string_view text)
...
+  std::for_each(data.begin(), data.end(), [](char& c) { c = ::toupper(c); });
+  return data;
+}
+
+inline std::string toLower(const std::string& input)
+{
+  std::string data = input;
+  std::for_each(data.begin(), data.end(), [](char& c) { c = ::tolower(c); });
+  return data;
+}
```
JDuffeyBQ, line 215:
```cpp
/**
 * @brief Converts a char to its uppercase version according to the current locale.
 * Uses std::toupper internally.
 * @param c
 * @return char
 */
inline char toUpper(char c)
{
  return static_cast<char>(std::toupper(static_cast<unsigned char>(c)));
}

/**
 * @brief Converts a char to its lowercase version according to the current locale.
 * Uses std::tolower internally.
 * @param c
 * @return char
 */
inline char toLower(char c)
{
  return static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
}

/**
 * @brief Converts the given string to uppercase.
 * Assumes that the string is ASCII.
 * @param input
 * @return std::string
 */
inline std::string toUpper(std::string input)
{
  for(char& c : input)
  {
    c = toUpper(c);
  }
  return input;
}

/**
 * @brief Converts the given string to lowercase.
 * Assumes that the string is ASCII.
 * @param input
 * @return std::string
 */
inline std::string toUpper(std::string input)
{
  for(char& c : input)
  {
    c = toLower(c);
  }
  return input;
}
```
I think making our own wrapper for `std::toupper` and `std::tolower` makes things cleaner with less casting. And with range based for loops, we don't have to include `<algorithm>` for `std::for_each`.

JDuffeyBQ (changes requested):

src/complex/Utilities/StringUtilities.hpp
imikejackson: Changes have been addressed in latest commits

src/complex/Parameters/FileSystemPathParameter.cpp
```diff
@@ -78,6 +79,19 @@ FileSystemPathParameter::FileSystemPathParameter(const std::string& name, const
 , m_AvailableExtensions(extensionsType)
 , m_ShouldValidateExtension(shouldValidateExtension)
 {
+  ExtensionsType validatedExtensions;
+  for(const auto& ext : m_AvailableExtensions)
+  {
+    if(ext.at(0) != '.')
```
JDuffeyBQ, line 85: Should we also add a check for empty extensions here?

JDuffeyBQ (approved):

PR #230, opened 2022-05-24
Updated pipeline node json

mmarineBlueQuartz:
* Moved node-specific json writing to toJsonImpl method defined in AbstractPipelineNode.
* AbstractPipelineNode::toJson() calls toJsonImpl() and adds the disabled state before returning the result.
* Creating a PipelineFilter or Pipeline object from json now sets the disabled state according to the provided json. If the state was not set in the provided json, the node is enabled by default.

### DIFF
--- a/src/complex/Pipeline/AbstractPipelineNode.cpp
+++ b/src/complex/Pipeline/AbstractPipelineNode.cpp
@@ -303,7 +303,7 @@ nlohmann::json AbstractPipelineNode::toJson() const
 
 bool AbstractPipelineNode::ReadDisabledState(const nlohmann::json& json)
 {
-  if(json.contains(k_IsDisabledKey.str()) == false)
+  if(!json.contains(k_IsDisabledKey.str()))
   {
     return false;
   }


src/complex/Pipeline/AbstractPipelineNode.cpp
```diff
@@ -287,3 +293,19 @@ void AbstractPipelineNode::sendCancelledMessage()
...
+nlohmann::json AbstractPipelineNode::toJson() const
+{
+  auto json = toJsonImpl();
+  json[k_IsDisabledKey] = m_IsDisabled;
+  return json;
+}
+
+bool AbstractPipelineNode::ReadDisabledState(const nlohmann::json& json)
+{
+  if(json.contains(k_IsDisabledKey.str()) == false)
```
JDuffeyBQ, line 306:
```suggestion
  if(!json.contains(k_IsDisabledKey.str()))
```

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #231, opened 2022-05-27
Move SegmentFeatures base class into complex

imikejackson:
This was done so that other filters can use this base class when they are stored in other plugins


src/Plugins/ComplexCore/test/RotateSampleRefFrameTest.cpp
```diff
@@ -1,219 +0,0 @@
-#include <catch2/catch.hpp>
```
JDuffeyBQ, line 1: Is this supposed to be here?

jmarquisbq (changes requested):

src/Plugins/ComplexCore/test/RotateSampleRefFrameTest.cpp
imikejackson: The file has been completely removed.

src/Plugins/ComplexCore/src/ComplexCore/Filters/ScalarSegmentFeaturesFilter.cpp
```diff
@@ -151,9 +154,12 @@ IFilter::PreflightResult ScalarSegmentFeaturesFilter::preflightImpl(const DataSt
     dataPaths.push_back(goodVoxelsPath);
   }
 
-  data.validateNumberOfTuples(dataPaths);
+  if(!dataStructure.validateNumberOfTuples(dataPaths))
+  {
+    return {nonstd::make_unexpected(std::vector<Error>{Error{-651, fmt::format("Input arrays do have matching tuple counts.")}})};
```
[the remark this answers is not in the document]
imikejackson: Opps. Fixed in latest push.

JDuffeyBQ:
If moving `RotateSampleRefFrame` is intended as part of this PR it should be documented.

PR #237, opened 2022-06-14
Filter/import hdf5 dataset

jmarquisbq:
- Add ImportHDF5DatasetParameter class
- Add ImportHDF5Dataset filter to the ComplexCore plugin
- Add ImportHDF5Dataset filter unit test

- Add H5::Type to complex::DataType convenience function to H5::DatasetReader
- Fix H5::DatasetReader::readIntoSpan function template definitions/exports so that it can be used by plugins
- Differences from SIMPL's version of this filter: I've added an additional tuple dimensions input for the parameter that can be used along side the component dimensions in lieu of the AttributeMatrix dimension restrictions.  Once we have the AttributeMatrix back, the intention is that the user can give either the tuple dimensions or an AttributeMatrix.


src/Plugins/ComplexCore/test/ImportHDF5DatasetTest.cpp
```diff
@@ -0,0 +1,674 @@
...
+  {
+    cDimsStr.append(StringUtilities::number(vector[i]));
+    if(i < vector.size() - 1)
+    {
+      cDimsStr.append(" " + separator + " ");
+    }
+  }
+
+  return cDimsStr;
+}
```
JDuffeyBQ, line 439:
```cpp
fmt::format("{}", fmt::join(vector, ", "));
```

src/Plugins/ComplexCore/test/ImportHDF5DatasetTest.cpp
```diff
@@ -0,0 +1,674 @@
...
+      DatasetTest<int64_t>(filter, importInfoList, resultsValid);
+      DatasetTest<uint64_t>(filter, importInfoList, resultsValid);
+      DatasetTest<float32>(filter, importInfoList, resultsValid);
+      DatasetTest<float64>(filter, importInfoList, resultsValid);
+    }
+  }
+
+  if(fs::exists(m_FilePath))
+  {
+    if(fs::remove(m_FilePath) == false)
```
JDuffeyBQ, line 669:
```suggestion
    if(!fs::remove(m_FilePath))
```

JDuffeyBQ (changes requested):

PR #243, opened 2022-06-24
Changed the FilterHandle to take in filter as a const ref instead of raw pointer

bpenniebq:


src/complex/Filter/FilterHandle.hpp
```diff
@@ -112,11 +119,12 @@ class COMPLEX_EXPORT FilterHandle
    * @param filterId
```
imikejackson, line 119:
```suggestion
   * @param filter The IFilter object that the handle wraps
```

imikejackson (approved):

imikejackson (approved):

PR #244, opened 2022-06-24
Topic/update h5 wrapper api

jmarquisbq:
- Add COMPLEX_EXPORT to AttributeReader template definitions for use outside of complex
- Add a getClassType function to H5DatasetReader and H5AttributeReader for returning the H5T_class_t enum value as an int64 type
- Add getObjectId function to H5ObjectReader for accessing the object's Id address
- Fix DatasetReader::getDimensions fct alg when type is a string ( DatasetReader::getType will currently never be able to return a string )
- Fix ObjectReader::getAttributeNames fct to actually populate the vector with the attribute names before returning
- Simplify id data member storage in the H5*[Reader/Writer] classes by using a single id data member in the ObjectReader class

### DIFF
--- a/src/complex/Utilities/Parsing/HDF5/H5DatasetWriter.hpp
+++ b/src/complex/Utilities/Parsing/HDF5/H5DatasetWriter.hpp
@@ -43,13 +43,6 @@ public:
   bool isValid() const override;
 
   /**
-   * @brief Returns the dataset's HDF5 ID. Until one of the write* methods are
-   * called, this method returns 0.
-   * @return H5::IdType
-   */
-  H5::IdType getId() const override;
-
-  /**
    * @brief Returns the target dataset name. Returns an empty string if the
    * DatasetWriter is invalid.
    * @return std::string
@@ -109,8 +102,8 @@ public:
     //  return -1;
     //}
     /* Open the object */
-    // m_DatasetId = H5Dcreate(getParentId(), getName(), dataType, dataspaceId, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
-    // if(m_DatasetId < 0)
+    // setId(H5Dcreate(getParentId(), getName(), dataType, dataspaceId, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT));
+    // if(getId() < 0)
     //{
     //  std::cout << "Error opening Object for Attribute operations." << std::endl;
     //  return -1;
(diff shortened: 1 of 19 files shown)


src/complex/Utilities/Parsing/HDF5/H5DatasetWriter.hpp
```diff
@@ -185,19 +178,18 @@ class COMPLEX_EXPORT DatasetWriter : public ObjectWriter
    */
   void createOrOpenDataset(H5::IdType typeId, H5::IdType dataspaceId);
 
+  /**
+   * @brief Closes the HDF5 dataset and resets the ID to 0.
+   */
+  virtual void closeHdf5() override;
```
mmarineBlueQuartz, line 184:
```suggestion
  void closeHdf5() override;
```
You do not need virtual and override on the same method.

src/complex/Utilities/Parsing/HDF5/H5FileWriter.hpp
```diff
@@ -78,10 +72,7 @@ class COMPLEX_EXPORT FileWriter : public GroupWriter
   /**
    * @brief Closes the HDF5 ID and resets it to 0.
    */
-  void closeHdf5();
-
-private:
-  H5::IdType m_FileId = 0;
+  virtual void closeHdf5() override;
```
mmarineBlueQuartz, line 75:
```suggestion
  void closeHdf5() override;
```

src/complex/Utilities/Parsing/HDF5/H5GroupWriter.hpp
```diff
@@ -74,10 +68,16 @@ class COMPLEX_EXPORT GroupWriter : public ObjectWriter
   /**
    * @brief Closes the HDF5 ID and resets it to 0.
    */
-  void closeHdf5();
+  virtual void closeHdf5() override;
```
mmarineBlueQuartz, line 71:
```suggestion
  void closeHdf5() override;
```

src/complex/Utilities/Parsing/HDF5/H5ObjectReader.cpp
```diff
@@ -43,9 +49,26 @@ H5::IdType H5::ObjectReader::getParentId() const
   return m_ParentId;
 }
 
+size_t H5::ObjectReader::getObjectId() const
```
mmarineBlueQuartz, line 52:
```suggestion
haddr_t H5::ObjectReader::getObjectId() const
```
size_t changes it's underlying type across systems. Stick to the HDF5 type or create an alias that does the same. This also protects against HDF5 changing the type alias. I vaguely remember happening that happening with herr_t or hid_t several years ago when we were using those types interchangeably. Additionally, haddr_t documents the return value better than stating it as a size.

src/complex/Utilities/Parsing/HDF5/H5ObjectReader.hpp
```diff
@@ -45,11 +45,23 @@ class COMPLEX_EXPORT ObjectReader
    */
   H5::IdType getParentId() const;
 
+  /**
+   * @brief Returns the object's address ID. Returns 0 if the object is invalid.
+   * @return H5::IdType
+   */
+  size_t getObjectId() const;
```
mmarineBlueQuartz, line 52:
```suggestion
  haddr_t getObjectId() const;
```

mmarineBlueQuartz (changes requested):

mmarineBlueQuartz (approved):

PR #246, opened 2022-07-02
[DataArray] Fix copyTuple bug.

imikejackson:
Multicomponent data arrays were not taken care of correctly.

### DIFF
--- a/test/DataArrayTest.cpp
+++ b/test/DataArrayTest.cpp
@@ -8,6 +8,7 @@
 #include "complex/DataStructure/DataGroup.hpp"
 #include "complex/DataStructure/DataStore.hpp"
 #include "complex/DataStructure/DataStructure.hpp"
+#include "complex/Utilities/DataArrayUtilities.hpp"
 
 using namespace complex;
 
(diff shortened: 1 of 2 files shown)


test/DataArrayTest.cpp
```diff
@@ -21,6 +22,38 @@ TEST_CASE("DataArrayCreation")
   REQUIRE(numTuples == 0);
 }
 
+TEST_CASE("complex::DataArray Copy TupleTest", "[complex][DataArray]")
+{
+  DataStructure dataStructure;
+  IDataStore::ShapeType tupleShape{5};
+  IDataStore::ShapeType componentShape{3};
+  Result<> result = CreateArray<int32>(dataStructure, tupleShape, componentShape, DataPath({"DataArray"}), IDataAction::Mode::Execute);
+  DataArray<int32>& dataArray = ArrayRefFromPath<int32>(dataStructure, DataPath({"DataArray"}));
```
JDuffeyBQ, line 31:
```suggestion
  auto& dataArray = dataStructure.getDataRefAs<DataArray<int32>>(DataPath({"DataArray"}));
```

test/DataArrayTest.cpp
```diff
@@ -21,6 +22,38 @@ TEST_CASE("DataArrayCreation")
   REQUIRE(numTuples == 0);
 }
 
+TEST_CASE("complex::DataArray Copy TupleTest", "[complex][DataArray]")
+{
+  DataStructure dataStructure;
+  IDataStore::ShapeType tupleShape{5};
+  IDataStore::ShapeType componentShape{3};
+  Result<> result = CreateArray<int32>(dataStructure, tupleShape, componentShape, DataPath({"DataArray"}), IDataAction::Mode::Execute);
```
JDuffeyBQ, line 30: Should this result be checked? Or would `DataArray<int32>::Create` be better here?

JDuffeyBQ, line 30: This `DataPath` should be a constant used everywhere in the test case.

test/DataArrayTest.cpp
```diff
@@ -21,6 +22,38 @@ TEST_CASE("DataArrayCreation")
...
+  DataArray<int32>& dataArray = ArrayRefFromPath<int32>(dataStructure, DataPath({"DataArray"}));
+
+  for(size_t i = 0; i < 5; i++)
```
JDuffeyBQ, line 33: The loops should use the values from the tuple/component shapes instead of being hardcoded.

test/DataArrayTest.cpp
imikejackson: Done

PR #248, opened 2022-07-05
Allow ImportHDF5DataSet to create data sets at the top level of the DataStructure.

imikejackson:


imikejackson:
#251  takes over from this PR.

PR #260, opened 2022-07-21
Improve FindNeighborListStatistics unit test

jmarquisbq:


mmarineBlueQuartz (approved):
I'm surprised the float equality checks did not require an EPSILON comparison, but if the tests pass, so be it.

jmarquisbq:
@mmarineBlueQuartz Yeah I was thinking they would as well.  I'm guessing it's because I used results that were previously run through the filter that we were confident in rather than putting in a manually calculated value.

jmarquisbq:
@mmarineBlueQuartz Perhaps a small EPSILON wouldn't be a bad idea anyway?

mmarineBlueQuartz (approved):

PR #263, opened 2022-07-24
Ensure all cases of shadowed variables are picked up.

imikejackson:

### DIFF
--- a/cmake/Utility.cmake
+++ b/cmake/Utility.cmake
@@ -52,21 +52,21 @@ function(complex_enable_warnings)
         /we4458 # C4458: declaration of 'identifier' hides class member
         /we4459 # C4459: declaration of 'identifier' hides global declaration
     )
-  elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  else()
+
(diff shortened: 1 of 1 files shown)


cmake/Utility.cmake
```diff
@@ -52,13 +52,21 @@ function(complex_enable_warnings)
         /we4458 # C4458: declaration of 'identifier' hides class member
         /we4459 # C4459: declaration of 'identifier' hides global declaration
     )
-  else()
+  elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
JDuffeyBQ, line 55: Since clang and gcc are mostly the same, we can move the check:
```cmake
if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
  set(SHADOW_WARNING "shadow")
else()
  set(SHADOW_WARNING "shadow-all")
endif()

...

-Werror=${SHADOW_WARNING}
```

JDuffeyBQ (changes requested):

cmake/Utility.cmake
imikejackson: Done

JDuffeyBQ (approved):

PR #275, opened 2022-08-07
Filter and Unit Test Fixes

imikejackson:


src/Plugins/ComplexCore/test/QuickSurfaceMeshFilterTest.cpp
```diff
@@ -80,25 +35,25 @@ TEST_CASE("ComplexCore::QuickSurfaceMeshFilter", "[SurfaceMeshing][QuickSurfaceM
...
     args.insertOrAssign(QuickSurfaceMeshFilter::k_FeatureIdsArrayPath_Key, std::make_any<DataPath>(featureIdsDataPath));
 
-    MultiArraySelectionParameter::ValueType selectedArrayPaths = {ebsdSanDataPath.createChildPath(k_ConfidenceIndex), ebsdSanDataPath.createChildPath(k_IpfColors),
+    MultiArraySelectionParameter::ValueType selectedArrayPaths = {ebsdSanDataPath.createChildPath(k_ConfidenceIndex), ebsdSanDataPath.createChildPath(k_IPFColors),
                                                                   ebsdSanDataPath.createChildPath(k_ImageQuality), ebsdSanDataPath.createChildPath(k_Phases)};
 
     args.insertOrAssign(QuickSurfaceMeshFilter::k_SelectedDataArrayPaths_Key, std::make_any<MultiArraySelectionParameter::ValueType>(selectedArrayPaths));
 
-    args.insertOrAssign(QuickSurfaceMeshFilter::k_ParentDataGroupPath_Key, std::make_any<DataPath>(triangleParentGroup));
+    // args.insertOrAssign(QuickSurfaceMeshFilter::k_ParentDataGroupPath_Key, std::make_any<DataPath>({}));
```
jmarquisbq, line 56:
```suggestion
```

src/complex/Common/Constants.hpp
```diff
@@ -0,0 +1,190 @@
...
+
+inline constexpr float k_Cos_OneEighthPiF = k_Cos_OneEighthPi<float>;
+inline constexpr double k_Cos_OneEighthPiD = k_Cos_OneEighthPi<double>;
+
+inline constexpr float k_Cos_ThreeEighthPiF = k_Cos_ThreeEighthPi<float>;
+inline constexpr double k_Cos_ThreeEighthPiD = k_Cos_ThreeEighthPi<double>;
+
+inline constexpr float k_Sin_ThreeEighthPiF = k_Sin_ThreeEighthPi<float>;
+inline constexpr double k_Sin_ThreeEighthPiD = k_Sin_ThreeEighthPi<double>;
+} // namespace complex::Constants
```
jmarquisbq, line 190: @imikejackson Is there a reason we didn't put these all in Numbers.hpp?  Or at least the numbers namespace if not the file itself? I feel like if I were looking for these I would not think to check for them here if there was a file called Numbers.hpp already.

jmarquisbq (changes requested):

src/complex/Common/Constants.hpp
imikejackson: The Numbers.hpp is a replacement for a C++20 header so our thinking is that when we move to C++20 and we get Numbers.hpp for free we don't have to then move a bunch of constants to a new file. Doing this now alleviates that need later on.

jmarquisbq (dismissed):

src/Plugins/ComplexCore/test/CreateFeatureArrayFromElementArrayTest.cpp
```diff
@@ -37,77 +34,62 @@ namespace fs = std::filesystem;
...
-const std::string k_FeatureIdsFileName = "FeatureIds.raw";
+const std::string k_FeatureIDs("FeatureIds");
+const std::string k_Computed_CellData("Computed_CellData");
+
+inline DataStructure LoadDataStructure(const fs::path& filepath)
+{
+  DataStructure exemplarDataStructure;
+  REQUIRE(fs::exists(filepath));
+  auto result = DREAM3D::ImportDataStructureFromFile(filepath);
+  REQUIRE(result.valid());
```
JDuffeyBQ, line 45:
```suggestion
  COMPLEX_RESULT_REQUIRE_VALID(result);
```

src/Plugins/ComplexCore/test/MinNeighborsTest.cpp
```diff
@@ -1,125 +1,142 @@
...
-inline const std::vector<usize> k_Dimensions{5, 5, 5};
-inline const std::vector<usize> k_CompShape{1};
 
-inline const std::string k_ImageGeomName = "Image Geom";
-inline const std::string k_FeaturePhasesName = "Feature Phases";
-inline const std::string k_FeatureIdsName = "Feature IDs";
-inline const std::string k_NumNeighborsName = "Num Neighbors";
-inline const std::vector<std::string> k_VoxelArrayNames = {"Array1", "Array2", "Array3"};
-inline const std::vector<DataPath> k_VoxelArrayPaths = {DataPath({k_ImageGeomName, "Array1"}), DataPath({k_ImageGeomName, "Array2"}), DataPath({k_ImageGeomName, "Array3"})};
+std::vector<int32> k_NumElements = {
```
JDuffeyBQ, line 23:
```suggestion
const std::vector<int32> k_NumElements = {
```

src/complex/Utilities/DataGroupUtilities.cpp
```diff
@@ -153,4 +146,27 @@ std::vector<std::shared_ptr<IDataArray>> GenerateDataArrayList(const DataStructu
   return arrays;
 }
 
+std::vector<DataPath> GetAllChildDataPaths(const DataStructure& dataStructure, const DataPath& parentGroup, DataObject::Type dataObjectType)
```
JDuffeyBQ, line 149: What did we decide for this function? Was it going to return an optional? Or just rethrow?

test/UnitTestCommon/include/complex/UnitTest/UnitTestCommon.hpp
```diff
@@ -95,6 +108,119 @@ inline constexpr StringLiteral k_ReducedGeometry("Reduced Geometry");
 
 namespace UnitTest
 {
+
+constexpr float EPSILON = 0.00001;
```
JDuffeyBQ, line 112:
```suggestion
inline constexpr float EPSILON = 0.00001;
```

.github/workflows/linux.yml
```diff
@@ -51,12 +57,20 @@ jobs:
           -name "GitHub"
           -username "BlueQuartzSoftware"
           -password "${{secrets.GITHUB_TOKEN}}"
+      - name: Download Test Files
+        shell: bash
+        run: >
+          wget --quiet "https://github.com/dream3d/DREAM3D_Data/releases/download/v6_7/TestFiles_6_7.tar.gz" -O ${{github.workspace}}/DREAM3D_Data/TestFiles.tar.gz
```
JDuffeyBQ, line 63: Is the data in this zip not already in DREAM3D_Data?

JDuffeyBQ (changes requested):

.github/workflows/linux.yml
imikejackson: It is NOT in DREAM3D_Data. The data is getting large and Git is asking to use Git LFS instead. Using this approach it is easier to just version the data used, IMO anyways. Maybe at some point in the near future we can just post the data to our own "Data Repository" that has enough bandwidth to serve all  the build bots.

src/complex/Utilities/DataGroupUtilities.cpp
imikejackson: Changed to std::optional. Worked better in the filters and is cleaner code.

JDuffeyBQ (approved):

PR #277, opened 2022-08-10
Exemplar data for WriteASCIIDataTest Unit test in complex_plugins

nyoungbq:
Just a bunch of data sets for WriteASCIIDataTest in complex_plugins. Dependency for the upcoming PR on the complex_plugins write_ascii_data branch.


imikejackson:
We should either put this data with the filter or the filter should come over to complex/src/Plugins/ComplexCore

PR #281, opened 2022-08-21
Fix issues where ComplexRange and tbb::blocked_range might be reversed.

imikejackson:
Remove license info since that is at the top of the project.


JDuffeyBQ (changes requested):
I've thought of a couple other things that should be changed in the parallel algorithm classes:
- All the algorithm types have virtual destructors when they don't need it.
- They have the auto partitioner as a member variable, but the auto partitioner is just an empty type. We can just instantiate a copy in the execute function instead.

src/complex/Utilities/ParallelDataAlgorithm.hpp
```diff
@@ -26,7 +26,12 @@ class COMPLEX_EXPORT ParallelDataAlgorithm
   using RangeType = Range;
 
   ParallelDataAlgorithm();
-  virtual ~ParallelDataAlgorithm();
+  ~ParallelDataAlgorithm();
+
+  ParallelDataAlgorithm(const ParallelDataAlgorithm&) = delete;
+  ParallelDataAlgorithm(ParallelDataAlgorithm&&) noexcept = delete;
+  ParallelDataAlgorithm& operator=(const ParallelDataAlgorithm&) = delete;
+  ParallelDataAlgorithm& operator=(ParallelDataAlgorithm&&) noexcept = delete;
```
JDuffeyBQ, line 34: I think all of these can be defaulted. Being copyable/movable should be fine for these classes. Also would we want to rename `ParallelDataAlgorithm` to `ParallelData1DAlgorithm` to be consistent with the other classes?

JDuffeyBQ (changes requested):

PR #282, opened 2022-08-23
Topic/triangle normal filter

nyoungbq:
Ported triangle normal filter with unit test


imikejackson:
Is there a documentation file to go with the filter?

PR #296, opened 2022-09-13
OStream API

nyoungbq:
Draft PR for reviewing API


JDuffeyBQ (changes requested):
* The implementation here should be able to be moved into a cpp file
* Functions should have the Javadoc style documentation comments
* It would be good to have an example usage and output in a test case
* The main issue is whether the usage of `std::map` is thread safe and what the standard says. My guess would be that it's not since we are adding new elements. That might require a different container or maybe pre-allocating elements if possible.

PR #305, opened 2022-09-16
Add Actions for 1DGeometry and StringArray.

imikejackson:
Fix bug when creating StringArray::CreateArray.

### DIFF
--- a/src/complex/Filter/Actions/CreateStringArrayAction.cpp
+++ b/src/complex/Filter/Actions/CreateStringArrayAction.cpp
@@ -1,0 +1,74 @@
+#include "CreateStringArrayAction.hpp"
+
+#include <fmt/core.h>
+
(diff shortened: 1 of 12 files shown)


src/complex/Filter/Actions/CreateStringArrayAction.cpp
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  if(m_Dims.empty())
+  {
+    return MakeErrorResult(-381, fmt::format("CreateStringArrayAction: Tuple Shape was empty. Please set the number of tuples."));
+  }
+
+  usize last = path().getLength() - 1;
+
+  std::string name = path()[last];
```
mmarineBlueQuartz, line 47:
```suggestion
  std::string name = path().getTargetName();
```

mmarineBlueQuartz (changes requested):

PR #315, opened 2022-09-25
Clean up compiler warnings for switch and braced initializers

imikejackson:

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/ApplyTransformationToGeometryFilter.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/ApplyTransformationToGeometryFilter.cpp
@@ -229,6 +229,7 @@ IFilter::PreflightResult ApplyTransformationToGeometryFilter::preflightImpl(cons
   case TransformType::Rotation:
   case TransformType::Translation:
   case TransformType::Scale:
+  case TransformType::ManualTransformMatrix:
     break;
   }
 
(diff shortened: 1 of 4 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/InterpolatePointCloudToRegularGridFilter.cpp
```diff
@@ -100,6 +101,8 @@ void mapPointCloudDataByKernel(IDataArray* source, INeighborList* dynamic, std::
   case DataType::float64:
     mapPointCloudDataByKernel<float64>(source, dynamic, kernelVals, kernel, dims, curX, curY, curZ, vertIdx);
     break;
+  case DataType::boolean:
```
JDuffeyBQ: It looks like preflight just skips bool arrays which is why they make it to execute. Should selecting bool arrays just be a preflight error?

imikejackson: I'm down deep into those pipelines trying to figure out what to do. i'm not really sure how one "interpolates" boolean values so this should indeed be a preflight error if someone selects a bool array to interpolate

imikejackson: I removed this fix because the entire InterpolatePointCloudToRegularGrid needs to be fixed. There are a bunch of issues with the filter.

JDuffeyBQ (approved):

PR #321, opened 2022-09-29
INodeXDGeometry: Fix bug get[Edge|Face|Cell]PointIds() method to properly return out to the end of the array

jmarquisbq:


imikejackson:
INodeGeometry2D::getFacePointIds() has the same issue

INodeGeometry3D::getCellPointIds() does not even do the check.

Could we also update those methods.

imikejackson (approved):

PR #324, opened 2022-10-03
Geometry API: Rename methods to better convey what they do.

imikejackson:
The goal is to cut down on the confusion when writing codes that combine both the complex API and VTK APIs where the word "id" is both used to convey different meanings

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/TriangleDihedralAngleFilter.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/TriangleDihedralAngleFilter.cpp
@@ -171,20 +171,15 @@ IFilter::PreflightResult TriangleDihedralAngleFilter::preflightImpl(const DataSt
 
   const auto* triangleGeom = dataStructure.getDataAs<TriangleGeom>(pTriangleGeometryDataPath);
   // Get the Face AttributeMatrix from the Geometry (It should have been set at construction of the Triangle Geometry)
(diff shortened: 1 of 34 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/TriangleDihedralAngleFilter.cpp
```diff
@@ -171,11 +171,15 @@ IFilter::PreflightResult TriangleDihedralAngleFilter::preflightImpl(const DataSt
 
   const auto* triangleGeom = dataStructure.getDataAs<TriangleGeom>(pTriangleGeometryDataPath);
   // Get the Face AttributeMatrix from the Geometry (It should have been set at construction of the Triangle Geometry)
-  const AttributeMatrix* faceAttributeMatrix = triangleGeom->getFaceData();
+  const AttributeMatrix* faceAttributeMatrix = triangleGeom->getFaceAttributeMatrix();
   if(faceAttributeMatrix == nullptr)
```
jmarquisbq:
```suggestion
  if(triangleGeom == nullptr)
```

jmarquisbq (changes requested):

src/Plugins/ComplexCore/src/ComplexCore/Filters/TriangleDihedralAngleFilter.cpp
imikejackson: Fixed in latest commits

jmarquisbq (approved):

PR #326, opened 2022-10-11
Add a PR template

imikejackson:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -27,23 +27,41 @@ close a related issues using keywords (https://help.github.com/articles/closing-
 @mentions (https://help.github.com/articles/basic-writing-and-formatting-syntax/#mentioning-people-and-teams)
 of the person or team responsible for reviewing proposed changes. -->
 
-## PR Checklist
 
+## Naming Conventions
+
+Naming of variables should descriptive where needed. Loop Control Variables can use `i` if warranted. Most of these conventions are enforced through the clang-tidy and clang-format configuration files.
+
+- [ ] Class and Structs are `UpperCamelCase`
+- [ ] Class private member variables are `m_UpperCamelCase`
+- [ ] Class methods are `lowerCamelCase`
+- [ ] Method arguments are `lowerCamelCase`
+- [ ] Normal variables are `lowerCamelCase`
+- [ ] Constants are `k_UpperCamelCase`
+- [ ] Global statics are `s_UpperCamelCase`
+- [ ] Free Functions are `UpperCamelCase`
+- [ ] Macros are `ALL_UPPER_SNAKE_CASE`
+- [ ] Unit test will test data output from the filter.
+- [ ] Reused strings should be constants in an anonymous namespace
+- [ ] If parallelization is used, proper use of the abstracted complex classes are used. Using TBB specifically in a filter should be frowned upon unless for a really good reason.
+
+## Filter Checklist
+
+- [ ] Parameters should be generally broken down into "Input Parameters", "Required Data Objects", "Created Data Objects". There can be exceptions to this.
+- [ ] ChoicesParameter selections should be an enumeration defined in the filer header
 - [ ] Documentation copied from SIMPL Repo and updated (if necessary)
-- [ ] Parameter argument strings are snake_case
+- [ ] Parameter argument variables are k_CamelCase_Key
+- [ ] Parameter argument strings are lower_snake_case
 ```
 static inline constexpr StringLiteral k_AlignmentType_Key = "alignment_type";
 ```
-- [ ] Naming conventions for classes/structs/variables and such are being followed
-  - [ ] Class and Structs are `UpperCamelCase`
-  - [ ] Class private member variables are `m_UpperCamelCase`
-  - [ ] Class methods are `lowerCamelCase`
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,58 @@
...
+  - [ ] Class and Structs are `UpperCamelCase`
+  - [ ] Class private member variables are `m_UpperCamelCase`
+  - [ ] Class methods are `lowerCamelCase`
+  - [ ] Method arguments are `lowerCamelCase`
+  - [ ] Constants are `k_UpperCamelCase`
+  - [ ] Global statics are `s_UpperCamelCase`
+  - [ ] Macros are `ALL_UPPER_SNAKE_CASE`
+- [ ] Unit test will tests data output from the filter.
+- [ ] If parallelization is used, proper use of the abstracted complex classes are used. Using TBB specifically in a filter should be frowned upon unless for a really good reason.
+- [ ] No commented out code (rare exceptions to this is allowed..)
```
nyoungbq, line 65: require all reused strings to be included in a `constexpr` in an anonymous namespace at the top of a file?

`- [ ] Reused strings should be constants in an anonymous namespace`

.github/pull_request_template.md
```diff
@@ -0,0 +1,58 @@
...
+static inline constexpr StringLiteral k_AlignmentType_Key = "alignment_type";
+```
+- [ ] Naming conventions for classes/structs/variables and such are being followed
+  - [ ] Class and Structs are `UpperCamelCase`
+  - [ ] Class private member variables are `m_UpperCamelCase`
+  - [ ] Class methods are `lowerCamelCase`
+  - [ ] Method arguments are `lowerCamelCase`
+  - [ ] Constants are `k_UpperCamelCase`
+  - [ ] Global statics are `s_UpperCamelCase`
+  - [ ] Macros are `ALL_UPPER_SNAKE_CASE`
```
nyoungbq: Maybe enforce a minimum variable length outside of `for` loop

ex:
`  - [ ] Variables outside of for loops should be at least 3 characters in length`

joeykleingers: I would add template parameters as well to this.  There should be more descriptive names other than "T".

nyoungbq:
To preface this: 
I don't know about this specifically as it is a convention I choose to use, but in general I think that some standards should be enforced on coding syntax for parameters.
For 2 reasons:
1. simplicity of porting and reviewing code
2. easing the barrier to entry for those interested in contributing from outside the company, from experience I can say parameters were one of the most confusing things for me at first and it didn't help to find all sorts of different semantics and syntax for each filter (even if they did the same thing)

Also not entirely sure how everyone feels about this one, what about including something about making an enum (in header for filter file: ExampleFilter.hpp) for choice parameters. (There's definitely a better way this is just how I do it)

Example:
in Filter.hpp
`enum class Delimiter : uint64
  {
    Space = 0,
    Semicolon = 1,
    Comma = 2,
    Colon = 3,
    Tab = 4
  };`
for
` params.insert(std::make_unique<ChoicesParameter>(k_DelimiterChoiceInt_Key, "Delimiter", "Default Delimiter is Comma", to_underlying(Delimiter::Comma), ChoicesParameter::Choices{"Space", "Semicolon", "Comma", "Colon", "Tab"})); // sequence dependent DO NOT REORDER`

nyoungbq:
On that same topic of parameters, I also propose we should enforce a parameter grouping and ordering. I recently found out that this is the preferred ordering:
1. Input Parameters
2. Required Data Objects
3. Created Data Objects

.github/pull_request_template.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] Unit test will tests data output from the filter.
+- [ ] If parallelization is used, proper use of the abstracted complex classes are used. Using TBB specifically in a filter should be frowned upon unless for a really good reason.
+- [ ] No commented out code (rare exceptions to this is allowed..)
+- [ ] Filters should have both the Filter class and Algorithm class for anything beyond trivial needs
+- [ ] No API changes were made (or the changes have been approved)
+- [ ] No major design changes were made (or the changes have been approved)
+- [ ] Added test (or behavior not changed)
```
nyoungbq, line 69: Possibly discuss minimum test case types:
One instantiation test
At least one to test proper functionality
at least one to verify error checking (ie try to break the filter)

jmarquisbq: Yeah I think it would helpful to be a little more explicit if we can on what the unit tests should look like rather than just test the output.  Testing the output needs to be much more than just checking the results.valid() .

.github/pull_request_template.md
```diff
@@ -0,0 +1,58 @@
...
+of the person or team responsible for reviewing proposed changes. -->
+
+## PR Checklist
+
+- [ ] Documentation copied from SIMPL Repo and updated (if necessary)
+- [ ] Parameter argument strings are snake_case
+```
+static inline constexpr StringLiteral k_AlignmentType_Key = "alignment_type";
+```
+- [ ] Naming conventions for classes/structs/variables and such are being followed
```
JDuffeyBQ: We should also add regular variables and free functions.

JDuffeyBQ (approved):

PR #330, opened 2022-10-18
Implement Output API for "Write" Filters

nyoungbq:
Requires a code review

This is the basis for a dependent PR in complex plugins


src/complex/Utilities/StringUtilities.hpp
```diff
@@ -212,6 +212,10 @@ template <typename T>
 inline std::string number(T arg)
 {
   std::stringstream ss;
+  if(std::is_floating_point_v<T>)
```
JDuffeyBQ:
```suggestion
  if constexpr(std::is_floating_point_v<T>)
```
`if constexpr` will make sure this is only generated for floating point types.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+namespace OStreamUtilities
+{
+
+/**
+ * @brief This is the primary datastructure used in print string assembly. It is a wrapper struct for
+ * adding additionaly functionality/readability (underlying datastructure is a std::vector)
+ * @param numrows The number of rows for the matrix
+ * @param numcols The number of columns for the matrix
+ */
+struct PrintMatrix2D
```
JDuffeyBQ:
```suggestion
class PrintMatrix2D
```
Since this contains both public and private members, I'd change this to `class`. There is no difference in the code generated. It's a typical convention to use `struct` for objects that are all public.

Additionally is this part of the public API for these utilities? If not this and the associated classes below, could be moved into the cpp file.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+    ID   |  NumNeighbor  |       Values           ...        //Just an example usage
+  -----------------------------------------------------------
+  | "1"         "2"              "num"           "num"      |  In this case, UNITIALIZED are essentially empty values and will be ignored in
+  | "2"         "1"              "num"      "UNINITAILIZED" |  print if isBalanced bool is false (default)
+  | "3"         "0"         "UNINITAILIZED" "UNINITAILIZED" |  **checkBalanceType() should be run for uniform component arrays to speed up
+  | "4"         "2"              "num"           "num"      |  string assembly, but is not required [Neighbor lists and dataGroups will
+  | ...          ...              ...             ...       |  [Neighbor lists and dataGroups will almost always be unbalanced so its not
+  -----------------------------------------------------------  worth preparsing time, unless you know component count is uniform(Attribute Matrix)]**
+  */
+private:
```
JDuffeyBQ: We tend to put the private data at the bottom of the class and name member variables like `m_Rows`.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  size_t rows, columns;
```
JDuffeyBQ:
```suggestion
  size_t m_Rows = 0;
  size_t m_Columns = 0;
```
Variable declarations are easier to read on multiple lines. Also generally we give member variables default values so we don't access uninitialized data accidentally. This is most important for primitive types because class types will call their default constructor like the `std::vector` below but primitive types do nothing.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+    return columns;
+  }
+  bool getBalance() const
+  {
+    return isBalanced;
+  }
+  /*
+   * This function is used to set balance to true if all values are initialized.
+   */
+  void checkBalanceType() // processor heavy, avoid running if you know its unbalanced
```
JDuffeyBQ: The stored `isBalanced` can get out of sync with the actual state of being balanced if `checkBalanceType` is not called.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  , matrix(numrows * numcols, std::string("UNINITIALIZED"))
+  {
+  }
+  ~PrintMatrix2D() = default;
+
+  PrintMatrix2D(PrintMatrix2D&) = default;
+  PrintMatrix2D(PrintMatrix2D&&) noexcept = default;
+
+  PrintMatrix2D& operator=(const PrintMatrix2D&) = default;
+  PrintMatrix2D& operator=(PrintMatrix2D&&) noexcept = default;
```
JDuffeyBQ: Since these are all default we can actually omit them and have the compiler automatically generate them. Part of it is a stylistic different though we should generally follow the "rule of 3/5/0" https://en.cppreference.com/w/cpp/language/rule_of_three

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  size_t rows, columns;
+  // since its being allocated to heap anyways no significant advantage to array over vector (small sizing difference)
+  std::vector<std::string> matrix;
+  bool isBalanced = false;
+
+public:
+  PrintMatrix2D(size_t numrows, size_t numcols)
+  : rows(numrows)
+  , columns(numcols)
+  , matrix(numrows * numcols, std::string("UNINITIALIZED"))
```
JDuffeyBQ:
```suggestion
  , matrix(numrows * numcols, "UNINITIALIZED")
```
The `std::string` isn't required, and we can directly pass in the string literal.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+ * @tparam T The primitive type attacthed to **DataArray**
+ * @param outputMatrix The PrintMatrix2D that will store string values from *inputArray*
+ * @param inputArray The **DataArray** that will have its values translated into strings
+ * @param columnToWrite The column that the **DataArray** should be stored in
+ * @param hasHeaders bool to determine if index must be advanced 1 to avoid overwrite
+ */
+template <typename T>
+struct LoadDataArrayToMatrixImpl
+{
+  LoadDataArrayToMatrixImpl(PrintMatrix2D& outputMatrix, const DataArray<T>& inputArray, const size_t& columnToWrite, bool hasHeaders = false)
```
JDuffeyBQ:
```suggestion
  LoadDataArrayToMatrixImpl(PrintMatrix2D& outputMatrix, const DataArray<T>& inputArray, size_t columnToWrite, bool hasHeaders = false)
```
Primitives should be passed by value.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  , m_InputArray(inputArray)
+  , m_HasIndex(hasIndex)
+  , m_HasHeader(hasHeader)
+  {
+  }
+  ~LoadNeighborListToMatrixImpl() = default;
+
+  void loadToMatrix(size_t start, size_t end) const // use initial logic checking to avoid repeating ifs in loops
+  {
+    size_t row, column;
```
JDuffeyBQ: Since these variables are only used within the loops, they should be declared/defined in them.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+ * @brief Parallel enabled | implicit loading of **StringArray**'s elements to PrintMatrix2D vertically
+ * @param outputMatrix The PrintMatrix2D that will store string values from *inputArray*
+ * @param inputArray The **StringArray** that will have its values translated into strings
+ * @param columnToWrite The column that the **StringArray** should be stored in
+ * @param hasHeaders bool to determine if index must be advanced 1 to avoid overwrite
+ */
+struct LoadStringArrayToMatrixImpl
+{
+public:
+  LoadStringArrayToMatrixImpl(PrintMatrix2D& outputMatrix, const StringArray& inputArray, const size_t& columnToWrite, bool hasHeaders = false)
```
JDuffeyBQ:
```suggestion
  LoadStringArrayToMatrixImpl(PrintMatrix2D& outputMatrix, const StringArray& inputArray, size_t columnToWrite, bool hasHeaders = false)
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  void loadToMatrix(size_t start, size_t end) const
+  {
+    if(m_HasHeaders)
+    {
+      size_t row;
+      for(size_t i = start; i < end; i++)
+      {
+        row = i + 1;
+        m_OutputMatrix.getValue(row, m_ColumnToWrite) = m_InputArray[i];
+      }
```
JDuffeyBQ:
```suggestion
      for(size_t i = start; i < end; i++)
      {
        size_t row = i + 1;
        m_OutputMatrix.getValue(row, m_ColumnToWrite) = m_InputArray[i];
      }
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+    }
+  }
+  void operator()(const Range& range) const
+  {
+    loadToMatrix(range.min(), range.max());
+  }
+
+private:
+  PrintMatrix2D& m_OutputMatrix;
+  const size_t& m_ColumnToWrite;
```
JDuffeyBQ:
```suggestion
  size_t m_ColumnToWrite;
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
```
JDuffeyBQ:
```suggestion
  size_t m_ColumnToWrite;
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+                  sstrm << "\n";
+                  count = 0;
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    m_StringStore[start] = std::move(sstrm.str());
```
JDuffeyBQ:
```suggestion
    m_StringStore[start] = sstrm.str();
```
`std::stringstream::str()` returns a new `std::string` so `std::move` is not required.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  const PrintMatrix2D& m_Matrix;
+  std::vector<std::string>& m_StringStore;
+  const size_t m_MaxComp;
+  const std::string m_Delim = "";
+};
+
+/**
+ * @brief Wrapper class for the print functions
+ */
+class COMPLEX_EXPORT OutputFunctions
```
JDuffeyBQ: Since there is not data, this can be made into a namespace and the private member functions can be moved to the cpp file as free functions. And then they should be renamed to match our free function naming/case convention.

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+#include <algorithm>
+#include <fstream>
+#include <ostream>
+#include <sstream>
+#include <string>
+#include <vector>
+
+using namespace complex;
+
+void OStreamUtilities::OutputFunctions::neighborListImplWrapper(DataStructure& dataStructure, PrintMatrix2D& matrixRef, DataObject::Type type, const DataPath& path, bool hasIndex, bool hasHeaders)
```
JDuffeyBQ:
```suggestion
void OStreamUtilities::OutputFunctions::neighborListImplWrapper(const DataStructure& dataStructure, PrintMatrix2D& matrixRef, DataObject::Type type, const DataPath& path, bool hasIndex, bool hasHeaders)
```
`LoadNeighborListToMatrixImpl` takes a `NeighborList` by const ref so `DataStructure` should be passed by const ref also.

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+{
+  if(type == DataObject::Type::INeighborList) // unique because its loaded horizontally no dataAlg use
```
JDuffeyBQ: This `if` should be an early return instead. It will help readability by helping to avoid too much nesting.
```cpp
if(type != DataObject::Type::NeighborList)
{
  throw std::runtime_error("OStreamUtilities::OutputFunctions::neighborListImplWrapper can only process neighborLists")
}
```
Additionally it should use `DataObject::Type::NeighborList` because `DataObject::Type::INeighborList` will never be used. All `NeighborList` classes override `getDataObjectType`.

Unrelated to this PR, but those base class enums really should be removed or we should change how we check for derived types to avoid issues like this. @imikejackson

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+  {
+    auto dataType = dataStructure.getDataAs<INeighborList>(path)->getDataType();
+    switch(dataType)
+    {
+    case DataType::int8: {
+      LoadNeighborListToMatrixImpl(matrixRef, dataStructure.getDataRefAs<NeighborList<int8>>(path), hasIndex, hasHeaders).loadToMatrix(0, dataStructure.getDataAs<NeighborList<int8>>(path)->getSize());
+      break;
+    }
```
JDuffeyBQ:
```suggestion
    const auto& iNeighborList = dataStructure.getDataRefAs<INeighborList>(path);
    auto dataType = iNeighborList.getDataType();
    switch(dataType)
    {
    case DataType::int8: {
      const auto& neighborList = dynamic_cast<const NeighborList<int8>&>(iNeighborList);
      LoadNeighborListToMatrixImpl(matrixRef, neighborList, hasIndex, hasHeaders).loadToMatrix(0, neighborList.getSize());
      break;
    }
```
This should store the neighbor list to avoid having to check the data structure multiple times. And since we know we have a `INeighborList` we can use `getDataRefAs`. This can be repeated for this rest of the cases.

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+    }
+  }
+  else
+  {
+    throw std::runtime_error("Function can only process neighborLists");
+  }
+}
+
+void OStreamUtilities::OutputFunctions::dataAlgParallelWrapper(DataStructure& dataStructure, PrintMatrix2D& matrixRef, DataObject::Type type, const DataPath& path, const size_t& column,
```
JDuffeyBQ: The same change from `OStreamUtilities::OutputFunctions::neighborListImplWrapper` can be made in this function.

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+  }
+}
+
+template <typename T>
+size_t getMaxNeighborlistElements(NeighborList<T> list)
+{
+  size_t max = 0;
+  for(size_t i = 0; i < list.getNumberOfLists(); i++)
+  {
+    auto grain = list.getListReference(i);
```
JDuffeyBQ:
```suggestion
    const auto& grain = list.getListReference(i);
```

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+    }
+  }
+  else
+  {
+    throw std::runtime_error("Cannot process print data for this type");
+  }
+}
+
+template <typename T>
+size_t getMaxNeighborlistElements(NeighborList<T> list)
```
JDuffeyBQ:
```suggestion
size_t getMaxNeighborlistElements(const NeighborList<T>& list)
```

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+  size_t max = 0;
+  for(size_t i = 0; i < list.getNumberOfLists(); i++)
+  {
+    auto grain = list.getListReference(i);
+    for(size_t index = 0; index < grain.size(); index++)
+    {
+      if(grain[index] > max)
+      {
+        max = grain[index];
+      }
```
JDuffeyBQ:
```suggestion
      T value = grain[index];
      if(value > max)
      {
        max = value ;
      }
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+  {
+    size_t row, column;
+    if(m_HasHeader)
+    {
+      if(m_HasIndex) // yes index yes header
+      {
+        for(size_t i = start; i < m_InputArray.getNumberOfLists(); i++)
+        {
+          row = i + 1;
+          auto grain = m_InputArray.getListReference(i);
```
JDuffeyBQ:
```suggestion
          const auto& grain = m_InputArray.getListReference(i);
```
In this case, `auto` would be a plain `std::vector<T>` which would make a copy. In general it is important to specify the const-ness and whether it is a pointer or reference when using `auto` to avoid accidental copies.

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+            m_OutputMatrix.getValue(row, column) = StringUtilities::number(grain[index]);
+          }
+        }
+      }
+      else // no index yes header
+      {
+        for(size_t i = start; i < m_InputArray.getNumberOfLists(); i++)
+        {
+          row = i + 1;
+          auto grain = m_InputArray.getListReference(i);
```
JDuffeyBQ:
```suggestion
          const auto& grain = m_InputArray.getListReference(i);
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+        }
+      }
+    }
+    else // no header
+    {
+      if(m_HasIndex) // yes index no header
+      {
+        for(size_t i = start; i < m_InputArray.getNumberOfLists(); i++)
+        {
+          auto grain = m_InputArray.getListReference(i);
```
JDuffeyBQ:
```suggestion
          const auto& grain = m_InputArray.getListReference(i);
```

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -0,0 +1,765 @@
...
+            column = index + 2;
+            m_OutputMatrix.getValue(i, column) = StringUtilities::number(grain[index]);
+          }
+        }
+      }
+      else // no index no header
+      {
+        for(size_t i = start; i < m_InputArray.getNumberOfLists(); i++)
+        {
+          auto grain = m_InputArray.getListReference(i);
```
JDuffeyBQ:
```suggestion
          const auto& grain = m_InputArray.getListReference(i);
```

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+      hasNeighborLists = true;
+    }
+  }
+  auto sortedMap = std::move(createSortedMapbyType(objectPaths, objTypes));
+  auto matrices = unpackSortedMapIntoMatricies(sortedMap, dataStructure, mesgHandler, includeIndex, includeHeaders, includeNeighborLists);
+
+  ParallelDataAlgorithm dataAlg;
+
+  std::vector<std::map<size_t, std::string>> stringStoreList;
+  auto count = 0;
```
JDuffeyBQ:
```suggestion
  usize count = 0;
```
`auto` with literals can be tricky. In this case, the literal is `int`, but since since this is a count for `std::vector`, `size_t`/`usize` is more appropriate.

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,968 @@
...
+    }
+  }
+}
+
+// Creates OFStream from filepath [NO BINARY SUPPORT]
+void OStreamUtilities::OutputFunctions::printDataSetsToSingleFile(const std::vector<DataPath>& objectPaths, DataStructure& dataStructure, std::filesystem::path& filePath,
+                                                                  const IFilter::MessageHandler& mesgHandler, const std::string& delimiter, bool includeIndex, size_t componentsPerLine,
+                                                                  bool includeHeaders, bool includeNeighborLists)
+{
+  bool hasNeighborLists = false;
```
JDuffeyBQ: Is this used?

src/complex/Utilities/OStreamUtilities.cpp
JDuffeyBQ: It might be good to wrap these functions inside their respective namespace in order to cut down on line length and make it more readable.

JDuffeyBQ (changes requested):
Some of the suggestions can be applied in multiple places since I might not have found every case.

imikejackson:
The API is taking too long inside of the PrintMatrix2D::getBalance method. See attachment for a time stack trace. I had to quit DREAM3D as the filter never wrote anything after a few minutes. We will need to further look into the performance of the new API.

Screen Shot 2022-10-28 at 09 36 35

nyoungbq:
> The API is taking too long inside of the PrintMatrix2D::getBalance method. See attachment for a time stack trace. I had to quit DREAM3D as the filter never wrote anything after a few minutes. We will need to further look into the performance of the new API.
> 
> Screen Shot 2022-10-28 at 09 36 35

I can just strip the get balance method out completely, It was just there to try to speed things up during the print so I doesn't have to check for the "UNINITIALIZED" string.

PR #333, opened 2022-10-24
FILTER: CombineAttributeArrays implemented with Test and Docs

imikejackson:
Also added a few convenience functions in DataArrayUtilities


src/Plugins/ComplexCore/src/ComplexCore/Filters/CombineAttributeArraysFilter.cpp
```diff
@@ -0,0 +1,169 @@
...
+  size_t numComps = 0;
+  for(const auto& dataPath : selectedDataArrayPathsValue)
+  {
+    const auto* dataArray = dataStructure.getDataAs<IDataArray>(dataPath);
+    numComps += dataArray->getNumberOfComponents();
+  }
+
+  // Create the output array
+  {
+    DataPath parentGroup = selectedDataArrayPathsValue[0].getParent();
```
jmarquisbq: @imikejackson Technically there is nothing that stops the user from selecting arrays from different parents to combine... do we want to check for this case or are we okay with always using the first selected array's parent for the destination array?

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/CombineAttributeArrays.cpp
```diff
@@ -0,0 +1,210 @@
...
+    int32_t numArrays = m_InputArrays.size();
+
+    std::vector<InputArrayType*> inputArrays;
+
+    for(size_t i = 0; i < numArrays; i++)
+    {
+      inputArrays.push_back(dynamic_cast<InputArrayType*>(m_InputArrays[i]));
+    }
+
+    size_t numTuples = inputArrays[0]->getNumberOfTuples();
```
jmarquisbq, line 38: @imikejackson Okay so I know that the filter makes sure that we don't have an empty input array but since this class could technically be used without the corresponding filter we should probably do a check to make sure `inputArrays` isn't empty before accessing the data in it.

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/CombineAttributeArrays.cpp
```diff
@@ -0,0 +1,210 @@
...
+  }
+  std::vector<DataObject*> inputArrays;
+  for(const auto& dataPath : m_InputValues->SelectedDataArrayPaths)
+  {
+    inputArrays.push_back(m_DataStructure.getData(dataPath));
+  }
+
+  auto& outputArray = m_DataStructure.getDataRefAs<IDataArray>(m_InputValues->StackedDataArrayPath);
+  auto dataType = outputArray.getDataType();
+  switch(dataType)
```
jmarquisbq: @imikejackson So this is fine to leave as is if you want but I recently discovered the `ExecuteDataFunction` in `FilterUtilities.hpp` which I believe would let you condense this whole switch statement into one line with a few tweaks to the Impl class you are using to do the actual work.  The CropVertexGeometryFilter has an example of it being used.

src/Plugins/ComplexCore/test/CombineAttributeArraysTest.cpp
```diff
@@ -0,0 +1,342 @@
...
+  // Invalid because mismatch array types
+  Float32Array* array3 = complex::UnitTest::CreateTestDataArray<float32>(dataStructure, k_InvalidArrayType, {10ULL, 10ULL}, {3}, 0);
+  array3->fill(static_cast<float32>(3));
+  inputArrays.push_back(DataPath({k_InvalidArrayType}));
+  args.insertOrAssign(CombineAttributeArraysFilter::k_SelectedDataArrayPaths_Key, std::make_any<MultiArraySelectionParameter::ValueType>(inputArrays));
+  args.insertOrAssign(CombineAttributeArraysFilter::k_StackedDataArrayName_Key, std::make_any<DataObjectNameParameter::ValueType>(k_OutputArray));
+  preflightResult = filter.preflight(dataStructure, args);
+  COMPLEX_RESULT_REQUIRE_INVALID(preflightResult.outputActions);
+
+  // Invalid because mismatch array types
```
jmarquisbq:
```suggestion
  // Invalid because mismatch tuple shapes
```

src/Plugins/ComplexCore/test/CombineAttributeArraysTest.cpp
```diff
@@ -0,0 +1,342 @@
...
+    args.insertOrAssign(CombineAttributeArraysFilter::k_SelectedDataArrayPaths_Key, std::make_any<MultiArraySelectionParameter::ValueType>(inputArrays));
+    args.insertOrAssign(CombineAttributeArraysFilter::k_StackedDataArrayName_Key, std::make_any<DataObjectNameParameter::ValueType>(k_OutputArray));
+
+    // Preflight the filter and check result
+    auto preflightResult = filter.preflight(dataStructure, args);
+    COMPLEX_RESULT_REQUIRE_VALID(preflightResult.outputActions);
+
+    auto result = filter.execute(dataStructure, args);
+    COMPLEX_RESULT_REQUIRE_VALID(result.result);
+  }
```
jmarquisbq, line 166: @imikejackson Did you purposely not add the ValidateFilterOutput function check to the rest of these sections like you did in the UINT8 section?

jmarquisbq (changes requested):

src/Plugins/ComplexCore/src/ComplexCore/Filters/CombineAttributeArraysFilter.cpp
imikejackson: Fixed.

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/CombineAttributeArrays.cpp
imikejackson: @jmarquisbq For the Algorithms class, no other algorithm class really does this though. Whether that is by design or we didn't think about it. The implicit understanding of using an Algorithm class from somewhere else is that the developer has already done those checks ahead of time and the variables being passed into the algorithm class have passed some sort of sanity check.
I can put in a few checks but they feel awfully redundant....

imikejackson:
```    if(numArrays == 0)
    {
      return MakeWarningVoidResult(1, "No arrays were selected to combine.");
    }
```
it is a simple enough check...

jmarquisbq: @imikejackson Yeah I realize that it's redundant in this context since the filter already checks that so if you are okay with forcing the user to always put those checks in the filter then I'm fine with this the way it is but maybe put a comment or something so it's easier for the user to pick that up if they try to reuse this elsewhere?

src/Plugins/ComplexCore/test/CombineAttributeArraysTest.cpp
```diff
@@ -0,0 +1,374 @@
...
+    auto result = filter.execute(dataStructure, args);
+    COMPLEX_RESULT_REQUIRE_VALID(result.result);
+
+    Result<> validationResult = ValidateFilterOutput<uint64_t>(dataStructure);
+    COMPLEX_RESULT_REQUIRE_VALID(validationResult)
+  }
+
+  SECTION("INT64")
+  {
+    DataStructure dataStructure = CreateTestDataStructure<int8_t>();
```
jmarquisbq:
```suggestion
    DataStructure dataStructure = CreateTestDataStructure<int64_t>();
```

jmarquisbq (changes requested):

jmarquisbq (approved):

PR #337, opened 2022-10-25
Filter: Crop Image Geometry bug fixes and enhancements

jmarquisbq:
Implements the remove original geometry parameter feature and will copy over the rest of the data from the geometry that is not included in the crop algorithm


imikejackson:
Could we get the same preflight updated Geometry info that the Resample Image Geometry has?

imikejackson (approved):

PR #348, opened 2022-11-04
Just a Test PR.

imikejackson:
## Naming Conventions

- [ ] Class and Structs are `UpperCamelCase`
- [ ] Class private member variables are `m_UpperCamelCase`
- [ ] Class methods are `lowerCamelCase`
- [ ] Method arguments are `lowerCamelCase`
- [ ] Normal variables are `lowerCamelCase`
- [ ] Constants are `k_UpperCamelCase`
- [ ] Global statics are `s_UpperCamelCase`
- [ ] Free Functions are `UpperCamelCase`
- [ ] Macros are `ALL_UPPER_SNAKE_CASE`
- [ ] Unit test will test data output from the filter.
- [ ] Reused strings should be constants in an anonymous namespace
- [ ] If parallelization is used, proper use of the abstracted complex classes are used. Using TBB specifically in a filter should be frowned upon unless for a really good reason.

## Filter Checklist

- [ ] Parameters should be generally broken down into "Input Parameters", "Required Data Objects", "Created Data Objects". There can be exceptions to this.
- [ ] ChoicesParameter selections should be an enumeration defined in the filer header
- [ ] Documentation copied from SIMPL Repo and updated (if necessary)
- [ ] Parameter argument variables are k_CamelCase_Key
- [ ] Parameter argument strings are lower_snake_case
```
```

## Unit Testing
- [ ] 1 Unit test to instantiate the filter with the default arguments.

## Code Cleanup
- [ ] Filters should have both the Filter class and Algorithm class for anything beyond trivial needs
- [ ] Added Python wrapping to new files (if any) as necessary
- [ ] Classes and methods are properly documented.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,7 @@ set(CoreParameters
   ArraySelectionParameter
   NeighborListSelectionParameter
   BoolParameter
+  CalculatorParameter
   ChoicesParameter
   DataGroupCreationParameter
   DataGroupSelectionParameter
--- a/src/Plugins/TestTwo/src/TestTwo/Filters/Test2Filter.cpp
+++ b/src/Plugins/TestTwo/src/TestTwo/Filters/Test2Filter.cpp
@@ -1,6 +1,7 @@
 #include "Test2Filter.hpp"
 
 #include "complex/Common/StringLiteral.hpp"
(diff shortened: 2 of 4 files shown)


imikejackson:
This was just a test and nothing more.

PR #351, opened 2022-11-08
Add Filter Comments

jmarquisbq:
## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup


JDuffeyBQ (commented):
I think comments would be better suited in `PipelineFilter`. This would avoid having to manually add a comment parameter to every filter. And comments are not input to a filter so using a different holder would make more sense.

src/complex/Pipeline/PipelineFilter.hpp
```diff
@@ -103,6 +103,17 @@ class COMPLEX_EXPORT PipelineFilter : public AbstractPipelineNode
    */
   void setIndex(int32 index);
 
+  /**
+   * @brief Gets the node's filter comments.
+   */
+  std::string getComments() const;
```
JDuffeyBQ:
```suggestion
  const std::string& getComments() const;
```

JDuffeyBQ (changes requested):
Do we want a separator with "Input Parameters" on every filter? If so, we could add that to the NX layout by default.

joeykleingers (approved):

PR #353, opened 2022-11-09
ENH: OStreamUtilites API

nyoungbq:
The API for all primary std::ostream child classes. It has capability for printing every child of IArray. Ready for code review.

## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/complex/Utilities/DataArrayUtilities.hpp
+++ b/src/complex/Utilities/DataArrayUtilities.hpp
@@ -569,6 +569,13 @@ COMPLEX_EXPORT bool CheckArraysHaveSameTupleCount(const DataStructure& dataStruc
 
 COMPLEX_EXPORT void ResizeAttributeMatrix(AttributeMatrix& attributeMatrix, const std::vector<usize>& newShape);
 
+/**
+ * @brief Validates that
+ * @param dataStructure
+ * @param arrayPath
+ * @param featureIds
+ * @return
+ */
 COMPLEX_EXPORT Result<> ValidateNumFeaturesInArray(const DataStructure& dataStructure, const DataPath& arrayPath, const Int32Array& featureIds);
 
 /**
--- a/src/complex/Utilities/DataArrayUtilities.hpp
+++ b/src/complex/Utilities/DataArrayUtilities.hpp
@@ -570,11 +570,11 @@ COMPLEX_EXPORT bool CheckArraysHaveSameTupleCount(const DataStructure& dataStruc
 COMPLEX_EXPORT void ResizeAttributeMatrix(AttributeMatrix& attributeMatrix, const std::vector<usize>& newShape);
 
 /**
- * @brief Validates that
- * @param dataStructure
- * @param arrayPath
- * @param featureIds
- * @return
+ * @brief Validates that the number of features in the array are equivalent
+ * @param dataStructure the DataStructure containing the array
+ * @param arrayPath the DataPath to the array in the dataStructure
+ * @param featureIds the ids for the array
+ * @return void
  */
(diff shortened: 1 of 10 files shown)


nyoungbq:
Also stripped out BoolNeighborList typing as it is non-functional with current implementation

nyoungbq:
The unit testing for this API comes in the form of the WriteXXXXDataFilters in the dependent PR in complex plugins

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -569,6 +569,13 @@ COMPLEX_EXPORT bool CheckArraysHaveSameTupleCount(const DataStructure& dataStruc
 
 COMPLEX_EXPORT void ResizeAttributeMatrix(AttributeMatrix& attributeMatrix, const std::vector<usize>& newShape);
 
+/**
```
imikejackson, line 572: Finish this documentation blurb.

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,546 @@
...
+ * //params with defaults
+ * @param fileExtension The extension to create and write to files with
+ * @param exportToBinary The boolean that determines if it writes out binary
+ * @param delimiter The delimiter to be inserted into string | leave blank if binary is end output
+ * @param includeIndex The boolean that determines if "Feature_IDs" are printed | leave blank if binary is end output
+ * @param includeHeaders The boolean that determines if headers are printed | leave blank if binary is end output
+ * @param componentsPerLine The amount of elements to be inserted before newline character | leave blank if binary is end output
+ */
+void PrintDataSetsToMultipleFiles(const std::vector<DataPath>& objectPaths, DataStructure& dataStructure, const std::string& directoryPath, const IFilter::MessageHandler& mesgHandler,
+                                  const std::atomic_bool& shouldCancel, std::string fileExtension, bool exportToBinary, const std::string& delimiter, bool includeIndex, bool includeHeaders,
```
imikejackson: fileExtension should be passed by const reference

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,546 @@
...
+ * @param shouldCancel The atomic boolean that determines cancel
+ * //params with defaults
+ * @param delimiter The delimiter to be inserted into string
+ * @param includeIndex The boolean that determines if "Feature_IDs" are printed
+ * @param includeHeaders The boolean that determines if headers are printed
+ * @param componentsPerLine The amount of elements to be inserted before newline character
+ * @param neighborLists The list of dataPaths of neighborlists to include
+ */
+void PrintDataSetsToSingleFile(std::ostream& outputStrm, const std::vector<DataPath>& objectPaths, DataStructure& dataStructure, const IFilter::MessageHandler& mesgHandler,
+                               const std::atomic_bool& shouldCancel, const std::string& delimiter, bool includeIndex, bool includeHeaders, size_t componentsPerLine,
```
imikejackson: componentsPerLine is never used. Maybe we should remove it from the argument list

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -0,0 +1,546 @@
...
+      writers[writerIndex]->write(outputStrm, tupleIndex);
+      if(writerIndex != writersCount - 1)
+      {
+        outputStrm << delimiter;
+      }
+    }
+    outputStrm << '\n';
+  }
+
+  if(neighborLists.size() != 0)
```
imikejackson:
```suggestion
  if(!neighborLists.empty())
```

imikejackson (changes requested):

imikejackson (approved):

PR #355, opened 2022-11-10
BUG: CropImageGeometry: Fix runtime geometry and array tuple shape bugs

imikejackson:
The root of the problem was the flipping of the X and Z values when creating DataArrays and Creating the Image Geometry.
Refactored many sections to be more efficient in performance and less code to maintain

Also fixed in this commit:
Fix issue in ParallelTaskAlgorithm if the developer set the parallelization to (false) nothing would be executed. Add in some documentation to a few places to help developers understand the ordering of the axis. Updated the test data file to include the exemplar data as SIMPL would produce it.
Updated Resample Image Geometry and RotateSampleRefFrame to use the same kinds and types of functions as CropImageGeometry since they are doing close to the same operations.

Start complex commit messages with a standard prefix (and a space):

 * BUG: fix for runtime crash or incorrect result

## Naming Conventions

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/CropImageGeometry.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/CropImageGeometry.cpp
@@ -17,6 +17,7 @@
 #include "complex/Parameters/GeometrySelectionParameter.hpp"
 #include "complex/Parameters/StringParameter.hpp"
 #include "complex/Parameters/VectorParameter.hpp"
+#include "complex/Utilities/DataArrayUtilities.hpp"
 #include "complex/Utilities/DataGroupUtilities.hpp"
 #include "complex/Utilities/ParallelDataAlgorithm.hpp"
 #include "complex/Utilities/ParallelTaskAlgorithm.hpp"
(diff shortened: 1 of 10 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/CropImageGeometry.cpp
```diff
@@ -396,9 +348,12 @@ IFilter::PreflightResult CropImageGeometry::preflightImpl(const DataStructure& d
...
-  tDims[2] = (zMax - zMin) + 1;
+  // The ImageGeometryDimensions go from Fastest to Slowest, XYZ.
+  geomDims[0] = (xMax - xMin) + 1;
+  geomDims[1] = (yMax - yMin) + 1;
+  geomDims[2] = (zMax - zMin) + 1;
+
+  std::vector<usize> dataArrayShape = {geomDims[2], geomDims[1], geomDims[0]}; // The DataArray shape goes slowest to fastest (ZYX)
 
   FloatVec3 targetSpacing;
   targetSpacing = srcSpacing;
```
jmarquisbq: @imikejackson Why do we create the new variable `targetSpacing` here if we just assign it the same as `srcSpacing`.  Could we just use `srcSpacing` directly instead of creating an extra variable.

src/Plugins/ComplexCore/src/ComplexCore/Filters/CropImageGeometry.cpp
```diff
@@ -356,8 +310,8 @@ IFilter::PreflightResult CropImageGeometry::preflightImpl(const DataStructure& d
 
   // Validate the incoming DataContainer, Geometry, and AttributeMatrix.
   // Provides {0, 0, 0} or {1, 1, 1} respectfully if the geometry could not be found.
-  auto srcDimensions = getCurrentVolumeDataContainerDimensions(data, srcImagePath);
-  auto srcSpacing = getCurrentVolumeDataContainerResolutions(data, srcImagePath);
+  auto srcDimensions = GetCurrentVolumeDataContainerDimensions(data, srcImagePath);
```
jmarquisbq: @imikejackson If I am not mistaken this variable (`srcDimensions`) is the same as `dataArrayShape` below and is not even used.  We should either remove this variable or remove the `dataArrayShape` and use one or the other.

jmarquisbq (changes requested):

src/Plugins/ComplexCore/src/ComplexCore/Filters/CropImageGeometry.cpp
jmarquisbq: Maybe just rename the variable to `spacing` ?

imikejackson: fixed

src/Plugins/ComplexCore/src/ComplexCore/Filters/CropImageGeometry.cpp
imikejackson: fixed

jmarquisbq (approved):

PR #358, opened 2022-11-11
COMP: Fixing compiler warning because boolean enum was not in the switch

imikejackson:

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/InterpolatePointCloudToRegularGridFilter.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/InterpolatePointCloudToRegularGridFilter.cpp
@@ -20,6 +20,7 @@
 #include "complex/Parameters/NumberParameter.hpp"
 #include "complex/Parameters/NumericTypeParameter.hpp"
 #include "complex/Parameters/VectorParameter.hpp"
+#include "complex/Common/TypesUtility.hpp"
 
 namespace complex
 {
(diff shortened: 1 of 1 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/InterpolatePointCloudToRegularGridFilter.cpp
```diff
@@ -68,36 +69,45 @@ void mapPointCloudDataByKernel(IDataArray* source, INeighborList* dynamic, std::
...
+  // to a NumericType from DataType and switch on that. This will stop the compiler warning but
+  // silently skip over the boolean types. Not really sure if this is a problem.
+  auto numericTypeResult = ConvertDataTypeToNumericType(source->getDataType());
+  if(!numericTypeResult.has_value())
   {
-  case DataType::int8:
+    return;
+  }
+
+  switch(numericTypeResult.value())
```
JDuffeyBQ: Instead of converting, could be add a case for `DataType::boolean` that just returns?

imikejackson: I guess we can just put in an empty case statement that just has "break;".

JDuffeyBQ: Functionally we end up skipping bool arrays either way, but a single additional case seems simpler to me. We have handle all the values in `DataType` regardless so going through `NumericType` doesn't gain much.

imikejackson:
```
  case DataType::boolean:
    break;
  }
```

imikejackson: Done

JDuffeyBQ (approved):

PR #359, opened 2022-11-11
BUG: Patch Headers in OStreamUtilities

nyoungbq:
Broken merge overwrote crucial change

### DIFF
--- a/src/complex/Utilities/OStreamUtilities.cpp
+++ b/src/complex/Utilities/OStreamUtilities.cpp
@@ -299,7 +299,7 @@ public:
 
     for(size_t index = 0; index < m_NumComps; index++)
     {
-      outputStrm << m_DataArray.getName() << "_" << (index + 1);
+      outputStrm << m_DataArray.getName() << "_" << index;
 
       if(index < m_NumComps - 1)
       {
--- a/src/complex/Utilities/OStreamUtilities.cpp
+++ b/src/complex/Utilities/OStreamUtilities.cpp
@@ -299,7 +299,14 @@ public:
 
     for(size_t index = 0; index < m_NumComps; index++)
     {
-      outputStrm << m_DataArray.getName() << "_" << index;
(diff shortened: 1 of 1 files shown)


src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -299,7 +299,14 @@ class TupleWriter : public ITupleWriter
...
     {
-      outputStrm << m_DataArray.getName() << "_" << index;
+      if(index == 0)
+      {
+        outputStrm << m_DataArray.getName();
+      }
+      else
+      {
+        outputStrm << m_DataArray.getName() << "_" << index;
+      }
```
imikejackson: The issue here is that if there are 2 components then the first will be "Foo" while the second will be "Foo_1". To be clear what should happen is if there is only 1 component then the output should be:
`Foo`
if there are multiple components then the output should be:
`Foo_1,Foo_2`
 
Apologies for not being clear enough.

imikejackson (approved):

src/complex/Utilities/OStreamUtilities.cpp
nyoungbq: Oh ok lll get that patched

imikejackson: Waiting on the patch to re-approve.

imikejackson (changes requested):

src/complex/Utilities/OStreamUtilities.cpp
imikejackson: I so screwed that up... It _should_ have been:

if there are multiple components then the output should be:
```
Foo_0,Foo_1,Foo_2....
```

imikejackson: I just pushed the fix but that will screw up the complex_plugins PR.

PR #361, opened 2022-11-16
ENH: UUID Backward Compatibility Mapping

nyoungbq:
API and UUID updates in order to prep NX for backwards pipeline compatibility

## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
@@ -2,7 +2,7 @@
 
 #include <map>
 #include <string>
-
+// clang-format off
 namespace complex
 {
   // Regex to grep UUIDs : {\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\",\s\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(?:\s,\s[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})*)\"},\s\/\/\s(\w+)
@@ -122,3 +122,4 @@ namespace complex
   };
 
 } // namespace complex
+// clang-format on
--- a/src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
@@ -5,7 +5,7 @@
 
 namespace complex
 {
-  // Regex to grep UUIDs : /{\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\",\s\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(?:\s,\s[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})*)\"},\s\/\/\s(\w+)
+  // Regex to grep UUIDs : {\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\",\s\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(?:\s,\s[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})*)\"},\s\/\/\s(\w+)
   static const std::map<std::string, std::string> k_SIMPL_to_ComplexCore
   {
     // syntax std::make_pair {Dream3d UUID , Dream3dnx UUID}, // dream3d-class-name
--- a/src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
@@ -1,0 +1,124 @@
+#pragma once
+
+#include <map>
+#include <string>
+
+namespace complex
+{
+  // Regex to grep UUIDs : /{\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\",\s\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(?:\s,\s[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})*)\"},\s\/\/\s(\w+)
+  static const std::map<std::string, std::string> k_SIMPL_to_ComplexCore
+  {
+    // syntax std::make_pair {Dream3d UUID , Dream3dnx UUID}, // dream3d-class-name
+    {"ce1ee404-0336-536c-8aad-f9641c9458be", "87fa9e07-6c66-45b0-80a0-cf80cc0def5d"}, // AlignGeometries
+    {"c681caf4-22f2-5885-bbc9-a0476abc72eb", "f5bbc16b-3426-4ae0-b27b-ba7862dc40fe"}, // ApplyTransformationToGeometry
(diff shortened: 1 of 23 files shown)


src/complex/Core/Application.hpp
```diff
@@ -118,6 +118,18 @@ class COMPLEX_EXPORT Application
...
+   * @brief Returns the Complex filter UUID [v4] from the SIMPL filter UUID [v5]
+   * @return std::optional<Uuid>
+   */
+  std::optional<Uuid> getComplexUuid(const Uuid& simplUuid);
+
+  /**
+   * @brief Returns the SIMPL filter UUID(s) [v5] from the Complex filter UUID [v4]
+   * @return std::optional<std::vector<Uuid>>
+   */
+  std::optional<std::vector<Uuid>> getSimplUuid(const Uuid& complexUuid);
```
JDuffeyBQ: Why does this return a vector of uuids?

src/Plugins/ComplexCore/src/ComplexCore/ComplexCorePlugin.hpp
```diff
@@ -25,4 +26,24 @@ class COMPLEXCORE_EXPORT ComplexCorePlugin : public complex::AbstractPlugin
    * @return std::vector<complex::H5::IDataFactory*>
    */
   std::vector<complex::H5::IDataFactory*> getDataFactories() const override;
+
+  /**
+   * @brief Returns a map of UUIDs as strings, where SIMPL UUIDs are keys to
+   * their complex counterpart
+   * @return std::map<std::string, std::string>
+   */
+  std::map<std::string, std::string> getSimplToComplexMap() const override
```
JDuffeyBQ: The implementation of these functions should be in the cpp file.

src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
```diff
@@ -0,0 +1,125 @@
+#pragma once
+
+#include <map>
+#include <string>
+// clang-format off
+namespace complex
+{
+  // Regex to grep UUIDs : {\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\",\s\"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(?:\s,\s[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})*)\"},\s\/\/\s(\w+)
+  static const std::map<std::string, std::string> k_SIMPL_to_ComplexCore
```
JDuffeyBQ: I would make these a `std::map<Uuid, Uuid>` instead. And maybe we should consider a bidirectional map (or approximation of one) so we don't have to duplicate the mapping.

Also I think we should use the `FilterTraits<T>::uuid` here to avoid duplicating the uuid definitions.

src/complex/Core/Application.cpp
```diff
@@ -111,6 +112,45 @@ std::filesystem::path Application::getCurrentDir() const
   return m_CurrentPath.parent_path();
 }
 
+std::optional<Uuid> Application::getComplexUuid(const Uuid& simplUuid)
+{
+  std::string searchUuid = simplUuid.str();
+  if(m_SIMPL_To_Complex.count(searchUuid) == 1)
```
JDuffeyBQ: For early returns, I think that using the error condition is better. Code with less nesting is easier to read.
```cpp
if(m_SIMPL_To_Complex.count(searchUuid) == 0)
{
  return {};
}
return Uuid::FromString(m_SIMPL_To_Complex[searchUuid]);
```
The same kind of change can be made in the other function.

src/complex/Core/Application.cpp
```diff
@@ -171,6 +211,10 @@ void Application::loadPlugin(const std::filesystem::path& path, bool verbose)
   getFilterList()->addPlugin(pluginLoader);
 
   auto plugin = pluginLoader->getPlugin();
+
+  m_SIMPL_To_Complex.merge(plugin->getSimplToComplexMap());
```
JDuffeyBQ: `std::map::merge` skips entries that would be duplicates. We may want to report that a duplicate mapping exists.

JDuffeyBQ (changes requested):

src/complex/Core/Application.hpp
nyoungbq: There are multiple dream3d filters that map to a single complex filter

src/complex/Core/Application.cpp
imikejackson: We should report duplicate entries of complex UUIDs on plugin load (which I think we do?) Or maybe that was in SIMPL where I hit that.

imikejackson:
Looking at how to map back and forth I think storing the 2 lists as just 2 vectors where you know index=1 of the SIMPL vector has a direct relationship to index=1 of the COMPLEX vector. You can have duplicates in the COMPLEX vector which is an interesting case to figure out.

I think we have this mostly to go _from_ SIMPL _to_ COMPLEX. I don't see the need to go the other way so we can have duplicates in the COMPLEX vector and that is fine. the lookup time is going to be linear but these vectors are not going to be past a few hundred, maybe 500 tops in the lifetime of complex. those lookups are going to go really fast in actual wall clock time, fast enough that the performance wouldn't really matter. I'd guess on modern machines the look shouldn't take more than 10s of milliseconds so complicated multimaps and bimaps just are not needed, in my opinion. @JDuffeyBQ  thoughts?

src/Plugins/ComplexCore/src/ComplexCore/ComplexCoreLegacyUUIDMapping.hpp
nyoungbq: On the topic of bidirectional mapping I think that in this case I could do the same with two std::vectors that are equivalent by index. There are only 450 filter UUIDs and about 3 instances of multiple UUIDs. If it was imlemented this way we could make them std::vector<complex::Uuid> because it wouldn't require each key to be unique. Using some index tricks multiple SIMPL UUIDs could be pulled by searching for duplicate instances in the complex vector...but this introduces the chance for unwanted duplicates [especially since they are UUID version 4 (puesdo-random)].

src/Plugins/ComplexCore/src/ComplexCore/ComplexCorePlugin.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <vector>
 
+#include "ComplexCoreLegacyUUIDMapping.hpp"
```
JDuffeyBQ: This can be moved to the cpp as well.

src/complex/Core/Application.cpp
JDuffeyBQ: We do report duplicate filter uuids but not in this mapping. If one plugin attempts to map a SIMPL filter to the complex one when a mapping already exists, the new mapping will be silently ignored.

JDuffeyBQ (changes requested):

src/complex/Core/Application.cpp
```diff
@@ -171,6 +207,30 @@ void Application::loadPlugin(const std::filesystem::path& path, bool verbose)
...
+  for(auto const& [simplUuidString, ComplexUuidString] : simplToComplexUuids)
+  {
+    auto simplUuid = Uuid::FromString(simplUuidString).value();
+    auto complexUuid = Uuid::FromString(ComplexUuidString).value();
+    for(const auto& uuid : m_Simpl_Uuids)
+    {
+      if(uuid == simplUuid)
+      {
+        throw std::runtime_error(fmt::format("Duplicate UUIDs found in the SIMPL UUID maps! UUID: {} Plugin: {}", simplUuid.str(), plugin->getName()));
+        return;
```
JDuffeyBQ:
```suggestion
```
No need for `return` after throwing an exception.

src/complex/Core/Application.cpp
```diff
@@ -171,6 +207,30 @@ void Application::loadPlugin(const std::filesystem::path& path, bool verbose)
...
+      }
+    }
+    m_Simpl_Uuids.push_back(simplUuid);
+    m_Complex_Uuids.push_back(complexUuid);
+  }
+
+  if(m_Simpl_Uuids.size() != m_Complex_Uuids.size())
+  {
+    throw std::runtime_error(fmt::format("UUID maps are not of the same size! SIMPL UUID Vector size: {} Complex UUID Vector size: {}", m_Simpl_Uuids.size(), m_Complex_Uuids.size()));
+    return;
```
JDuffeyBQ:
```suggestion
```

JDuffeyBQ (changes requested):
`ComplexCoreLegacyUUIDMapping.hpp` also should use `Uuid` instead of string. And potentially use the `FilterTraits<T>::uuid` for complex filters to avoid the ids from getting out of sync.

src/complex/Core/Application.hpp
JDuffeyBQ: I think it's ok to return `std::vector<Uuid>` here because there isn't a special case for an empty vector that needs to be handled. An empty vector will mean that there were no associated SIMPL filters.

JDuffeyBQ (approved):

PR #362, opened 2022-11-16
Check if user has write permissions to output directories

imikejackson:


imikejackson:
Will fix issue #362

src/complex/Parameters/FileSystemPathParameter.cpp
```diff
@@ -47,9 +59,32 @@ Result<> ValidateInputDir(const FileSystemPathParameter::ValueType& path)
...
+//-----------------------------------------------------------------------------
+Result<> ValidateDirectoryWritePermission(const FileSystemPathParameter::ValueType& path)
+{
+  auto checkedPath = path;
+  while(!fs::exists(checkedPath))
+  {
+    checkedPath = checkedPath.parent_path();
+  }
+  // We should be at the top of the tree with an existing directory.
+  int accessPerms = FSPP_ACCESS(checkedPath.string().c_str(), FSPP_CHECK_WRITABLE);
```
JDuffeyBQ: I'd wrap this in it's own function. It'll make it easy to extract if we need it somewhere else in the future.
```cpp
bool HasWriteAccess(const std::string& path)
{
  return FSPP_ACCESS(path.c_str(), FSPP_CHECK_WRITABLE) == 0;
}
```

src/complex/Parameters/FileSystemPathParameter.cpp
```diff
@@ -9,6 +9,17 @@
 #include <nlohmann/json.hpp>
 
 #include <stdexcept>
+#ifdef _MSC_VER
```
JDuffeyBQ:
```suggestion
#ifdef _WIN32
```

JDuffeyBQ (changes requested):

src/complex/Parameters/FileSystemPathParameter.cpp
imikejackson: What about clang on windows? i'm not sure how that behaves. I would assume it behaves like unix/linux which is why I went for MSC_VER instead of Win32, but I may be wrong on that one.

JDuffeyBQ: From what I've read, I think `_WIN32` works the best here. There's some overlap between the macros but since we're using Windows specific headers I think that one is the most relevant.

imikejackson:
Any idea why the test dies on Windows v141 toolset? @JDuffeyBQ

PR #364, opened 2022-11-17
Filter:Array Calculator

jmarquisbq:
## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup


imikejackson (approved):
Seems to work well. Some small code cleanups and a styling pass at the widget. Feel free to change anything you don't like.

imikejackson (approved):

PR #382, opened 2022-12-09
FILTER: Implement Multiple Filters

nyoungbq:
## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ExtractComponentAsArray.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ExtractComponentAsArray.cpp
@@ -7,17 +7,61 @@
 using namespace complex;
 namespace
 {
-//struct ExtractComponentsFunctor
-//{
-//  template <class T>
-//  void operator()(IDataArray& inputDataRef, IDataArray& extractedCompRef, int32 componentsToExtract, bool insertInNew)
-//  {
-//    auto& inputDataRef = dynamic_cast<DataArray<T>&>(inputDataPtr);
-//    auto& maskedData = dynamic_cast<DataArray<T>&>(maskedDataPtr);
-//
-//    inputDataRef.getIDataStore()->
-//  }
-//};
+struct RemoveComponentsFunctor
+{
+  template <class ScalarType>
+  void operator()(IDataArray& originalArray, IDataArray& resizedArray, usize componentIndexToRemove) // Due to logic structure originalArray cannot be const
+  {
+    const auto& originalArrayRef = dynamic_cast<const DataArray<ScalarType>&>(originalArray);
+    auto& resizedArrayRef = dynamic_cast<DataArray<ScalarType>&>(resizedArray);
+
+    usize originalTupleCount = originalArrayRef.getNumberOfTuples();
+    usize originalCompCount = originalArrayRef.getNumberOfComponents();
+
+    usize distanceToShuffle = 0;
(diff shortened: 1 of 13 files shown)


nyoungbq:
This PR includes: 
- new ExtractComponentAsArray [has been combined with RemoveComponentFromArray]
- Updated ConditionalSetValue [has been combined with ReplaceConditionalValueInArray]
- minor change to FilterUtilities to catch fringe case in neighborlist bool

All filters have Unit Tests, UUID Mapping, and documentation updated.

imikejackson (changes requested):
# Replace Value in Array (Conditional)

Documentation needs to spell out there is an option to use a mask and what the ramifications of using a mask array are.

Take a look at FindArrayStatistics and see how that filter groups the option for using a mask with the selection of the mask array. Place this section of parameters __below__ the "Value to Replace" and "New Value" parameters.

Change the Separator "Required Objects" to "Required Input Data"

# Extract/Remove Components

The mouse hover for the "Move Extracted components to new Array" should read in the positive instead of the double negative. 

The error text when I place a component index that is larger than the number of components should read "The component index '4' is larger than the total number of components. Valid values are between 0 and 2 (inclusive)"

The default name of the created data array should be in singular form since we are only extracting a single component at a time.

Make the component index to extract the first parameter

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ExtractComponentAsArray.cpp
```diff
@@ -0,0 +1,114 @@
...
+  const bool removeComponentsFromArrayBool = m_InputValues->RemoveComponentsFromArray;
+  const usize compToRemoveNum = static_cast<usize>(abs(m_InputValues->CompNumber));
+  // this will be the original array if components are not being removed, else it is resized array
+  IDataArray& baseArrayRef = m_DataStructure.getDataRefAs<IDataArray>(m_InputValues->BaseArrayPath);
+  if((!removeComponentsFromArrayBool) && moveComponentsToNewArrayBool)
+  {
+    ExecuteDataFunction(ExtractComponentsFunctor{}, baseArrayRef.getDataType(), baseArrayRef, m_DataStructure.getDataRefAs<IDataArray>(m_InputValues->NewArrayPath), compToRemoveNum);
+    return {};
+  }
+  // will not exist if remove components isnt occuring, hence the early bailout ^
```
imikejackson: Fix the spelling mistake. Check use of contractions. Use the ' or full spell out the formal words.

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ExtractComponentAsArray.cpp
```diff
@@ -0,0 +1,114 @@
...
+  IDataArray& tempArrayRef = m_DataStructure.getDataRefAs<IDataArray>(m_InputValues->TempArrayPath); // will not exist if remove components isnt true, hence the early bailout ^
```
imikejackson: check use of contractions

imikejackson (changes requested):

imikejackson:
Screen Shot 2022-12-09 at 16 10 21

here are the clang-tidy warnings. These are optional fixes but I am tending to do them now.
The ones about "2 Adjacent variables" you can ignore though.. .sort of hard to fix those.

imikejackson (approved):

PR #386, opened 2022-12-13
BUG: Range2D, Dream3dIO, H5AttributeReader fixes

jmarquisbq:
These changes are in support of the GBCD Filter that is being ported.

## Code Cleanup

### DIFF
--- a/src/complex/Common/Range2D.cpp
+++ b/src/complex/Common/Range2D.cpp
@@ -31,25 +31,25 @@ Range2D::RangeType Range2D::getRange() const
 // -----------------------------------------------------------------------------
 size_t Range2D::minRow() const
 {
-  return m_Range[0];
+  return m_Range[2];
 }
 
 // -----------------------------------------------------------------------------
 size_t Range2D::minCol() const
 {
-  return m_Range[1];
+  return m_Range[0];
 }
 
 // -----------------------------------------------------------------------------
 size_t Range2D::maxRow() const
 {
-  return m_Range[2];
+  return m_Range[3];
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


src/complex/Utilities/TimeUtilities.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "complex/Utilities/StringUtilities.hpp"
+
+#include <fmt/core.h>
+#include <fmt/ranges.h>
+#include <string>
+
+namespace complex
+{
+inline std::string ConvertMillisToHrsMinSecs(unsigned long long int millis)
```
nyoungbq, line 17: Here I would make use of the std::chrono module to do the conversions for you with duration_casts. You could completely avoid stringUtilities.hpp. A good example of this would be here: https://stackoverflow.com/a/64169432/20736404

src/complex/Common/Range2D.cpp
```diff
@@ -31,25 +31,25 @@ Range2D::RangeType Range2D::getRange() const
 // -----------------------------------------------------------------------------
 size_t Range2D::minRow() const
 {
-  return m_Range[0];
+  return m_Range[2];
```
nyoungbq, line 34: Not required, but could be useful to document what these index represent next to the member variable such as:
`auto m_Range = {}; // layout: { min Column, max Column, min Row, max Row } `
and then next to the return statements just mention the documentation like:
`return m_Range[0]; // see member declaration for details`

imikejackson: The layout is documented in the header

nyoungbq (approved):

imikejackson (approved):

PR #395, opened 2022-12-26
FILTER: ImportVolumeGraphicsFile

joeykleingers:


src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+inline const std::string k_geometryBlock("[geometry]");
+// inline const std::string k_representationBlock("[representation]");
+
+inline const std::string k_RegionOfInterestStart("RegionOfInterestStart");
+inline const std::string k_RegionOfInterestEnd("RegionOfInterestEnd");
+
+inline const std::string k_FileFormat("FileFormat");
+inline const std::string k_Size("Size");
+inline const std::string k_Name("Name");
+inline const std::string k_Datatype("Datatype");
```
imikejackson: I think we settled on k_CamelCase for constant variables

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+#include "complex/Utilities/StringUtilities.hpp"
+
+#include <filesystem>
+namespace fs = std::filesystem;
+
+using namespace complex;
+
+namespace
+{
+static inline constexpr int32 k_FileDoesNotExist = -91501;
```
imikejackson: static is redundant inside an anonymous namespace

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+  SizeVec3 Size = {0, 0, 0};
+  std::string Name = "AS_N610_02.vol";
+  std::string Datatype = "float";
+  FloatVec2Type datarange = {0.0F, 0.0F};
+  int8 BitsPerElement = 32;
+};
+
+struct GeometryBlock
+{
+  std::string status = "visible";
```
imikejackson: We should be consistent about case inside of structs. This one has 'lowerCame' and the struct above has mixed lowerCamel and UpperCamel

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+inline const std::string k_rotate("rotate");
+inline const std::string k_unit("unit");
+
+struct FileBlock
+{
+  SizeVec3 RegionOfInterestStart = {0, 0, 0};
+  SizeVec3 RegionOfInterestEnd = {0, 0, 0};
+  std::string FileFormat = "raw";
+  SizeVec3 Size = {0, 0, 0};
+  std::string Name = "AS_N610_02.vol";
```
imikejackson: This should be initialized to an empty string.

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+  std::array<int32, 9> rotate = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+  std::string unit = "mm";
+};
+
+Result<SizeVec3> ParseSizeVec3(const std::vector<std::string>& tokens)
+{
+  SizeVec3 vec = {0, 0, 0};
+  std::string vgiParseErrorStr = "Error parsing usize value ({}) from vgi file.";
+  std::string vgiOutOfRangeErrorStr = "Value ({}) in vgi file is out of range for storage type usize.";
+
```
imikejackson: validate that there are enough tokens instead of throwing the exception (for out of range)

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+  return {vec};
+}
+
+Result<FloatVec2Type> ParseFloatVec2(const std::vector<std::string>& tokens)
+{
+  FloatVec2Type vec = {0.0F, 0.0F};
+  std::string vgiParseErrorStr = "Error parsing floating point value ({}) from vgi file.";
+  std::string vgiOutOfRangeErrorStr = "Floating point value ({}) in vgi file is out of range for storage type float32.";
+
```
imikejackson: Same token length check here also. It will help simplify the code

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+  return {vec};
+}
+
+Result<FloatVec3> ParseFloatVec3(const std::vector<std::string>& tokens)
+{
+  FloatVec3 vec = {0.0F, 0.0F, 0.0F};
+  std::string vgiParseErrorStr = "Error parsing floating point value ({}) from vgi file.";
+  std::string vgiOutOfRangeErrorStr = "Floating point value ({}) in vgi file is out of range for storage type float32.";
+
```
imikejackson, line 89: Same token length check

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  return {vec};
+}
+
+template <typename T, usize Dimension = 9>
+Result<std::array<T, Dimension>> ParseIntArray(const std::vector<std::string>& tokens)
+{
+  std::array<T, Dimension> vec;
+
```
imikejackson: Token length check

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  fs::path fiHdr = fs::path(vgHeaderFilePath);
+  metadata.dataFilePath = (fs::absolute(fiHdr.parent_path()) / fileBlock.Name).string();
+  return {metadata};
+}
+
+struct ImportVolumeGraphicsFileFilterCache
+{
+  fs::path vgiDataFilePath;
```
imikejackson: make case consistent with other structs in this file

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+};
+
+static std::atomic_int32_t s_InstanceId = 0;
```
imikejackson: static is redundant for anonymous namespace

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+//------------------------------------------------------------------------------
+std::string ImportVolumeGraphicsFileFilter::humanName() const
+{
+  return "Import Volume Graphics File (.vgi/.vol)";
+}
+
+//------------------------------------------------------------------------------
+std::vector<std::string> ImportVolumeGraphicsFileFilter::defaultTags() const
+{
+  return {"#IO", "#Input", "#Read", "#Import"};
```
imikejackson: Add #CT #VolumeGraphics #vgi

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+}
+
+//------------------------------------------------------------------------------
+Parameters ImportVolumeGraphicsFileFilter::parameters() const
+{
+  Parameters params;
+
+  params.insert(std::make_unique<FileSystemPathParameter>(k_VGHeaderFile_Key, "VolumeGraphics .vgi File", "", fs::path("DefaultInputFileName"), FileSystemPathParameter::ExtensionsType{".vgi"},
```
imikejackson: Need the help text filled in. These are used for the Python documentation so they help. It may seem obvious.. .but..

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
...
+std::vector<std::string> ImportVolumeGraphicsFileFilter::defaultTags() const
+{
+  return {"#IO", "#Input", "#Read", "#Import"};
+}
+
+//------------------------------------------------------------------------------
+Parameters ImportVolumeGraphicsFileFilter::parameters() const
+{
+  Parameters params;
+
```
imikejackson, line 389: Break up the parameters such that generally you have:
params.insertSeparator(Parameters::Separator{"Parameters"});

params.insertSeparator(Parameters::Separator{"Input Data"});

params.insertSeparator(Parameters::Separator{"Created Data"});

You can embellish those if the code calls for it and would ultimately help the user.

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ImportVolumeGraphicsFile.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "complex/DataStructure/DataArray.hpp"
+#include "complex/DataStructure/DataGroup.hpp"
+#include "complex/DataStructure/Geometry/ImageGeom.hpp"
+
+using namespace complex;
+namespace fs = std::filesystem;
+
+namespace
+{
+inline constexpr int32 k_VolBinaryAllocateMismatch = -91504;
```
imikejackson: I think these should just be `inline const int32`. By using an anonymous namespace we don't need static and constexpres should generally not be used inside of an anonymous namespace. Which if that is the case then there are a whole bunch of filters that need to be updated

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ImportVolumeGraphicsFile.cpp
```diff
@@ -0,0 +1,69 @@
...
+inline constexpr int32 k_VolOpenError = -91505;
+inline constexpr int32 k_VolReadError = -91506;
```
imikejackson: if we are going to be using `usize` then we should be using `int32` instead of `int32_t`

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/ImportVolumeGraphicsFile.hpp
```diff
@@ -0,0 +1,46 @@
...
+namespace complex
+{
+
+struct COMPLEXCORE_EXPORT ImportVolumeGraphicsFileInputValues
+{
+  FileSystemPathParameter::ValueType VGHeaderFile;
+  DataPath ImageGeometryPath;
+  std::string CellAttributeMatrixName;
+  std::string DensityArrayName;
+  std::filesystem::path VGDataFile;
```
imikejackson: You should inlcude the <filesystem> header if you are using std::filesystem in the header. Along with <string> and strip out includes that are not directly being used in this header like ArrayCreationParameter.hpp

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
+#include "ImportVolumeGraphicsFileFilter.hpp"
+
+#include <fstream>
+
+#include "Algorithms/ImportVolumeGraphicsFile.hpp"
```
imikejackson: This should be the complete path to the Algorithms folder. so ComplexCore/Filters/Algorithms/.....

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportVolumeGraphicsFileFilter.cpp
```diff
@@ -0,0 +1,558 @@
+#include "ImportVolumeGraphicsFileFilter.hpp"
+
+#include <fstream>
```
imikejackson: Group the C++ STL includes after all other includes

imikejackson (changes requested):

nyoungbq (approved):

PR #408, opened 2023-01-11
FILTER: Execute process

jmarquisbq:
## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/docs/ExecuteProcessFilter.md
+++ b/src/Plugins/ComplexCore/docs/ExecuteProcessFilter.md
@@ -7,13 +7,16 @@ Core (Misc)
 
 ## Description ##
 
-This filter allows the user to execute any application, program, shell script or any other executable program on the computer system. Any output can be viewed with the Standard Output dock widget in SIMPLView.
+This filter allows the user to execute any application, program, shell script or any other executable program on the computer system. Any output can be found in the user specified log file.
 
 ## Parameters ##
 
 | Name             | Type | Description |
 |------------------|------|-------------|
 | Command Line | String| The complete command to execute. |
+| Should Block | Bool | Whether to block the pipeline process while the command executes or not. |
+| Timeout | Int32 | The amount of time (in ms) to wait for the command to start/finish when blocking is selected. |
+| Output Log File | File Path | The log file where the output from the process will be stored. |
 
 
 ## Required Geometry ##
--- a/src/Plugins/ComplexCore/docs/ExecuteProcessFilter.md
+++ b/src/Plugins/ComplexCore/docs/ExecuteProcessFilter.md
@@ -1,0 +1,43 @@
+# Execute Process
+
+
+## Group (Subgroup) ##
+
+Core (Misc)
+
+## Description ##
+
+This filter allows the user to execute any application, program, shell script or any other executable program on the computer system. Any output can be viewed with the Standard Output dock widget in SIMPLView.
+
+## Parameters ##
+
+| Name             | Type | Description |
+|------------------|------|-------------|
+| Command Line | String| The complete command to execute. |
+
+
+## Required Geometry ##
+
+Not Applicable
+
+## Required Objects ##
+
+None
+
+## Created Objects ##
+
+None
+
+
+## Example Pipelines ##
+
+
+
+## License & Copyright ##
+
+Please see the description file distributed with this **Plugin**
+
+## DREAM.3D Mailing Lists ##
+
+If you need more help with a **Filter**, please consider asking your question on the [DREAM.3D Users Google group!](https://groups.google.com/forum/?hl=en#!forum/dream3d-users)
+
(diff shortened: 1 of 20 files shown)


src/complex/Parameters/FileSystemPathParameter.cpp
```diff
@@ -73,6 +73,10 @@ Result<> ValidateDirectoryWritePermission(const FileSystemPathParameter::ValueTy
   while(!fs::exists(checkedPath))
   {
     checkedPath = checkedPath.parent_path();
+    if(checkedPath.empty())
```
imikejackson: Take a look at https://github.com/BlueQuartzSoftware/complex/pull/417/files and we can discuss the differences

src/complex/DataStructure/DataStructure.hpp
```diff
@@ -117,10 +117,10 @@ class COMPLEX_EXPORT DataStructure
   /**
    * @brief Returns true if the DataStructure contains a DataObject with the
    * given key. Returns false otherwise.
-   * @param identifier
+   * @param id
```
imikejackson: I think we took a step backwards with the change from "identifier" to "id"?

src/complex/DataStructure/DataArray.hpp
```diff
@@ -58,16 +58,16 @@ class DataArray : public IDataArray
...
    *
-   * @param dataStructure The parent DataStructure that will own the DataArray
+   * @param ds The parent DataStructure that will own the DataArray
    * @param name The name of the DataArray
    * @param store The IDataStore instance to use. The DataArray instance WILL TAKE OWNERSHIP of the DataStore Pointer.
    * @param parentId = {}
    * @return DataArray<T>* Instance of the DataArray object that is owned and managed by the DataStructure
    */
-  static DataArray* Create(DataStructure& dataStructure, std::string name, std::shared_ptr<store_type> store, const std::optional<IdType>& parentId = {})
+  static DataArray* Create(DataStructure& ds, std::string name, std::shared_ptr<store_type> store, const std::optional<IdType>& parentId = {})
```
imikejackson: Why was the change from "dataStructure" to "ds"?

src/Plugins/ComplexCore/test/ExecuteProcessTest.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+TEST_CASE("ComplexCore::ExecuteProcessFilter: Valid filter execution")
+{
+  // Instantiate the filter, a DataStructure object and an Arguments Object
+  ExecuteProcessFilter filter;
+  DataStructure ds;
+  Arguments args;
+
+  fs::path processOutput(fmt::format("{}/ExecuteProcessUnitTestOutput.txt", unit_test::k_BuildDir));
+  std::string testCommand = fmt::format("{}/PipelineRunner_d", unit_test::k_BuildDir);
```
imikejackson, line 26: hard coding with the debug extension would surely fail for release builds?

src/Plugins/ComplexCore/docs/ExecuteProcessFilter.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Group (Subgroup) ##
+
+Core (Misc)
+
+## Description ##
+
+This filter allows the user to execute any application, program, shell script or any other executable program on the computer system. Any output can be found in the user specified log file.
+
```
imikejackson, line 11: I would like to see some examples of what would be entered into the parameter. Even just simple things but showing that you should use the full absolute path, arguments to the executable.

For example for simple directory listing:
```
/bin/ls
```
for something with arguments
```
/bin/ls -laF
```
for something with a space in the path or name we need double quotes

```
cp "/tmp/Name of File.txt" "/tmp/Better_Name_of_File.txt"
```

imikejackson (changes requested):

src/complex/Parameters/FileSystemPathParameter.cpp
jmarquisbq: @imikejackson I think we essentially solved it with the same checks.  I put the absolute vs relative path checks down in the main validation function because you might run in to relative paths you can't for input files/directories on top of output in which case the ValidateDirectoryWritePermission fct won't get hit.  Other than that though I like your updates better since your error messaging is more descriptive.

src/Plugins/ComplexCore/test/ExecuteProcessTest.cpp
jmarquisbq: @imikejackson That's what the "#if NDEBUG // release build" section just below this is for.  I can certainly switch it around to be "{}/PipelineRunner" by default and use "#ifndef NDEBUG // debug builds" to fix it for debug mode if you prefer. I didn't love the double negative with statement but I don't really have a strong preference.

src/complex/DataStructure/DataArray.hpp
jmarquisbq: @imikejackson I honestly don't know where these changes came from because I don't recall making them.  It was most likely a bad rebase.  I'll fix it 👍

src/complex/DataStructure/DataStructure.hpp
jmarquisbq: @imikejackson Same as above, pretty sure this was a bad rebase lol

imikejackson (approved):

PR #412, opened 2023-01-14
RotateSampleRefFrame: Fix unit test to compare output to exemplar arrays

imikejackson:
Copy paste error from original port left the unit test comparing generated array versus generated array. This test now compares the exemplar array as generated from DREAM3D 6.6.x

The parsing of the DataContainer name into a floating point number was causing differences in the input axis-angle data values as compared to running in the GUI or from a JSON pipeline file. A mapping was added for each Test DataContainer that has the proper axis-angle that was used to rotate that particular Data Container.


imikejackson:
When this is merged we need to Edit the https://github.com/BlueQuartzSoftware/complex/releases/tag/Data_Archive release to remove the `RotateSampleRefFrameTest.dream3d.tar.gz` entry and associated Data File. A new exemplar data file was uploaded. 

I could not do this now because I would break every other PR.

nyoungbq (approved):

PR #415, opened 2023-01-17
STYLE: Strip Unused Includes

nyoungbq:


imikejackson:
This all generally looks good but if we are going down the clean up route we should also ensure we #include headers in a consistent ordering. That ordering is:

Current Plugin or Library
Dependent Complex Library

3rdParty dependent libraries (HDF5/fmt/span/mp11)

STL/C++ includes

jmarquisbq (approved):

PR #420, opened 2023-01-20
Returning a reasonable error message if bad_any_cast is thrown.

joeykleingers:


joeykleingers:
We discussed this and the check is done further up in IFilter.cpp.  Additionally, no other part of the code base will use the parameter validate methods, so this is not needed.

PR #421, opened 2023-01-20
Feature: Plugin defined Out-of-Core Data Storage

mmarineBlueQuartz:
* Moved complex/Common to its own library, NXCommon
* Extracted complex HDF5 utilities to NXH5Support.
* Updated to support AbstractDataStore defined in external plugins.
* Updated to abstract DataStructure IO classes to allow plugins to add additional format support.
* Updated DataStructure HDF5 IO classes to support out-of-core data.


JDuffeyBQ (commented):
I think reading new dream3d files isn't working correctly. I was running this with ITKImageProcessing, and ITKImageWriter is failing due to reading the geometry dimensions as (0, 0, 0). It looks like some of the reading code wasn't brought over to the new files. Other geometries could be broken as well though I haven't tested them yet.

Also the DREAM3DFileTest seems to be deleted as well. Was this intentional? I don't think that test would have caught this issue, but we should have tests for the reading/writing of geometries in new dream3d files if we don't already.

mmarineBlueQuartz:
The DREAM3DFileTest was deleted when the file was no longer found after a rebase, so I thought it had been deleted / moved on develop (why it was removed from CMakeLists). As for reading the dimensions, I'll take a look and get that updated.

PR #427, opened 2023-01-20
FILTER: Find Vertex to Triangle Distances

nyoungbq:
## Naming Conventions

## Filter Checklist

```
```

## Unit Testing

## Code Cleanup


src/Plugins/ComplexCore/docs/FindVertexToTriangleDistances.md
```diff
@@ -0,0 +1,37 @@
+Find Vertex to Triangle Distances 
+=============
+
+## Group (Subgroup) ##
+Sampling (Spatial)
+
+## Description ##
+This **Filter** computes distances between points in a **Vertex Geoemtry** and triangles in a **Triangle Geoemtry**.  Specifically, for each point in the **Vertex Geometry**, the Euclidean distance to the closest triangle in the **Triangle Geoemtry** is stored.  This distance is *signed*: if the point lies on the side of the triangle to which the triangle normal points, then the distance is positive; otherwise, the distance is negative. Additionally, the ID the closest triangle is stored for each point.
```
imikejackson: Did we ever figure out what is being used on the triangle to measure the distance? Is it the points of the triangle or the centroid? Please note this in the documentation.

src/Plugins/ComplexCore/test/FindVertexToTriangleDistancesTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+inline constexpr StringLiteral k_DistancesName("Distances");
+inline constexpr StringLiteral k_ClosestTriangleIdsName("ClosestTriangleId");
+inline constexpr StringLiteral k_DistancesNameNX("DistancesNX");
+inline constexpr StringLiteral k_ClosestTriangleIdsNameNX("Closest Triangle Ids");
+} // namespace
+
+TEST_CASE("ComplexCore::FindVertexToTriangleDistancesFilter: Valid filter execution", "[ComplexCore][FindVertexToTriangleDistancesFilter]")
+{
+  // Read the Small IN100 Data set
```
imikejackson, line 22: Are we really reading the Small IN100 Dataset?

src/Plugins/ComplexCore/src/ComplexCore/Filters/FindVertexToTriangleDistancesFilter.cpp
```diff
@@ -0,0 +1,146 @@
...
+//------------------------------------------------------------------------------
+std::string FindVertexToTriangleDistancesFilter::humanName() const
+{
+  return "Find Vertex to Triangle Distances";
+}
+
+//------------------------------------------------------------------------------
+std::vector<std::string> FindVertexToTriangleDistancesFilter::defaultTags() const
+{
+  return {"#Sampling", "#Spatial"};
```
imikejackson, line 50: Add the following tags: #triangle #vertex #distance

src/Plugins/ComplexCore/src/ComplexCore/Filters/FindVertexToTriangleDistancesFilter.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+//------------------------------------------------------------------------------
+Parameters FindVertexToTriangleDistancesFilter::parameters() const
+{
+  Parameters params;
+
+  // Create the parameter descriptors that are needed for this filter
+  params.insertSeparator(Parameters::Separator{"Required Input Arrays"});
```
imikejackson, line 59: This should say "Required Geometries"

src/Plugins/ComplexCore/src/ComplexCore/Filters/FindVertexToTriangleDistancesFilter.cpp
```diff
@@ -0,0 +1,146 @@
...
+  params.insert(std::make_unique<GeometrySelectionParameter>(k_VertexDataContainer_Key, "Source Vertex Geometry", "The Vertex Geometry point cloud to map to triangles", DataPath{},
+                                                             GeometrySelectionParameter::AllowedTypes{IGeometry::Type::Vertex}));
+  params.insert(std::make_unique<GeometrySelectionParameter>(k_TriangleDataContainer_Key, "Target Triangle Geometry", "The triangle geometry to compare against", DataPath{},
+                                                             GeometrySelectionParameter::AllowedTypes{IGeometry::Type::Triangle}));
+  params.insert(
+      std::make_unique<ArraySelectionParameter>(k_TriangleNormalsArrayPath_Key, "Triangle Normals", "The triangle geometry's normals array", DataPath{}, std::set<DataType>{DataType::float64}));
+
+  params.insertSeparator(Parameters::Separator{"Created Output Arrays"});
+  params.insert(std::make_unique<ArrayCreationParameter>(k_DistancesArrayPath_Key, "Distances Array", "The array to store distance between vertex and triangle", DataPath{}));
```
imikejackson, line 68: These might be better served as DataObjectNameParameter and then in the Preflight you can create the output arrays in the proper vertex attribute matrix that you get from the VertexGeometry.

imikejackson (changes requested):

imikejackson:
the algorithm is based on the code at https://github.com/embree/embree/blob/master/tutorials/common/math/closest_point.h with some additional code to determine if the point is above or below the triangle where above means on the side of the triangle that the normal is pointing.

The filter as ported has a number of issues. 
* There is a lot of unnecessary copying of triangles and vertices into temporary std::vector<std::vector<T>> objects.
* The interaction with the `TriangleBounds` is a sort of poor mans way of trying to quickly filter out points that are not even close to the triangle but has a failure mode that isn't tested or taken into account. For example if you have a single triangle where the vertices are all in the positive quadrant and the point you are trying to find the distance is (-1, -1, -1) then the distance is reported as std::max<float> because it will never drop into the conditional that checks if the point is within _any_ of the triangle edge bounds.

The whole algorithm would be better served using the RTree algorithm to quickly find a set of closest triangles and then just find the minimal distance to those triangles.

imikejackson (changes requested):
Lots of issues with the original algorithm. Take a look at the conversation and we can figure out how to rewrite the algorithm

PR #429, opened 2023-01-21
CMAKE: Only the ComplexCore plugin depends on Reproc++

imikejackson:


src/Plugins/ComplexCore/CMakeLists.txt
```diff
@@ -141,13 +141,16 @@ add_subdirectory(${${PLUGIN_NAME}_SOURCE_DIR}/test)
 #------------------------------------------------------------------------------
 # If there are additional library packages that need to be found, there is where
 # to do that using the usual find_package(....) call
-# find_package(....)
-
+# ------------------------------------------------------------------------------
+# The ExecuteProcess Filter requires the `reproc` library
+# ------------------------------------------------------------------------------
+find_package("reproc" CONFIG REQUIRED)
```
JDuffeyBQ:
```suggestion
```

JDuffeyBQ (changes requested):

src/Plugins/ComplexCore/CMakeLists.txt
imikejackson: Look for the reproc++ library instead. Fixed

JDuffeyBQ (approved):

PR #430, opened 2023-01-23
Set the debug postfix in 1 spot and use that variable throughout cmake

imikejackson:


src/complex/Core/Application.cpp
```diff
@@ -141,6 +141,13 @@ std::vector<Uuid> Application::getSimplUuid(const Uuid& complexUuid)
   return uuidList;
 }
 
+inline bool ends_with(std::string const& value, std::string const& ending)
```
JDuffeyBQ:
```suggestion
bool ends_with(std::string_view value, std::string_view ending)
```
Maybe we should add this to the string utilities file.

src/complex/Core/Application.cpp
```diff
@@ -141,6 +141,13 @@ std::vector<Uuid> Application::getSimplUuid(const Uuid& complexUuid)
...
+{
+  if(ending.size() > value.size())
+    return false;
```
JDuffeyBQ:
```suggestion
  if(ending.size() > value.size())
  {
    return false;
  }
```

src/complex/Core/Application.cpp
```diff
@@ -151,7 +158,12 @@ void Application::loadPlugins(const std::filesystem::path& pluginDir, bool verbo
   {
     std::filesystem::path path = entry.path();
     std::string extension = path.extension().string();
-    if(extension == ".complex")
+    std::string fileName = path.filename().string();
```
JDuffeyBQ:
```suggestion
```

src/PipelineRunner/CMakeLists.txt
```diff
@@ -18,8 +18,8 @@ add_executable(PipelineRunner)
 
 set_target_properties(PipelineRunner
   PROPERTIES
-    DEBUG_POSTFIX "_d"
-)
+  DEBUG_POSTFIX "${COMPLEX_DEBUG_POSTFIX}"
+  )
```
JDuffeyBQ:
```suggestion
)
```

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #432, opened 2023-01-25
Vec3 basic array operations added

VKUDRI:
Basic vector 3x1 operations such as add/subtract/multiply/divide/dot/cross were added.

### DIFF
--- a/src/complex/Common/Array.hpp
+++ b/src/complex/Common/Array.hpp
@@ -650,10 +650,10 @@ public:
     return Vec3<T>((*this)[1] * v[2] - (*this)[2] * v[1], (*this)[2] * v[0] - (*this)[0] * v[2], (*this)[0] * v[1] - (*this)[1] * v[0]);
   }
 
-    /**
-   * @brief Returns the dotproduct of this and an input Vec3
-   * @return Vec3<T>
-   */
+  /**
+  * @brief Returns the dotproduct of this and an input Vec3
+  * @return Vec3<T>
+  */
   inline T dotProduct(const Vec3<T>& v) const
   {
     return (*this)[0] * v[0] + (*this)[1] * v[1] + (*this)[2] * v[2];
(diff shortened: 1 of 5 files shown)


src/complex/Common/Array.hpp
```diff
@@ -640,6 +641,63 @@ class Vec3 : public Array<T, 3>
...
+   */
+  inline T dot(const Vec3<T>& v) const
+  {
+    return (*this)[0] * v[0] + (*this)[1] * v[1] + (*this)[2] * v[2];
+  }
+
+  /**
+   * @brief Returns the magnitude of the 3x1 vector
+   */
+  inline T magnitude()
```
JDuffeyBQ, line 666:
```suggestion
  inline T magnitude() const
```

src/complex/Common/Array.hpp
```diff
@@ -640,6 +641,63 @@ class Vec3 : public Array<T, 3>
...
+  inline Vec3& operator/=(const T& r)
+  {
+    (*this)[0] /= r, (*this)[1] /= r, (*this)[2] /= r;
+    return *this;
+  }
+  inline Vec3& operator*=(const T& r)
+  {
+    (*this)[0] *= r, (*this)[1] *= r, (*this)[2] *= r;
+    return *this;
+  }
```
JDuffeyBQ: Split comma operator usage into multiple lines.

JDuffeyBQ: For primitive type `T`, pass by value.

JDuffeyBQ (changes requested):
How much are we doing with the matrix class? I'd want to avoid turning it into Eigen.

src/complex/Common/Array.hpp
imikejackson: There are some very basic vector3 operations that complex/dream3d tend to do and this is one of them so I'll let this through.

src/complex/Common/Array.hpp
imikejackson: Done

imikejackson: Done

imikejackson (approved):

JDuffeyBQ:
Are we keeping the matrix classes?

PR #434, opened 2023-01-26
REPROC: Add option to disable the ExecuteProcess filter

imikejackson:
This allows the developer to disable the ExecuteProcess filter which disables the need for the reproc++ library.

There are real issues on all 3 platforms compiling against the reproc++ library.

Linux and MacOS: Since VCPKG does not work in any meaningful way we have to build using the NXSuperbuild. While the NXSuperbuild has been updated to build reproc++ there are configuration issues that those platforms are facing.
Windows with NXSuperbuild (VCPKG isn't used on the Azure self hosted build bots): There are link errors against the reproc library which I think are related to static/dynamic/multithreaded msvc runtime libraries.

### DIFF
--- a/src/Plugins/ComplexCore/CMakeLists.txt
+++ b/src/Plugins/ComplexCore/CMakeLists.txt
@@ -31,7 +31,6 @@ set(FilterList
   CropImageGeometry
   CropVertexGeometry
   DeleteData
-  ExecuteProcessFilter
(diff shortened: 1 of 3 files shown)


imikejackson:
Closing this as this isn't necessary anymore, but it does serve as a template if we want this functionality in the future.

PR #442, opened 2023-02-09
Bug Fix: Better error checking for node-based geometries in PartitionGeometry filter

joeykleingers:


nyoungbq (approved):
Looks good

nyoungbq (approved):
Good idea to return a result instead of optional. Much easier to deal with.

PR #445, opened 2023-02-13
ENH: Expose the C++ Classname of AbstractPipelineNode subclasses.

imikejackson:


imikejackson:
Better ways to do this

PR #446, opened 2023-02-14
ENH: Updated CLI interface

mmarineBlueQuartz:
## Changes:

* Renamed PipelineRunner to NX-cli.
* Restructured parsing for parameters.
  * Changed syntax for preflight and execute.
  * Added syntax for help.
* Renamed namespace from PipelineRunner to CLI

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/DREAM3DNX-cli/src/DREAM3DNX_cli.cpp
```diff
@@ -0,0 +1,280 @@
...
+    args.emplace_back(ArgumentType::Preflight, argStr);
+  }
+
+  if(argc > index)
+  {
+    std::string ss = "Too many arguments to parse";
+    return complex::MakeErrorResult<CliArguments>(k_FailedParsingArguments, ss);
+  }
+
+  return {args};
```
joeykleingers: Unless I'm missing it, this algorithm should have a default case that handles when the argument is unrecognized/unexpected.  For example, if argc = 2 and argv = [executable_path, "--blah"], then this code will go all the way through without setting args to anything or erroring out, and then will return empty args.

src/DREAM3DNX-cli/src/DREAM3DNX_cli.cpp
```diff
@@ -0,0 +1,280 @@
...
+  std::cout << fmt::format("Preflighting pipeline at path: '{}'\n", pipelinePath) << std::endl;
+
+  Pipeline pipeline = loadPipelineResult.value();
+  return PreflightPipeline(pipeline);
+}
+
+void DisplayDefaultHelp()
+{
+  std::cout << "Options:\n";
+  std::cout << fmt::format("\t {} <filepath>\t", k_ExecuteParamLong) << "\t Execute the pipeline at the taget filepath\n";
```
joeykleingers:
```suggestion
  std::cout << fmt::format("\t {} <filepath>\t", k_ExecuteParamLong) << "\t Execute the pipeline at the target filepath\n";
```

src/DREAM3DNX-cli/src/DREAM3DNX_cli.cpp
```diff
@@ -0,0 +1,280 @@
...
+  std::cout << fmt::format("\t {} <filepath>\t", k_ExecuteParamShort) << "\t Execute the pipeline at the taget filepath\n";
```
joeykleingers:
```suggestion
  std::cout << fmt::format("\t {} <filepath>\t", k_ExecuteParamShort) << "\t Execute the pipeline at the target filepath\n";
```

src/DREAM3DNX-cli/src/DREAM3DNX_cli.cpp
```diff
@@ -0,0 +1,280 @@
...
+  std::cout << fmt::format("\t {} <filepath>\t", k_PreflightParamLong) << "\t Preflight the pipeline at the taget filepath" << std::endl;
```
joeykleingers:
```suggestion
  std::cout << fmt::format("\t {} <filepath>\t", k_PreflightParamLong) << "\t Preflight the pipeline at the target filepath" << std::endl;
```

src/DREAM3DNX-cli/src/DREAM3DNX_cli.cpp
```diff
@@ -0,0 +1,280 @@
...
+  std::cout << fmt::format("\t {} <filepath>\t", k_PreflightParamShort) << "\t Preflight the pipeline at the taget filepath" << std::endl;
```
joeykleingers:
```suggestion
  std::cout << fmt::format("\t {} <filepath>\t", k_PreflightParamShort) << "\t Preflight the pipeline at the target filepath" << std::endl;
```

joeykleingers (changes requested):

imikejackson:
@mmarineBlueQuartz I would also like to have the option "--logfile" so that all output can be redirected to a file.

imikejackson:
in the context of the flow of events, if possible, I would recommend that the arguments are all figured out, then the plugins are loaded and the pipeline executed. 

Take a look at my changes/additions and see what you think

imikejackson (approved):

imikejackson (approved):

imikejackson (approved):

PR #447, opened 2023-02-15
BUG: Fix magnitude calculation.

imikejackson:
Remove Matrix class


src/complex/Common/Array.hpp
```diff
@@ -665,7 +665,7 @@ class Vec3 : public Array<T, 3>
    */
   inline T magnitude()
```
JDuffeyBQ:
```suggestion
  inline T magnitude() const
```

JDuffeyBQ (changes requested):

src/complex/Common/Array.hpp
imikejackson: Done

PR #454, opened 2023-02-16
BUG: Fixes to various filters for crashing, parameter layout, initialization

imikejackson:


joeykleingers (approved):
Looks reasonable to me.

PR #462, opened 2023-02-17
FILTER : FindBiasedFeatures

jmarquisbq:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


imikejackson (approved):

imikejackson:
I was thinking this filter needed some sort of progress feedback but my guess is that it runs fast enough that it isn't needed? 

Also this _might_ be parallelized, but I'm not not sure that is worth it at this point in time.

jmarquisbq:
@imikejackson There is some feedback in the 3D version for updating the user on which phase its on. But yeah every time I've run it, it goes pretty quickly so I'm not sure how much we'd really gain from adding more feedback. I can go ahead and do the parallelization if you'd like but like I said, the data sets I was running (SmallIN100) were going pretty fast so not sure what I'd test it on to actually make reasonable use of that feature.

nyoungbq (approved):

imikejackson:
@jmarquisbq Agreed on both topics. We can always circle back to this filter if real users are finding it difficult to track the progress or the data sets get so large that we need this kind of feedback.

PR #472, opened 2023-03-01
BUG: Add example pipelines to the unit tests

imikejackson:
Moves example pipelines to the appropriate plugins and adds each to the unit test suite


imikejackson:
Related PR:

https://github.com/BlueQuartzSoftware/complex_plugins/pull/105
https://github.com/BlueQuartzSoftware/DREAM3DNX/pull/340

nyoungbq:
Major bug: The `ComplexCore Prebuilt Pipeline 3`, also known as the `Import_ASCII.d3dpipeline`, seems to have an invalid filter *UUID* for `CreateEnsembleInfoFilter` or the filter was not accessible for some reason

Details: see line 371 of build (ubuntu-20.04, g++-9)

imikejackson:
That example pipeline uses orientation analysis which isn't compiled by the CI...

nyoungbq (approved):

imikejackson:
The test driver bat and bash files can be replaced with better use of add_test()

https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-tests

[2:12](https://bluequartzsoftware.slack.com/archives/DCQEANJM6/p1678216367689429)
https://cmake.org/cmake/help/latest/command/add_test.html

[2:13](https://bluequartzsoftware.slack.com/archives/DCQEANJM6/p1678216381186459)
https://cmake.org/cmake/help/latest/prop_test/DEPENDS.html

nyoungbq:
@imikejackson I looked over the code, and I think that the execute process changes could be have adverse effects across the different operating systems since its being forced to absolute. Once you get the execute process test working on at least one of the operating systems we can debug further [EDIT: It looks like its passing on windows...which probably means it is a filesystem issue since they are  most lenient]. Also to clarify, are the windows bots' fails expected in this case (like they are dependent on another plugin)?

PR #506, opened 2023-03-24
FILTER: ImportH5OimData & ImportH5EspritData

jmarquisbq:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/ImportH5Data.hpp
```diff
@@ -0,0 +1,58 @@
...
+public:
+  ImportH5Data(DataStructure& dataStructure, const IFilter::MessageHandler& mesgHandler, const std::atomic_bool& shouldCancel, const ImportH5DataInputValues* inputValues);
+  virtual ~ImportH5Data() noexcept;
+
+  ImportH5Data(const ImportH5Data&) = delete;
+  ImportH5Data(ImportH5Data&&) noexcept = delete;
+  ImportH5Data& operator=(const ImportH5Data&) = delete;
+  ImportH5Data& operator=(ImportH5Data&&) noexcept = delete;
+
+  Result<> execute();
```
imikejackson: These need to be fully declared with the template also which means the code will have to come into the header. Or have all of the details on the inside of the class. macOS and Linux will not compile this code.

I think there are some `external` shenanigans that you can do to keep the implementation details out of the header.

imikejackson (changes requested):

imikejackson (approved):

PR #512, opened 2023-03-28
ENH: Feature/data store format param

mmarineBlueQuartz:
Built on top of bluequartzsoftware/complex#507 improved iterators.

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/complex/Parameters/DataStoreFormatParameter.cpp
```diff
@@ -0,0 +1,81 @@
...
+Result<std::any> DataStoreFormatParameter::fromJson(const nlohmann::json& json) const
+{
+  static constexpr StringLiteral prefix = "FilterParameter 'DataStoreFormatParameter' JSON Error: ";
+  if(!json.is_string())
+  {
+    return MakeErrorResult<std::any>(-2, fmt::format("{}JSON value for key '{}' is not a string", prefix, name()));
+  }
+  auto string = json.get<ValueType>();
+  return {{string}};
+}
```
nyoungbq:
```suggestion
{
  if(!json.is_string())
  {
    return MakeErrorResult<std::any>(-2, fmt::format("FilterParameter 'DataStoreFormatParameter' JSON Error: JSON value for key '{}' is not a string", name()));
  }
  return {{json.get<ValueType>()}};
}
```
no reason for excess copies just inline the r-values

src/complex/Parameters/DataStoreFormatParameter.cpp
```diff
@@ -0,0 +1,81 @@
...
+IParameter::AcceptedTypes DataStoreFormatParameter::acceptedTypes() const
+{
+  return {typeid(ValueType)};
+}
+
+nlohmann::json DataStoreFormatParameter::toJson(const std::any& value) const
+{
+  const auto& stringValue = GetAnyRef<ValueType>(value);
+  nlohmann::json json = stringValue;
+  return json;
```
nyoungbq:
```suggestion
  return GetAnyRef<ValueType>(value);
```
just return the value directly

nyoungbq:
Additionally I messaged you on slack about a change related to the `addDefaultValue()` API

PR #513, opened 2023-03-29
FILTER: AppendImageGeometryZSlice

jmarquisbq:
New feature from original filter : you can now save the appended geometry to a separate geometry object instead of editing the original destination geometry.

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/complex/Utilities/DataArrayUtilities.hpp
+++ b/src/complex/Utilities/DataArrayUtilities.hpp
@@ -976,12 +976,12 @@ public:
 
 private:
   template <class K>
-  static void AppendData(K* inputArray, K* destArray, usize offset)
+  static void AppendData(const K* inputArray, K* destArray, usize offset)
   {
     const usize numElements = inputArray->getNumberOfTuples() * inputArray->getNumberOfComponents();
     for(usize i = 0; i < numElements; ++i)
     {
-      (*destArray)[offset + i] = (*inputArray)[i];
+      (*destArray)[offset + i] = inputArray->at(i);
     }
   }
 
@@ -989,7 +989,7 @@ private:
   class AppendArray
   {
   public:
-    AppendArray(IArray& destCellArray, IArray& inputCellArray, usize tupleOffset)
+    AppendArray(IArray& destCellArray, const IArray& inputCellArray, usize tupleOffset)
     : m_ArrayType(destCellArray.getArrayType())
     , m_InputCellArray(&inputCellArray)
     , m_DestCellArray(&destCellArray)
@@ -1016,22 +1016,22 @@ private:
         {
           destArray->addEntry(destArray->getNumberOfTuples() - 1, 0);
         }
-        AppendData<NeighborListT>(dynamic_cast<NeighborListT*>(m_InputCellArray), destArray, offset);
+        AppendData<NeighborListT>(dynamic_cast<const NeighborListT*>(m_InputCellArray), destArray, offset);
       }
       if(m_ArrayType == IArray::ArrayType::DataArray)
       {
         using DataArrayT = DataArray<T>;
-        AppendData<DataArrayT>(dynamic_cast<DataArrayT*>(m_InputCellArray), dynamic_cast<DataArrayT*>(m_DestCellArray), offset);
+        AppendData<DataArrayT>(dynamic_cast<const DataArrayT*>(m_InputCellArray), dynamic_cast<DataArrayT*>(m_DestCellArray), offset);
       }
       if(m_ArrayType == IArray::ArrayType::StringArray)
       {
-        AppendData<StringArray>(dynamic_cast<StringArray*>(m_InputCellArray), dynamic_cast<StringArray*>(m_DestCellArray), offset);
+        AppendData<StringArray>(dynamic_cast<const StringArray*>(m_InputCellArray), dynamic_cast<StringArray*>(m_DestCellArray), offset);
       }
     }
 
   private:
     IArray::ArrayType m_ArrayType = IArray::ArrayType::Any;
-    IArray* m_InputCellArray = nullptr;
+    const IArray* m_InputCellArray = nullptr;
     IArray* m_DestCellArray = nullptr;
     usize m_TupleOffset;
   };
@@ -1040,7 +1040,7 @@ private:
   class CombineArrays
   {
   public:
-    CombineArrays(IArray& inputCellArray1, IArray& inputCellArray2, IArray& destCellArray)
+    CombineArrays(const IArray& inputCellArray1, const IArray& inputCellArray2, IArray& destCellArray)
     : m_ArrayType(destCellArray.getArrayType())
     , m_InputCellArray1(&inputCellArray1)
     , m_InputCellArray2(&inputCellArray2)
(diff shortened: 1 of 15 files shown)


jmarquisbq:
@nyoungbq @JDuffeyBQ I spent some time trying to figure out how to better abstract the algorithm class using templates but can't seem to get things to where I want them because NeighborList doesn't accept bool types but DataArray does. If either of you has a minute and knows of a way I could template this so that we don't have to have so many essentially duplicated impl classes at the top of the AppendImageGeometryZSlice.cpp file then I would love to know (and be able to possible move some of that to a utility file for reusability).

nyoungbq (approved):

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+    }
+
+  private:
+    IArray::ArrayType m_ArrayType = IArray::ArrayType::Any;
+    IArray* m_InputCellArray1 = nullptr;
+    IArray* m_InputCellArray2 = nullptr;
+    IArray* m_DestCellArray = nullptr;
+  };
+
+  static void RunAppendBoolAppend(IArray* destCellArray, IArray* inputCellArray, usize tupleOffset)
```
JDuffeyBQ:
```suggestion
  inline void RunAppendBoolAppend(IArray* destCellArray, IArray* inputCellArray, usize tupleOffset)
```

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+  {
+    using dArrayT = DataArray<bool>;
+    const usize offset = tupleOffset * destCellArray->getNumberOfComponents();
+    AppendData<dArrayT>(dynamic_cast<dArrayT*>(inputCellArray), dynamic_cast<dArrayT*>(destCellArray), offset);
+  }
+
+  static void RunCombineBoolAppend(IArray* inputCellArray1, IArray* inputCellArray2, IArray* destCellArray)
```
JDuffeyBQ:
```suggestion
  inline void RunCombineBoolAppend(IArray* inputCellArray1, IArray* inputCellArray2, IArray* destCellArray)
```

src/complex/Utilities/ParallelAlgorithmUtilities.hpp
```diff
@@ -109,7 +109,5 @@ auto ExecuteParallelFunction(DataType dataType, ParallelRunnerT&& runner, ArgsT&
       return runner.template execute<>(ClassT<float64>(std::forward<ArgsT>(args)...));
     }
   }
-
-  throw std::runtime_error("complex::ExecuteParallelFunction has all supported DataTypes disabled. Error: Invalid DataType");
```
JDuffeyBQ, line 113: Should this be deleted?

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+ * @code
+ *
+ * @endcode
+ */
+class CopyFromArray
+{
+public:
+  CopyFromArray()
+  {
+  }
```
JDuffeyBQ:
```suggestion
  CopyFromArray() = default;
```

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+ *     auto* newDataArray = dataStructure.getDataAs<IArray>(newCellDataPath.createChildPath(name));
+ *     const IArray::ArrayType arrayType = destDataArray->getArrayType();
+ *     CopyFromArray::RunParallel<CopyFromArray::Combine>(arrayType, destDataArray, taskRunner, inputDataArray, newDataArray);
+ *   }
+ *   taskRunner.wait();
+ * @code
+ *
+ * @endcode
+ */
+class CopyFromArray
```
JDuffeyBQ: Does this need to be a class?

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+    }
+    if constexpr(TemplateTypeOptions::UsingAppend)
+    {
+      ExecuteParallelFunction<AppendArray, NoBooleanType>(dataType, std::forward<ParallelRunnerT>(runner), arrayType, destArray, std::forward<ArgsT>(args)...);
+    }
+  }
+
+private:
+  template <class K>
+  static void AppendData(K* inputArray, K* destArray, usize offset)
```
JDuffeyBQ: This could probably be a free function.

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+      auto value = (*inputArray)[i]; // make sure we are getting a copy not a ref
+      (*destArray)[offset + i] = value;
+    }
+  }
+
+  template <typename T>
+  class AppendArray
+  {
+  public:
+    AppendArray(IArray::ArrayType arrayType, IArray* destCellArray, IArray* inputCellArray, usize tupleOffset)
```
JDuffeyBQ: Would it make sense for the array type to be inferred from the first array? And I think that all these functions should take in there arrays by reference. They'll still be stored in a pointer, but we can protect the interface a bit more. Also arrays that are only read from should be const.

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+  }
+
+private:
+  template <class K>
+  static void AppendData(K* inputArray, K* destArray, usize offset)
+  {
+    const usize numElements = inputArray->getNumberOfTuples() * inputArray->getNumberOfComponents();
+    for(usize i = 0; i < numElements; ++i)
+    {
+      auto value = (*inputArray)[i]; // make sure we are getting a copy not a ref
```
JDuffeyBQ: Out of curiosity, when would this get a reference and be an issue?

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+  {
+    static inline constexpr bool UsingCombine = UseCombine;
+    static inline constexpr bool UsingAppend = UseAppend;
+  };
+
+  using Combine = TemplateTypeOptions<true, false>;
+  using Append = TemplateTypeOptions<false, true>;
+
+  template <class TemplateTypeOptions = Combine, class ParallelRunnerT, class... ArgsT>
+  static void RunParallel(IArray::ArrayType arrayType, IArray* destArray, ParallelRunnerT&& runner, ArgsT&&... args)
```
JDuffeyBQ: If combine and append cannot be used together, I feel like it would make sense to keep them separate.

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -904,4 +906,215 @@ class CopyTupleUsingIndexList
...
+    AppendArray(AppendArray&&) noexcept = default;
+    AppendArray& operator=(const AppendArray&) = delete;
+    AppendArray& operator=(AppendArray&&) noexcept = delete;
+
+    void operator()() const
+    {
+      const usize offset = m_TupleOffset * m_DestCellArray->getNumberOfComponents();
+      if(m_ArrayType == IArray::ArrayType::NeighborListArray)
+      {
+        using nListT = NeighborList<T>;
```
JDuffeyBQ: For type aliases, I would use something like `ArrayT` or `ArrayType` to be consistent with our other aliases.

JDuffeyBQ (changes requested):

src/complex/Utilities/ParallelAlgorithmUtilities.hpp
jmarquisbq: No I don't think so, probably just mistakenly got committed after we were playing around with how to template out the stuff I needed.

src/complex/Utilities/DataArrayUtilities.hpp
jmarquisbq: @JDuffeyBQ I was trying to hide some of the functions so that the user has to go through the proper checks in order to do the actual copying (specifically with the AppendData function below).  Basically, in the case that we want to save the appended to a new geometry, you use the create actions in preflight to create arrays/neighborlists/string arrays for each valid cell data array.  The CreateNeighborListAction doesn't allocate so you could get down to that AppendData function and the `[]` operator for NeighborList tries to dereference a nullptr.

src/complex/Utilities/ParallelAlgorithmUtilities.hpp
nyoungbq: @JDuffeyBQ I would argue so. it's a runtime check for something that is determined at compile time this would be more of an annoyance for users than anything a static_assert would be preferred, but that would be an incredibly long line i think.

src/complex/Utilities/DataArrayUtilities.hpp
jmarquisbq: @JDuffeyBQ You know I don't honestly remember what I was think when I put this in. I think this edit was one of my earlier changes and must have thought one of the bracket operators was doing something else...

src/complex/Utilities/DataArrayUtilities.hpp
jmarquisbq: @JDuffeyBQ @nyoungbq and I went back and forth on that one. We were trying to avoid duplicating more code but I can an argument either way.

src/complex/Utilities/DataArrayUtilities.hpp
JDuffeyBQ: Sorry I should have been more specific. I was thinking that it could be a namespace instead since it doesn't store any data.

src/complex/Utilities/DataArrayUtilities.hpp
JDuffeyBQ: Is it possible to have two functions that share most of the important code in one function?

src/complex/Utilities/DataArrayUtilities.hpp
jmarquisbq: @JDuffeyBQ Yeah I get why we'd want to do a namespace instead. I was mostly just thinking that it makes it more clear to the user how the functions should be used together. I though it might be easy for a user of the code to see the AppendData function signature and try to use that without proper bounds checking and all the other steps to set up that function but maybe that's okay.  I've got a big comment at the top of this section explaining how it's meant to be used and if I comment the AppendData function well enough too then we can just put that on the caller to make sure they are using it properly.

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -893,4 +895,215 @@ class CopyTupleUsingIndexList
...
+ *
+ * @endcode
+ */
+namespace CopyFromArray
+{
+/**
+ * @brief Appends all of the data from the inputArray into the destination array starting at the given offset. This function DOES NOT do any bounds checking!
+ */
+template <class K>
+static void AppendData(const K* inputArray, K* destArray, usize offset)
```
JDuffeyBQ:
```suggestion
void AppendData(const K& inputArray, K& destArray, usize offset)
```
I think this is the only other change I would make.

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #515, opened 2023-03-29
Reverse Commit of ThreadSafeMessenger #495

imikejackson:
Revert the Thread Safe Messenger due to data races uncovered by ThreadSanitizer and anecdotally by other GitHub CI


CMakeLists.txt
```diff
@@ -618,9 +617,7 @@ set(COMPLEX_SRCS
   ${COMPLEX_SOURCE_DIR}/Utilities/OStreamUtilities.cpp
   ${COMPLEX_SOURCE_DIR}/Utilities/GeometryHelpers.cpp
   ${COMPLEX_SOURCE_DIR}/Utilities/ColorPresetsUtilities.cpp
-  ${COMPLEX_SOURCE_DIR}/Utilities/ImageRotationUtilities.cpp
```
nyoungbq, line 621: Was Image rotation utilities meant to be deleted too?

imikejackson: Probably not?

JDuffeyBQ: The original `ImageRotationUtilities` line was not aligned with the rest so the commit from #495 moved it to be in line with the rest.

JDuffeyBQ (approved):

PR #520, opened 2023-03-30
TEST: Create tests for the Orientation Analysis Example Pipelines

imikejackson:
The pipelines are run in the correct order and the output files are removed as the example pipelines are executed.

### DIFF
--- a/src/Plugins/OrientationAnalysis/test/CMakeLists.txt
+++ b/src/Plugins/OrientationAnalysis/test/CMakeLists.txt
@@ -174,7 +174,7 @@ function(add_pipeline_test)
 
   get_filename_component(test "${ARGS_PIPELINE_PATH}" NAME)
   set(SIMPL_CTEST_TEST_DRIVER "${${PLUGIN_NAME}_BINARY_DIR}/PrebuiltPipelineTests/${test}.${TEST_SCRIPT_FILE_EXT}")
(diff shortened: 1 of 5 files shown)


src/Plugins/OrientationAnalysis/test/CMakeLists.txt
```diff
@@ -139,20 +139,134 @@ endif()
...
+
+#------------------------------------------------------------------------------
+# This function creates a CTest for a given pipeiline file.
+#  PLUGIN_NAME: Name of the plugin: Typically you can use ${PLUGIN_NAME}
+#  PIPELINE_PATH: The absolute path to the pipeline file
+#  TEST_INDEX: The numerical test index. This can be reset for each plugin
+#  DELETE_FILE_LIST: The list of files that should be deleted from the system
+#                    after the test is run
+#------------------------------------------------------------------------------
+function(add_pipeline_test)
```
JDuffeyBQ, line 157: Don't we have a function for this somewhere?

imikejackson: We do but it uses ctest to drive the pipeline test instead of generating a batch/bash file that gets executed. We need the batch/bash so that we can remove some data files that are no longer needed.

JDuffeyBQ (approved):

PR #525, opened 2023-04-05
FILTER: Nearest Point Fuse Regular Grids

nyoungbq:
Ready for review

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/NearestPointFuseRegularGrids.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/NearestPointFuseRegularGrids.cpp
@@ -11,8 +11,8 @@ using namespace complex;
 
 namespace
 {
-template <class K>
-void CopyData(const K& inputArray, K& destArray, const ImageGeom& sampleImageGeom, const ImageGeom& refImageGeom, const std::atomic_bool& shouldCancel)
+template <class K, typename T>
+void CopyData(const K& inputArray, K& destArray, const ImageGeom& sampleImageGeom, const ImageGeom& refImageGeom, const std::atomic_bool& shouldCancel, T fillValue)
 {
   // Get dimensions and resolutions of two grids
   Vec3<float32> sampleRes = sampleImageGeom.getSpacing();
@@ -27,6 +27,7 @@ void CopyData(const K& inputArray, K& destArray, const ImageGeom& sampleImageGeo
   usize refIndex = 0;
   usize sampleIndex = 0;
   usize planeComp = 0, rowComp = 0;
+  usize numComps = destArray.getNumberOfComponents();
   for(usize i = 0; i < refDims[2]; i++)
   {
     if(shouldCancel)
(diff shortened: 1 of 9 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/NearestPointFuseRegularGrids.cpp
```diff
@@ -0,0 +1,198 @@
...
+      //        destArray.addEntry(destArray.getNumberOfTuples() - 1, 0);
+      //      }
+      //      CopyData<NeighborListT>(dynamic_cast<const NeighborListT&>(m_SourceArray1), destArray);
+    }
+    else if(m_ArrayType == IArray::ArrayType::DataArray)
+    {
+      using DataArrayT = DataArray<T>;
+      auto destArray = dynamic_cast<DataArrayT&>(m_DestArray);
+      destArray.fill(static_cast<T>(m_FillValue));
+      CopyData<DataArrayT>(dynamic_cast<const DataArrayT&>(m_SourceArray), destArray, m_SampleGeom, m_RefGeom, m_ShouldCancel);
```
mmarineBlueQuartz: This will have a performance hit as DataArrays get larger and larger. If `destArray` is larger than `m_SourceArray`, there is no need to fill the first n values (consider `std::fill(destArray.begin() + offset, destArray.end(), m_FillValue)`) and skip the fill value if `m_SourceArray` is larger or equal in size.

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/NearestPointFuseRegularGrids.cpp
```diff
@@ -0,0 +1,198 @@
...
+    {
+      return;
+      //      using NeighborListT = NeighborList<T>;
+      //      auto& destArray = dynamic_cast<NeighborListT&>(m_DestArray);
+      //      // Make sure the destination array is allocated AND each tuple list is initialized, so we can use the [] operator to copy over the data
+      //      if(destArray.getValues().empty() || destArray.getList(0) == nullptr)
+      //      {
+      //        destArray.addEntry(destArray.getNumberOfTuples() - 1, 0);
+      //      }
+      //      CopyData<NeighborListT>(dynamic_cast<const NeighborListT&>(m_SourceArray1), destArray);
```
mmarineBlueQuartz: Is there a reason there is still commented out code here?

```suggestion
```

mmarineBlueQuartz (changes requested):

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/NearestPointFuseRegularGrids.cpp
nyoungbq: @mmarineBlueQuartz the issue is that the entire array must be filled because its not a straight forward copy. One possible outcome could be structured as follows:

Source Array: { 1, 2, 3, 4, 5}
Dest Array: { 0, 0, 0, 0, 0, 0, 0}

With your suggestion one would use your described fill to get the following assuming the fill value is 10:

Source Array: { 1, 2, 3, 4, 5}
Dest Array: { 0, 0, 0, 0, 0, 10, 10}

Leaving the remaining spaces to be filled. This is valid, however the end goal of the filter is to only copy over the overlap between the two arrays based on origin and spacing. So one possible solution could be:

Dest Array: {10, 10, 10, 10, 3, 10, 5}

To fix this I substituted by moving the fill in front of every continue in the copyData function. I am pushing the change now.

mmarineBlueQuartz (approved):

PR #546, opened 2023-04-10
BUG/ENH: Changes to Printing of ASCII Data

### LINKED ISSUE #544: Export ASCII Data: Update the filter parameters a bit
I think what should happen is that if the user selected "Single File" the FileSystemPath parameter should be a "Output File Path"

If the user selects the "Multiple Files" then the FileSystemPath parameter should be "output directory" type. We might need to do this with 2 separate parameters? Then link those parameters to the combo box so that the proper UI elements show up when selected. 

@jmarquisbq @nyoungbq Work together to figure out who wants to tackle this or come up with an alternate design? To me, it seems un-natural to ask for a single file then have to define the path, name and extension.

### LINKED ISSUE #545: FeatureDataCSVWriterFilter: Does not write correct number of features on first line
The output file is not correct when user selects to write the number of features option.

imikejackson:
Also added 2 more prebuilt pipelines that should be tested.


nyoungbq (approved):

src/complex/Utilities/OStreamUtilities.hpp
```diff
@@ -76,13 +76,15 @@ COMPLEX_EXPORT void PrintSingleDataObject(std::ostream& outputStrm, const DataPa
...
+ * @param writeFirstIndex Write the first tuple of the arrays. For Feature based data, this would be FeatureId = 0.
+ * @paragraph indexName The name of the "Index" column
  * @param includeHeaders The boolean that determines if headers are printed
  * @param neighborLists The list of dataPaths of neighborlists to include
  * @param writeNumOfFeatures The amount of elements per tuple printed at top
  */
 COMPLEX_EXPORT void PrintDataSetsToSingleFile(std::ostream& outputStrm, const std::vector<DataPath>& objectPaths, DataStructure& dataStructure, const IFilter::MessageHandler& mesgHandler,
                                               const std::atomic_bool& shouldCancel, const std::string& delimiter = "", bool includeIndex = false, bool includeHeaders = false,
-                                              const std::vector<DataPath>& neighborLists = {}, bool writeNumOfFeatures = false);
+                                              bool writeFirstIndex = true, const std::string& indexName = "Index", const std::vector<DataPath>& neighborLists = {}, bool writeNumOfFeatures = false);
```
nyoungbq, line 87: Make sure this API change is percolated to all the other filters that call this function. That's why the unit tests are failing on every platform

nyoungbq (commented):
After reviewing the failed build bots' unit tests this change is altering the actual output of the WriteASCIIData filter

imikejackson:
This was due to the correction of a bug in the algorithm that was not outputting the correct values to the data file.

@nyoungbq Can you update the exemplar files to get the tests to match?

nyoungbq:
yes

src/complex/Utilities/OStreamUtilities.cpp
```diff
@@ -484,13 +488,15 @@ void PrintDataSetsToSingleFile(std::ostream& outputStrm, const std::vector<DataP
...
     size_t featureCount = 0;
-    if(includeHeaders)
-    {
-      featureCount++;
-    }
-    for(const auto& path : objectPaths)
+    //    if(includeHeaders)
+    //    {
+    //      featureCount++;
+    //    }
```
jmarquisbq:
```suggestion
```

jmarquisbq (changes requested):

PR #547, opened 2023-04-11
Filter: Write Pole Figure

imikejackson:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/TiffWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    return out;
+  }
+};
+
+enum class Endianess
+{
+  Little = 0,
+  Big
+};
```
nyoungbq:
```suggestion
```
There is no reason to redefine this. It is already defined in the `Bit.hpp` file, under the enum `endian`

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/TiffWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+Endianess checkEndianess()
```
nyoungbq: Personally, I think this function should just be moved into the `Bit.hpp` class, and modifed to use the aforementioned pe-existing `endian` enum

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/TiffWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+  constexpr uint32_t nextOffset = 0;
+  outputFile.write(reinterpret_cast<const char*>(&nextOffset), sizeof(nextOffset));
+
+  // Now write the actual image data
+  int32_t imageByteCount = width * height * samplesPerPixel;
+  outputFile.write(reinterpret_cast<const char*>(data), imageByteCount);
+
+  // and we are done.
+  return {0, "No Error"};
+}
```
nyoungbq: Unless I am missing some very subtle but stark differences, I feel like these two functions can be combined into one since there is so much code duplication

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/TiffWriter.cpp
imikejackson: Done

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/TiffWriter.cpp
imikejackson: Done

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/TiffWriter.cpp
imikejackson: Done

nyoungbq (approved):

PR #551, opened 2023-04-11
ENH: Import Ascii Parameter Updates

### LINKED ISSUE #540: Import ASCII Data Array: Allow user to set Tuple Shape or add to existing Attribute Matrix
This filter should basically be the same as the "Create Data Array" with respect to defining the tuple shape or using an existing Attribute Matrix.

nyoungbq:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportTextFilter.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportTextFilter.cpp
@@ -66,7 +66,7 @@ Parameters ImportTextFilter::parameters() const
                                                   "This value will specify which data format is used by the array's data store. An empty string results in in-memory data store.", ""));
 
   params.insertSeparator(Parameters::Separator{"Tuple Handling"});
-  params.insertLinkableParameter(std::make_unique<BoolParameter>(
+  params.insert(std::make_unique<BoolParameter>(
       k_AdvancedOptions_Key, "Set Tuple Dimensions [not required if creating inside an Attribute Matrix]",
       "This allows the user to set the tuple dimensions directly rather than just inheriting them \n\nThis option is NOT required if you are creating the Data Array in an Attribute Matrix", true));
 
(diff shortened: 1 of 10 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportTextFilter.cpp
```diff
@@ -52,12 +55,27 @@ Parameters ImportTextFilter::parameters() const
...
+
+  params.insertSeparator(Parameters::Separator{"Created DataArray"});
   params.insert(std::make_unique<ArrayCreationParameter>(k_DataArrayKey, "Created Array Path", "DataPath or Name for the underlying Data Array", DataPath{}));
+  params.insert(std::make_unique<StringParameter>(k_DataFormat_Key, "Data Format",
+                                                  "This value will specify which data format is used by the array's data store. An empty string results in in-memory data store.", ""));
+
+  params.insertSeparator(Parameters::Separator{"Tuple Handling"});
+  params.insertLinkableParameter(std::make_unique<BoolParameter>(
+      k_AdvancedOptions_Key, "Set Tuple Dimensions [not required if creating inside an Attribute Matrix]",
+      "This allows the user to set the tuple dimensions directly rather than just inheriting them \n\nThis option is NOT required if you are creating the Data Array in an Attribute Matrix", true));
```
jmarquisbq, line 71: @nyoungbq You've inserted this as a linkable parameter but I don't see it actually being linked to anything. Was this on purpose?

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportTextFilter.cpp
```diff
@@ -70,9 +88,50 @@ IFilter::PreflightResult ImportTextFilter::preflightImpl(const DataStructure& da
...
+  }
+  else
+  {
+    tupleDims = parentAM->getShape();
+    if(useDims)
+    {
+      return {ConvertResultTo<OutputActions>(
+          MakeWarningVoidResult(-77604, "You checked Set Tuple Dimensions, but selected a DataPath that has an Attribute Matrix as the parent. The Attribute Matrix tuples will override your "
+                                        "custom dimensions. It is recommended to uncheck Set Tuple Dimensions for the sake of clarity."),
+          {})};
```
jmarquisbq: @nyoungbq So we are only making this a warning however we return from preflight before actually doing the create array action. I feel like if it's only a warning, the action should still get executed.

jmarquisbq (changes requested):

jmarquisbq (approved):

PR #571, opened 2023-04-20
API: Euclidean Geometry Updates

nyoungbq:
- Removed Point2D and Point3D classes -> Now a templated type-alias for `Vec3`
- Updated BoundingBox class to interface with Vec2 and Vec3 and standardized usage throughout code base
- Ported and updated a majority of the GeometryMath class from SIMPL
- Extracted non-templated function definitions out of GeometryUtilities header into a cpp file

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/complex/Common/BoundingBox.hpp
```diff
@@ -2,20 +2,22 @@
...
 
-#include "complex/Common/Point3D.hpp"
+#include "complex/Common/Array.hpp"
 
 namespace complex
 {
-
 /**
  * @class BoundingBox
  * @brief The BoundingBox class is designed to describe a box in 3D space
```
JDuffeyBQ: Since this supports 2D now, the documentation should be updated.

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,29 +47,45 @@ class BoundingBox
...
 
   /**
-   * @brief Constructs a new BoundingBox defined by the array of position values.
+   * @brief Constructs a new BoundingBox3D defined by the array of position values.
    * The format is min X, min Y, min Z, max X, max Y, max Z.
    * @param arr
    */
-  BoundingBox(const std::array<T, 6>& arr)
-  : m_Bounds(arr)
+  explicit BoundingBox(const std::array<T, 6>& arr)
```
JDuffeyBQ:
```suggestion
  template <class = std::enable_if_t<PointType::Dimensions == 6>>
  explicit BoundingBox(const std::array<T, 6>& arr)
```
We should conditionally enable these constructors depending on the point type. This will require a static constexpr member in the `Vec` class.

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,29 +47,45 @@ class BoundingBox
...
+  , m_Upper(Point3D<T>(arr[3], arr[4], arr[5]))
+  {
+  }
+
+  /**
+   * @brief Constructs a new BoundingBox2D defined by the array of position values.
+   * The format is min X, min Y, max X, max Y.
+   * @param arr
+   */
+  explicit BoundingBox(const std::array<T, 4>& arr)
```
JDuffeyBQ:
```suggestion
  template <class = std::enable_if_t<PointType::Dimensions == 4>>
  explicit BoundingBox(const std::array<T, 4>& arr)
```
See previous.

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,29 +47,45 @@ class BoundingBox
...
-   * std::array<T, 6> constructor, min X, min Y, min Z, max X, max Y, max Z.
+   * The pointer is assumed to contain either 4 (2D) or 6 (3D) values formatted the same as the
+   * std::array<T, 6> constructor, min X, min Y, max X, max Y or the
+   * std::array<T, 6> constructor, min X, min Y, min Z, max X, max Y, max Z
+   * respectively
    * @param arr
    */
-  BoundingBox(Pointer arr)
-  : m_Bounds({arr[0], arr[1], arr[2], arr[3], arr[4], arr[5]})
+  explicit BoundingBox(Pointer arr)
```
JDuffeyBQ:
```suggestion
  explicit BoundingBox(const Pointer arr)
```

src/complex/Common/BoundingBox.hpp
```diff
@@ -81,145 +95,62 @@ class BoundingBox
    * @brief Returns the Min point
    * @return
    */
-  std::array<ValueType, 3> getMinPoint() const
+  PointType getMinPoint() const
```
JDuffeyBQ:
```suggestion
  const PointType& getMinPoint() const
```
Since we're changing the class anyways, I'd make the getters return by const ref.

src/complex/Common/BoundingBox.hpp
```diff
@@ -81,145 +95,62 @@ class BoundingBox
...
-    return {m_Bounds[0], m_Bounds[1], m_Bounds[2]};
+    return m_Lower;
   }
 
   /**
    * @brief Returns the Max point
    * @return
    */
-  std::array<ValueType, 3> getMaxPoint() const
+  PointType getMaxPoint() const
```
JDuffeyBQ:
```suggestion
  const PointType& getMaxPoint() const
```
See previous.

src/complex/Common/BoundingBox.hpp
```diff
@@ -81,145 +95,62 @@ class BoundingBox
...
+   * @return
    */
-  void setMaxZ(ValueType value)
+  PointType center() const
   {
-    m_Bounds[5] = value;
+    PointType boxSize = sideLengths();
+    if constexpr(std::is_same_v<Point<ValueType>, Point3D<ValueType>>)
+    {
+      return {m_Lower[0] + boxSize[0] / 2.0, m_Lower[1] + boxSize[1] / 2.0, m_Lower[2] + boxSize[2] / 2.0};
```
JDuffeyBQ: Maybe we should have a division operator on `Vec`. I see we have multiplication already.

src/complex/Common/BoundingBox.hpp
```diff
@@ -269,7 +175,8 @@ class BoundingBox
    */
   BoundingBox& operator=(const BoundingBox& rhs)
   {
-    m_Bounds = rhs.m_Bounds;
+    m_Lower = rhs.m_Lower;
```
JDuffeyBQ: I think we can default all the copy/move constructors/assignment.

src/complex/Utilities/Math/GeometryMath.hpp
```diff
@@ -23,12 +24,18 @@ namespace GeometryMath
...
- * @return float
+ * @return T
  */
 template <typename T>
-T CosThetaBetweenVectors(const complex::Point3D<T>& a, const complex::Point3D<T>& b)
+inline T CosThetaBetweenVectors(const complex::Point3D<T>& a, const complex::Point3D<T>& b)
 {
-  throw std::runtime_error("");
+  T norm1 = sqrt(a.sumOfSquares());
+  T norm2 = sqrt(b.sumOfSquares());
```
JDuffeyBQ:
```suggestion
  T norm1 = std::sqrt(a.sumOfSquares());
  T norm2 = std::sqrt(b.sumOfSquares());
```
All `<cmath>` functions should have the `std::` namespace prefixed.

JDuffeyBQ (changes requested):

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,30 +47,56 @@ class BoundingBox
...
+   * std::array<T, 4> constructor, min X, min Y, max X, max Y or the
+   * std::array<T, 6> constructor, min X, min Y, min Z, max X, max Y, max Z
+   * respectively
    * @param arr
    */
-  BoundingBox(Pointer arr)
-  : m_Bounds({arr[0], arr[1], arr[2], arr[3], arr[4], arr[5]})
+  explicit BoundingBox(const Pointer arr)
   {
+    if(arr->size() == 6)
```
JDuffeyBQ: `Pointer` is just `T` right?

src/complex/Common/Array.hpp
```diff
@@ -439,6 +439,118 @@ class Vec2 : public Array<T, 2>
...
+  inline Vec2 operator*(const T r) const
+  {
+    return Vec2((*this)[0] * r, (*this)[1] * r);
+  }
+  /**
+   * @brief Multiples this Vec2 with another Vec2 using element wise multiplication and returns a new instance
+   * @param Vec2
+   * @return new Vec2 instance with the result
+   */
+  inline Vec2 operator*(const Vec2& v) const
```
JDuffeyBQ: I know `operator*` with `Vec3` was added previously, but I'm thinking it might not be obvious that it's element-wise multiplication. That might be a topic for another PR though.

src/complex/Common/BoundingBox.hpp
nyoungbq: @JDuffeyBQ `Pointer` is `T*` || `ValueType` is `T`

src/complex/Common/Array.hpp
nyoungbq: I don't mind refactoring it to a different operator in this PR what would you suggest the operator should be used to make this operation more clear? Or should I just make this a function such as `elemMulOp()`

JDuffeyBQ: I'd say just a normal function.

src/complex/Common/BoundingBox.hpp
JDuffeyBQ: Then shouldn't this constructor take a `span` instead if we want to check the size at runtime?

nyoungbq: Yes I will fix that

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,30 +49,56 @@ class BoundingBox
...
-   * std::array<T, 6> constructor, min X, min Y, min Z, max X, max Y, max Z.
+   * The pointer is assumed to contain either 4 (2D) or 6 (3D) values formatted the same as the
+   * std::array<T, 4> constructor, min X, min Y, max X, max Y or the
+   * std::array<T, 6> constructor, min X, min Y, min Z, max X, max Y, max Z
+   * respectively
    * @param arr
    */
-  BoundingBox(Pointer arr)
-  : m_Bounds({arr[0], arr[1], arr[2], arr[3], arr[4], arr[5]})
+  explicit BoundingBox(nonstd::span<ValueType> arr)
```
JDuffeyBQ:
```suggestion
  explicit BoundingBox(nonstd::span<const ValueType> arr)
```

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,30 +49,56 @@ class BoundingBox
...
   {
+    if(arr->size() == 6)
```
JDuffeyBQ:
```suggestion
    if(arr.size() == 6)
```

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,30 +49,56 @@ class BoundingBox
...
+    {
+      m_Lower = Point3D<T>(arr[0], arr[1], arr[2]);
+      m_Upper = Point3D<T>(arr[3], arr[4], arr[5]);
+    }
+    else if(arr->size() == 4)
```
JDuffeyBQ:
```suggestion
    else if(arr.size() == 4)
```

src/complex/Common/BoundingBox.hpp
```diff
@@ -49,30 +49,56 @@ class BoundingBox
...
+    {
+      m_Lower = Point3D<T>(arr[0], arr[1]);
+      m_Upper = Point3D<T>(arr[2], arr[3]);
```
JDuffeyBQ: Actually looking at this again, shouldn't these in this branch be `Point2D`? And if so, that means we can't branch at runtime since there are different point types.

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #575, opened 2023-04-28
BUG: Fix issue where ITKImageWriter is double looping over the Z dim

imikejackson:
The issue starts in ITKImageWriter.cpp on in the `executeImp()` function where we figure out which plane we are on, make a copy of the input data array and then copy the data into a temp Data Array (Thus making a COPY of the original data). We then feed that (and the original imageGeom) into the cxITKImageWriter::SaveImageData(...) function. .. which again loops over the complete Z direction. Thus we end up with a LOT of extra images. That function is the `WriteAs2DStack()` function.

That function also will dictate the file name padding with zeros and such. We probably want to take control of this, do the looping ourselves, generate the filename ourselves and then call the `WriteAsOneFile()` function instead.

*Also, the data is not getting copied into the target correctly as every image shows as black, all ZEROs*


imikejackson:
Need to implement a unit test for 3D Image Geometry @imikejackson

PR #577, opened 2023-05-03
Filter:SurfaceNets3 Meshing and Smoothing Filter.

imikejackson:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/SurfaceNets.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/SurfaceNets.cpp
@@ -37,21 +37,21 @@ public:
     for(int i = 0; i < 4; i++)
       vertexIndices[i] = m_vertexIndices[i];
   }
-  void getLabels(unsigned short labels[2])
+  void getLabels(int32_t labels[2])
   {
     for(int i = 0; i < 2; i++)
       labels[i] = m_labels[i];
   }
-  void setVertexIndices(int vertexIndices[4])
-  {
-    for(int i = 0; i < 4; i++)
-      m_vertexIndices[i] = vertexIndices[i];
-  }
-  void setLabels(unsigned short labels[2])
-  {
-    for(int i = 0; i < 2; i++)
-      m_labels[i] = labels[i];
-  }
+  //  void setVertexIndices(int vertexIndices[4])
+  //  {
+  //    for(int i = 0; i < 4; i++)
+  //      m_vertexIndices[i] = vertexIndices[i];
+  //  }
+  //  void setLabels(int32_t labels[2])
+  //  {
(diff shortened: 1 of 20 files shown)


imikejackson:
There are a number of issues with this initial implementation: 
+ The actual coordinate values generated for each vertex is based off of an origin of 0,0,0. We can either update the codes to take the real origin of the input Image Geometry into account, or just apply the proper translation to the final vertices at the end of the algorithm
+ The code only accepts uint16 at this point and needs to be converted to int32 at the minimum. Having a templated version to handle any `integer` type would be fantastic
+ The codes will allocate arrays in old C-Style and uses an Array-of-Structs (AoS) type of allocation. This should be rewritten to use individual arrays for each property in the MMCell class. 
+ I think a 2 pass approach could be used to figure out how many vertices and how many triangles are going to be created. This will save a whole bunch of memory allocations during the actual filter itself.
+ The node types need to be figured out to have better compatibility with existing DREAM.3D Filters.


Filters that use NodeType:
+ SurfaceMeshToVtk
+ SurfaceMeshToNonconformalVtk
+ FindGBCDMetricBased
+ FindGBPDMetricBased
+ LaplacianSmoothing
+ BInaryNodesTriangleReader
+ ExtractInternalSurfacesFromTriangleGeometry
+ ExtractTripleLinesFromTriangleGeometry

src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/SurfaceNets.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+  // Get the Created Triangle Geometry
+  TriangleGeom& triangleGeom = m_DataStructure.getDataRefAs<TriangleGeom>(m_InputValues->TriangleGeometryPath);
+
+  auto gridDimensions = imageGeom.getDimensions();
+  auto voxelSize = imageGeom.getSpacing();
+  IntVec3 arraySize(static_cast<int32>(gridDimensions[0]), static_cast<int32>(gridDimensions[1]), static_cast<int32>(gridDimensions[2]));
+
+  Int32Array& featureIds = m_DataStructure.getDataRefAs<Int32Array>(m_InputValues->FeatureIdsArrayPath);
+  auto& featureIdDataStore = featureIds.getIDataStoreRefAs<Int32DataStore>();
```
mmarineBlueQuartz:
```suggestion
  auto& featureIdDataStore = featureIds.getDataStoreRef();
```
Using `getDataStoreRef()` returns an `AbstractDataStore<T>&` that is expandable to other forms of data storage. If you do not require `DataStore<T>`'s `data()` method, there is no need to force the use of an in-memory implementation.

mmarineBlueQuartz (changes requested):

imikejackson:
replaced by #619

PR #580, opened 2023-05-08
FILTER: Import Deform V12,  KMedoids, KMeans, Silhouette

nyoungbq:
Changes included in this PR:
- ImportDeformFileV12 filter. algorithm, and documentation added
- KMedoids filter, algorithm, and documentation added
- KMeans filter, algorithm, and documentation added
- Silhouette filter algorithm, and documetation added
- API changes to `ParallelAlgorithmUtilities.hpp`, `FilterUtilites.hpp`, and `Types.hpp` to consolidate API to make type-aliases more accessible
- Additional API function added to `FilterUtilites.hpp` for executing template classes
- Added Utilities class for KFilters and algorithms aptly named `KUtilities`
- Added extra error message for when a user tries to create a new object at an existing object's path in `ImportH5ObjectPathsAction.cpp`

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/complex/Utilities/KUtilities.hpp
```diff
@@ -0,0 +1,137 @@
...
+  SquaredPearson
+};
+
+/**
+ * @brief The DistanceTemplate class contains a templated function getDistance to find the distance, via a variety of
+ * metrics, between two vectors of arbitrary dimensions. The developer should ensure that the pointers passed to
+ * getDistance do indeed contain vectors of the same component dimensions and start at the desired tuples.
+ */
+template <typename leftDataType, typename rightDataType>
+auto GetDistance(const leftDataType& leftVector, usize leftOffset, const rightDataType& rightVector, usize rightOffset, usize compDims, DistanceMetric distMetric)
```
mmarineBlueQuartz:
```suggestion
float64 GetDistance(const leftDataType& leftVector, usize leftOffset, const rightDataType& rightVector, usize rightOffset, usize compDims, DistanceMetric distMetric)
```

The return type should probably be explicitly defined so that knows the type returned at a glance instead of having to read the entire function. `dist` is defined in the first line, but if operations are performed on the line the function returns, the return value could be implicitly cast to another type and it wouldn't be obvious why the return value is different than expected. This does not happen in this function, but that is not obvious without reading the definition first.

mmarineBlueQuartz (approved):

PR #583, opened 2023-05-10
ENH: Update MoveData filter to allow multiple moves at once

jmarquisbq:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/complex/Filter/Actions/MoveDataAction.cpp
+++ b/src/complex/Filter/Actions/MoveDataAction.cpp
@@ -15,6 +15,7 @@ constexpr int32 k_MissingParentObjectCode = -6403;
 constexpr int32 k_FailedAddParentCode = -6404;
 constexpr int32 k_FailedRemoveParentCode = -6405;
 constexpr int32 k_FailedFindOldParentCode = -6406;
+constexpr int32 k_RecursiveParentCode = -6407;
 } // namespace
 
 namespace complex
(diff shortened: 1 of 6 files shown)


src/complex/Filter/Actions/MoveDataAction.cpp
```diff
@@ -54,6 +60,11 @@ Result<> MoveDataAction::apply(DataStructure& dataStructure, Mode mode) const
     return MakeErrorResult(k_MissingParentObjectCode, fmt::format("{}Could not find a possible parent object at '{}'", prefix, m_NewParentPath.toString()));
   }
 
+  if(const auto* dataGroup = dataStructure.getDataAs<BaseGroup>(m_Path); dataGroup != nullptr && dataGroup->isParentOf(newParent))
```
mmarineBlueQuartz: isParentOf only checks one degree of separation (parent-child). If we're going to check here, this should be a recursive check. One additional degree of separation (grandparent-grandchild) would pass this test.

I would also recommend moving the check to BaseGroup::canInsert so that all filters and actions that attempt to move or add parent objects get it built in.

mmarineBlueQuartz (changes requested):

src/complex/Filter/Actions/MoveDataAction.cpp
jmarquisbq: Well I checked trying to move a grandparent into one of its grandchildren and this check caught that. I think maybe because the DataObject::hasParent function is using that StringUtilities::contains function. I could write a slightly different recursive approach if we think that'll be more robust though.

mmarineBlueQuartz (approved):

PR #606, opened 2023-06-09
BUG FIX: ApplyTransformationToGeometry now applies precomputed transformations correctly.

joeykleingers:
When using the precomputed transformation matrix option on an image geometry in the ApplyTransformationToGeometry filter, the filter would crash because the precomputed transformation matrix cannot be read during preflight and so the transformed image geometry, cell attribute matrix, and cell data arrays were being created in preflight using the default identity matrix, which was causing these objects to have the same dimensions as the source objects.  But then in execute the objects were expected to be the correct sizes using the actual precomputed transformation matrix, causing an index-out-of-bounds error.

I updated the filter so that the destination geometry's dimensions, origin, and spacing and the cell attribute matrix's tuple shape are adjusted in execute right before calculating and copying values into the destination arrays.  I also updated the PreflightUpdatedValue to display "Precomputed transformation matrix unknown during preflight." when the precomputed transformation matrix option is chosen.


jmarquisbq (approved):
I just added a separate issue for updating the unit test so we can get this fix in for the rc3 release

imikejackson:
Good catch.

PR #613, opened 2023-06-16
BUG: Fixes to H5Support and H5.cpp

imikejackson:
H5.cpp: 

* Cleanup of clang-tidy warnings about "else" after return
* Properly use H5Tequal() API

H5Support.cpp
* move functions into the .cpp file since they are not templates
* Add StringForHDFType() function.

### DIFF
--- a/src/complex/Utilities/Parsing/HDF5/H5Support.hpp
+++ b/src/complex/Utilities/Parsing/HDF5/H5Support.hpp
@@ -79,7 +79,7 @@ namespace Support
  * @param objectName The name of the object to check
  * @return True if the given hdf5 object identifier is a group
  */
-inline bool COMPLEX_EXPORT IsGroup(hid_t nodeId, const std::string& objectName);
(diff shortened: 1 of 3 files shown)


src/complex/Utilities/Parsing/HDF5/H5Support.hpp
```diff
@@ -79,36 +79,14 @@ namespace Support
...
-    break;
-  case H5O_TYPE_NAMED_DATATYPE:
-    isGroup = false;
-    break;
-  default:
-    isGroup = false;
-  }
-  return isGroup;
-}
+inline bool COMPLEX_EXPORT IsGroup(hid_t nodeId, const std::string& objectName);
```
mmarineBlueQuartz:
```suggestion
bool COMPLEX_EXPORT IsGroup(hid_t nodeId, const std::string& objectName);
```
Inline or forward declare functions.

mmarineBlueQuartz (changes requested):

mmarineBlueQuartz (approved):

PR #624, opened 2023-06-27
Handle DataStructure::makePath errors

### LINKED ISSUE #127: DataStructure::makePath leaves it in an incomplete state on failure
`DataStructure::makePath` creates intermediate paths as it needs them which will persist on failure which may be unexpected. It also doesn't check if the result of some it's casts which can lead to segfaults/undefined behavior on failure.

mmarineBlueQuartz:
Removes created DataPaths if DataStructure::makePath causes an error.

Fixes #127

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/complex/DataStructure/DataStructure.cpp
+++ b/src/complex/DataStructure/DataStructure.cpp
@@ -159,6 +159,8 @@ bool DataStructure::containsData(const DataPath& path) const
 
 Result<LinkedPath> DataStructure::makePath(const DataPath& path)
 {
+  std::vector<DataObject::IdType> createdIds;
+
   try
   {
     std::vector<DataObject::IdType> pathIds;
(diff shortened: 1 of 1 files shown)


nyoungbq (approved):

src/complex/DataStructure/DataStructure.cpp
```diff
@@ -167,17 +169,24 @@ Result<LinkedPath> DataStructure::makePath(const DataPath& path)
...
+      createdIds.push_back(data->getId());
     }
     const BaseGroup* parent = dynamic_cast<const BaseGroup*>(data);
     pathIds.push_back(data->getId());
 
     for(usize i = 1; i < path.getLength(); i++)
     {
+      if(parent == nullptr)
+      {
+        return complex::MakeErrorResult<LinkedPath>(-3, "Target parent object in path is not derived from BaseGroup.");
```
imikejackson: Please incorporate the "path" argument into the error message. This will help the user/developer better diagnose what is going wrong.

src/complex/DataStructure/DataStructure.cpp
```diff
@@ -187,6 +196,11 @@ Result<LinkedPath> DataStructure::makePath(const DataPath& path)
     return {LinkedPath(this, pathIds)};
   } catch(const std::exception& e)
   {
+    for(const auto& id : createdIds)
+    {
+      removeData(id);
+    }
+
     return complex::MakeErrorResult<LinkedPath>(-2, "Exception thrown when attempting to create a path in the DataStructure");
```
imikejackson: Please also incorporate that actual value of 'path' into the error message. If there are ways to better determine the type of exception thrown that would also help in the error message.

imikejackson (changes requested):

src/complex/DataStructure/DataStructure.cpp
nyoungbq:
```suggestion
        return complex::MakeErrorResult<LinkedPath>(-3, fmt::format("Target parent object {} in path {} is not derived from BaseGroup.", name, path.toString()));
```

src/complex/DataStructure/DataStructure.cpp
nyoungbq:
```suggestion
    return complex::MakeErrorResult<LinkedPath>(-2, fmt::format("Exception thrown when attempting to create a path in the DataStructure: {}", e.what()));
```

nyoungbq (approved):

PR #630, opened 2023-06-28
BUG: FindArrayStats & RobustAutomaticThreshold

joeykleingers:
* FindArrayStatistics now uses uint64 for internal summation array calculations.  This was originally set to the input array's type, and it was causing the summations to not be calculated properly if the input array happened to have a boolean type.
* Robust Automatic Threshold now creates a mask array with proper tuple shape.

### DIFF
--- a/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/FindArrayStatistics.cpp
+++ b/src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/FindArrayStatistics.cpp
@@ -64,7 +64,7 @@ public:
     std::vector<uint64> length(numCurrentFeatures, 0);
     std::vector<T> min(numCurrentFeatures, std::numeric_limits<T>::max());
     std::vector<T> max(numCurrentFeatures, std::numeric_limits<T>::min());
-    std::vector<uint64> summation(numCurrentFeatures, 0);
+    std::vector<float32> summation(numCurrentFeatures, 0);
     usize progressCount = 0;
 
     usize progressIncrement = numTuples / 100;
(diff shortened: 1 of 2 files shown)


src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/FindArrayStatistics.cpp
```diff
@@ -64,7 +64,7 @@ class FindArrayStatisticsByIndexImpl
     std::vector<uint64> length(numCurrentFeatures, 0);
     std::vector<T> min(numCurrentFeatures, std::numeric_limits<T>::max());
     std::vector<T> max(numCurrentFeatures, std::numeric_limits<T>::min());
-    std::vector<T> summation(numCurrentFeatures, 0);
+    std::vector<uint64> summation(numCurrentFeatures, 0);
```
jmarquisbq: @joeykleingers Can I ask why you chose uint64? I think everywhere else that we are dealing with a summation we use float32 (and the summation data array itself is a float32 array) so would it be better to use a float32 here instead since we are casting it to float32 below anyway?

joeykleingers: I was following what the length data array does.  It is also float32 but the array used to calculate it is uint64.  We could probably change both.

jmarquisbq (approved):
I assume we are just halting all merges until the update @imikejackson is working on for the unit test packages goes in but I'm going to go ahead and approve this so you can get it merged as soon as the tests are passing again.

PR #637, opened 2023-06-30
BUG: Fix issue with assumed A: drive being available on CI.

imikejackson:


src/Plugins/ComplexCore/test/WriteBinaryDataTest.cpp
```diff
@@ -30,6 +39,53 @@ constexpr uint64 k_MultipleFiles = 0;     // enum representation
...
+  for(char letter = 'A'; letter < 'Z'; letter++)
+  {
+    const UINT driveType = GetDriveTypeA(&letter);
+    ret.push_back(driveType);
+    std::cout << letter << "   " << driveType << std::endl;
+  }
+  return ret;
+}
+
+static inline bool isDriveReady(const wchar_t* path)
```
JDuffeyBQ:
```suggestion
bool isDriveReady(const wchar_t* path)
```

src/Plugins/ComplexCore/test/WriteBinaryDataTest.cpp
```diff
@@ -30,6 +39,53 @@ constexpr uint64 k_MultipleFiles = 0;     // enum representation
 constexpr uint64 k_SingleFile = 1;        // enum representation
 } // namespace
 
+#if _WIN32
+
+std::vector<UINT> CheckAllDrives()
```
JDuffeyBQ: These functions should be moved into the above anonymous namespace.

JDuffeyBQ (changes requested):

PR #643, opened 2023-07-05
BUG: Catch nullptr trying to rename DataObject

imikejackson:
Fix an unchecked pointer access in DataObject::canRename() in the lambda

### DIFF
--- a/src/complex/DataStructure/DataObject.cpp
+++ b/src/complex/DataStructure/DataObject.cpp
@@ -164,14 +164,15 @@ std::string DataObject::getName() const
 
 bool DataObject::canRename(const std::string& name) const
 {
-  if(!IsValidName(name))
-  {
-    return false;
-  }
-
+  
   if(name == getName())
   {
     return true;
+  }
+
+  if(!IsValidName(name))
+  {
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


src/complex/DataStructure/DataObject.cpp
```diff
@@ -169,16 +169,34 @@ bool DataObject::canRename(const std::string& name) const
...
+  });
 }
 
 bool DataObject::rename(const std::string& name)
 {
+  if(name == m_Name)
+  {
+    return true;
+  }
+
```
jmarquisbq, line 199:
```suggestion
```

This check feels a bit redundant since we check the exact same thing in the canRename function which gets called immediately after.

jmarquisbq (approved):

src/complex/DataStructure/DataObject.cpp
imikejackson: I sort of agree. Let me take another look at it.

jmarquisbq: It's slightly different since it would mean re-assigning the name to the same thing below the canRename if statement. Not sure it matters too much one way or the other.

PR #644, opened 2023-07-06
ENH: Add option in ApplyTransformationToGeometry to translate geometry to global origin before transformation.

joeykleingers:


joeykleingers:
Originally, translating the geometry to the global origin was automatically done for only the rotation and scaling modes, but now it's an option in the UI that the user can turn on for any transformation mode.

jmarquisbq (approved):

PR #669, opened 2023-08-18
ENH:MultiThresholdObjects: Add ability to set custom TRUE and FALSE values.

joeykleingers:


src/Plugins/ComplexCore/src/ComplexCore/Filters/MultiThresholdObjects.cpp
```diff
@@ -438,6 +483,26 @@ IFilter::PreflightResult MultiThresholdObjects::preflightImpl(const DataStructur
     }
   }
 
+  if(useCustomTrueValue)
```
JDuffeyBQ, line 498: We should have a check to prevent selecting custom true/false values if the output mask array uses bools.

src/Plugins/ComplexCore/test/MultiThresholdObjectsTest.cpp
```diff
@@ -86,19 +113,19 @@ TEST_CASE("ComplexCore::MultiThresholdObjects: Valid Execution", "[ComplexCore][
     auto executeResult = filter.execute(dataStructure, args);
     COMPLEX_RESULT_REQUIRE_VALID(executeResult.result)
 
-    auto* thresholdArray = dataStructure.getDataAs<BoolArray>(k_ThresholdArrayPath);
+    auto* thresholdArray = dataStructure.getDataAs<UInt8Array>(k_ThresholdArrayPath);
```
JDuffeyBQ: I think that we should keep a bool mask array in addition to the custom values test to cover all options.

JDuffeyBQ (changes requested):

src/Plugins/ComplexCore/test/MultiThresholdObjectsTest.cpp
imikejackson: I agree.

imikejackson:
The documentation MUST be updated to include something to the effect that if you chose custom values for true/false this will break other filters that rely on a 0/1 for true/false.

@joeykleingers Please make sure the documentation is updated.

JDuffeyBQ (approved):

PR #670, opened 2023-08-18
Tweak filter human labels to remove version number string.

joeykleingers:


mmarineBlueQuartz (approved):

mmarineBlueQuartz (changes requested):
Change DREAM3DFileTest.cpp in ComplexCoreUnitTest as well. It uses a human label string comparison using the (V8) label.

`constexpr StringLiteral k_ExportD3DFilterName = "Write DREAM3D NX File (V8)";`

mmarineBlueQuartz (approved):

PR #679, opened 2023-08-28
ENH: Calculate "Most Populated Bin" Array In Find Attribute Array Statistics

joeykleingers:
+ Calculates the "Most Populated Bin" array when "Find Histogram" is ON.  The "Most Populated Bin" array is a 2-component array that stores the index and value count of the most populated bin in the first and second components, respectively.  This array is also calculated per feature when "Compute Statistics Per Feature/Ensemble" is ON.

+ Fixed a crash that was occurring when calculating a histogram per feature.  The edge case where a feature happens to have identical values was not being handled.  The value count for the feature is stored in the first component of the histogram and the rest of the components are 0s.

+ Added a missing serial execution option to FindArrayStatistics... previously if multithreaded was turned off then the filter would just do nothing.

+ Fixed a compile error that was occurring when multithreading was turned off.

+ Required the number of histogram bins to be at least 1, because 0 or negative bins makes no sense.

+ Updated the histogram array to output uint64 values instead of floats, because the output values should be value counts for each bin.  Therefore, floats don't make sense.


joeykleingers:
@mmarineBlueQuartz I'm intending you as the primary reviewer, but I also listed @imikejackson in case he wants to also take a look at it.

mmarineBlueQuartz (approved):

imikejackson (approved):

imikejackson (approved):

PR #696, opened 2023-09-21
[WIP] DEBUG: Changes to allow easier debugging of the FindNeighborHoods filter

imikejackson:
@nyoungbq  This is a validation branch only. If you use the pipelines we were working with you can run the FindNeighborHoods filter in both parallel and serial in the same pipeline, then use the output from the pipeline to run through the commented out unit test. It validated on macOS (x64).


imikejackson:
This was just for debugging. Closing.

PR #702, opened 2023-09-26
Remove dependency on EbsdLib for the Parameter.

imikejackson:
This change removes a dependency on EbsdLib for complex when OrientationAnalysis is included. The code removed should really be down in the filter instead where the specific filter using this parameter knows best what it is looking for.


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/ImportH5Data.hpp
```diff
@@ -20,6 +20,8 @@ struct ORIENTATIONANALYSIS_EXPORT ImportH5DataInputValues
   DataPath ImageGeometryPath;
   DataPath CellAttributeMatrixPath;
   DataPath CellEnsembleAttributeMatrixPath;
+  bool EdaxHexagonalAlignment;
+  bool ConvertPhaseToInt32;
```
JDuffeyBQ, line 24: Do these get used?

imikejackson: yes

imikejackson:
Refactored the code as requested.

JDuffeyBQ (approved):

PR #706, opened 2023-09-29
ENH: Read CSV File Redesign

joeykleingers:
* Renamed the filter from ImportCSVData to ReadCSVFile, and also renamed all parameters, widgets, utilities, and helper classes.
* Updated the parameter to only include the data required for the filter, removing unnecessary fields and make sure that all fields can be converted easily with the Python bindings.
* Updated the filter to do additional error checking and cache the file's total lines and headers to avoid reading the entire file every preflight.
* Moved the DynamicTableData tuple dimensions parameter into the CSVImporterData custom parameter.
* Removed all references to "wizard", since the UI no longer involves using a wizard.
* Updated the Python bindings to expose the new ReadCSVData struct to Python.
* Updated the Read CSV File unit test and any other unit tests that use the filter.
* Updated the regular filter documentation as well as the Python documentation.
* Renamed ImportTextFilter to ReadTextDataArrayFilter, to avoid naming confusion.


src/complex/Utilities/FileUtilities.hpp
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  // Allocate the buffer
+  std::vector<char> buffer(actualSize, 0);
+
+  // Copy the file contents into the buffer
+  try
+  {
+    in.read(buffer.data(), actualSize);
+  } catch(std::exception& e)
```
JDuffeyBQ:
```suggestion
  } catch(const std::exception& e)
```

src/complex/Utilities/FileUtilities.hpp
```diff
@@ -0,0 +1,128 @@
...
+  if(!hasNewLines && !hasCarriageReturns && !hasTabs)
+  {
+    // This might be a binary file
+    return MakeErrorResult(-304, fmt::format("The file \"{}\" might be a binary file, because line-feed, tab, or carriage return characters have not been detected. Using this file may crash the "
+                                             "program or cause unexpected results.  Please import a different file.",
+                                             absPath.string()));
+  }
+
+  // Close the file and free the memory from the buffer
+  in.close();
```
JDuffeyBQ:
```suggestion
```
Not need because it's already handled by the `std::fstream` destructor.

src/complex/Utilities/FileUtilities.hpp
```diff
@@ -0,0 +1,128 @@
...
+  if(!fs::exists({absPath}))
+  {
+    return MakeErrorResult(-300, fmt::format("File does not exist: {}", absPath.string()));
+  }
+
+  // Obtain the file size
+  const size_t fileSize = fs::file_size(absPath);
+
+  // Open the file
+  std::fstream in(absPath.c_str(), std::ios_base::in | std::ios_base::binary);
```
JDuffeyBQ:
```suggestion
  std::ifstream in(absPath.c_str(), std::ios_base::binary);
```

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportCSVDataFilter.cpp
```diff
@@ -433,58 +577,88 @@ IFilter::PreflightResult ImportCSVDataFilter::preflightImpl(const DataStructure&
...
+  }
+  if(csvImporterData.tabAsDelimiter)
+  {
+    delimiters.push_back('\t');
+  }
 
   float32 threshold = 0.0f;
-  usize numTuples = numLines - beginIndex + 1;
-  for(usize lineNum = beginIndex; lineNum <= numLines; lineNum++)
+  usize numTuples = std::accumulate(csvImporterData.tupleDims.begin(), csvImporterData.tupleDims.end(), 1, std::multiplies<>());
```
JDuffeyBQ:
```suggestion
  usize numTuples = std::accumulate(csvImporterData.tupleDims.cbegin(), csvImporterData.tupleDims.cend(), static_cast<usize>(1), std::multiplies<>());
```
Unfortunately there's a warning that keeps cropping up where `std::accumulate` and other standard library functions deduce the wrong return type if you don't static cast the initial value.

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportCSVDataFilter.cpp
```diff
@@ -433,58 +577,88 @@ IFilter::PreflightResult ImportCSVDataFilter::preflightImpl(const DataStructure&
...
+  if(useExistingGroup)
+  {
+    const AttributeMatrix& am = dataStructure.getDataRefAs<AttributeMatrix>(groupPath);
+    numTuples = std::accumulate(am.getShape().begin(), am.getShape().end(), 1, std::multiplies<>());
```
JDuffeyBQ:
```suggestion
    numTuples = std::accumulate(am.getShape().cbegin(), am.getShape().cend(), static_cast<usize>(1), std::multiplies<>());
```

src/Plugins/ComplexCore/src/ComplexCore/Filters/ImportCSVDataFilter.cpp
```diff
@@ -343,88 +362,213 @@ IFilter::UniquePointer ImportCSVDataFilter::clone() const
   return std::make_unique<ImportCSVDataFilter>();
 }
 
+//------------------------------------------------------------------------------
+IFilter::PreflightResult ImportCSVDataFilter::readHeaders(const std::string& inputFilePath, usize headersLineNum, bool useTab, bool useSemicolon, bool useSpace, bool useComma,
```
JDuffeyBQ: This function could take the header cache as an argument instead of being a member function.

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

JDuffeyBQ:
It's probably a good idea to add a unit test for the blank line handling.

joeykleingers:
@JDuffeyBQ Blank lines test has been added.  Approve unless there's anything else you'd like fixed/changed.

joeykleingers:
Actually, hang on... let me add one more test that checks to make sure that the filter handles unexpected inputs for the other arguments.  Like if the user inputs 0 for the start row or a number larger than the total file size for the header row, for example.

joeykleingers:
@JDuffeyBQ Last unit test has been added.  Ready for review again.

JDuffeyBQ:
I don't have a good suggestion but we have a similarly named filter in ImportTextFilter so I think that they should be easier to distinguish.

imikejackson:
These need renamed:

[ ] ImportTextFilter -> ReadTextDataArrayFilter
[ ] ImportTextDataFilter -> ReadCSVFileFilter

[ ] Rename the parameter from `ImportTextDataParameter` to `ReadCSVFileParameter`

[ ] TextImporterData::writeJson convert the delimiters to a std::vector<std::string> BEFORE writing
[ ] TextImporterData::readJson convert from std::vector<std::string> to std::vector<char> after reading in from the JSON file.

joeykleingers:
@imikejackson All of the most recent suggestions have been implemented.  This is ready for final review and approval.

imikejackson (approved):

PR #747, opened 2023-10-19
ENH: GenerateColorTable allows use of a mask array.

### LINKED ISSUE #530: Generate Color Table: Add "mask" parameter and default color for mask value
Add the ability to also use a mask to apply a specific RGB value where the mask is true or false, i.e., the inverse of the mask.

imikejackson:
Mask values that have a value of FALSE will get the "Invalid Color" applied.


src/Plugins/ComplexCore/src/ComplexCore/Filters/Algorithms/GenerateColorTable.cpp
```diff
@@ -126,14 +163,20 @@ class GenerateColorTableImpl
   int m_NumControlColors;
   const std::vector<std::vector<float64>>& m_ControlPoints;
   UInt8Array& m_ColorArray;
+  const complex::IDataArray* m_GoodVoxels = nullptr;
+  std::vector<uint8> m_InvalidColor;
```
nyoungbq:
```suggestion
  const std::vector<uint8>& m_InvalidColor;
```

nyoungbq (approved):

PR #751, opened 2023-10-24
PY: Added cx.DataPath constructor to take a "/Path/Like/This" string

imikejackson:


src/Plugins/ComplexCore/wrapping/python/complexpy.cpp
```diff
@@ -256,6 +256,12 @@ Result<> ExecutePipeline(Pipeline& pipeline, DataStructure& dataStructure)
   return result;
 }
 
+complex::DataPath CreateDataPath(const std::string& path)
```
JDuffeyBQ:
```suggestion
complex::DataPath CreateDataPath(std::string_view path)
```

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

JDuffeyBQ (approved):

PR #754, opened 2023-10-25
ENH: Add IndexToDataType Method

joeykleingers:


src/complex/Common/TypesUtility.hpp
```diff
@@ -374,6 +374,63 @@ inline constexpr DataType StringToDataType(std::string_view dataTypeString)
...
+  {
+    return DataType::float64;
+  }
+  else if(dataTypeString == DataTypeToHumanString(DataType::boolean).view())
+  {
+    return DataType::boolean;
+  }
+  else
+  {
+    throw std::runtime_error("complex::HumanStringToDataType: No known DataType matches the given string value.");
```
imikejackson: In the error message, please also include the value of the 'dataTypeString' variable:

.... matches the given string value: '{}'", dataTypeString);

if you were to use fmt::format

imikejackson (changes requested):

src/complex/Common/TypesUtility.hpp
```diff
@@ -374,6 +374,63 @@ inline constexpr DataType StringToDataType(std::string_view dataTypeString)
   }
 }
 
+/**
+ * @brief Returns a DataType for the passed in string representation
+ * @param dataTypeString
+ * @return
+ */
+inline constexpr DataType HumanStringToDataType(std::string_view dataTypeString)
```
imikejackson: You could consider using an std::optional or nonstd::expected type rather than the exception? If we use the exception I think we are supposed to label the method as "throws" (Look up the spec or ping Jared for the specifics).

imikejackson (changes requested):

imikejackson (approved):

PR #768, opened 2023-11-15
BUG: Fix validation logic in ValidateNumFeaturesInArray

imikejackson:
As long as the max 'featureId' value is Less than the total number of tuples in the source attribute matrix, this is fine. That just means that there are unused features which is odd but not critical.


nyoungbq (approved):
Looks good, pretty sure I was the one that put this extra validation in. Read your description and it makes sense on why it's unnecessary now.

PR #769, opened 2023-11-15
FILTER: Initialize Data (Initialize Data Rework)

nyoungbq:
PRELIMINARY PR
So far I have:
- refactored the old Initialize Data
- created options for initializing by
  - value
  - incremental processing
  - random
  - ranged random
- Added boolean functionality
- Allowed for single and multi-component initialization for multi component arrays
- ways to control how the randomization is gated and seeded
- cleaned up all the user feedback
- verified functionality of all components
- got it cross compiling

Todo Checklist:
- [x] Unit Test
- [x] Documentation Updates
 

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/Plugins/ComplexCore/src/ComplexCore/Filters/InitializeImageGeomCellData.cpp
```diff
@@ -0,0 +1,369 @@
...
+  {
+    std::uniform_real_distribution<T> distribution(rangeMin, rangeMax);
+    return std::make_pair(distribution, generator);
+  }
+}
+
+struct InitializeArrayFunctor
+{
+  template <class T>
+  void operator()(IDataArray& dataArray, const std::array<usize, 3>& dims, uint64 xMin, uint64 xMax, uint64 yMin, uint64 yMax, uint64 zMin, uint64 zMax, InitializeImageGeomCellData::InitType initType,
```
mmarineBlueQuartz, line 107: Since this is templated, can you use DataArray<T>? That also slightly simplifies line 123 so you can use dataArray.getDataStoreRef() to receive the AbstractDataStore<T> reference.

src/complex/Utilities/DataArrayUtilities.hpp
```diff
@@ -153,7 +153,7 @@ struct ConvertTo<bool>
       return {floatResult.value() != 0.0};
     }
 
-    return {false};
+    return {true};
```
mmarineBlueQuartz, line 156: What was the reason this was changed, and is this used somewhere that it could break assumptions should the input not have a clear conversion?

src/Plugins/ComplexCore/test/InitializeImageGeomCellDataTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, 0, xMin - 1, 0, yMin - 1, 0, zMin - 1, 0.0));
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, xMin, xMax, yMin, yMax, zMin, zMax, initValue));
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, xMax + 1, k_ImageDims[0] - 1, yMax + 1, k_ImageDims[1] - 1, zMax + 1, k_ImageDims[2] - 1, 0.0));
+      break;
+    }
+    case DataType::float64: {
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, 0, xMin - 1, 0, yMin - 1, 0, zMin - 1, 0.0));
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, xMin, xMax, yMin, yMax, zMin, zMax, initValue));
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, xMax + 1, k_ImageDims[0] - 1, yMax + 1, k_ImageDims[1] - 1, zMax + 1, k_ImageDims[2] - 1, 0.0));
+      break;
```
mmarineBlueQuartz: This looks like a lot of copy-pasted code with identical parameters. Can each case be simplified to a single templated method for each type that calls the same checks? I don't think test values will need to be updated in the future, but I would not be happy with a code block like this if they were.

For example:
> case: DataType:: int8:
>    RangeCheck<int8>(dataStore, xMin, yMin, zMin, xMax, yMax, zMax, k_ImageDims, 0.0);
>    break;

src/Plugins/ComplexCore/test/InitializeImageGeomCellDataTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+    }
+    case DataType::float32: {
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, 0, xMin - 1, 0, yMin - 1, 0, zMin - 1, 0.0));
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, xMax + 1, k_ImageDims[0] - 1, yMax + 1, k_ImageDims[1] - 1, zMax + 1, k_ImageDims[2] - 1, 0.0));
+      break;
+    }
+    case DataType::float64: {
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, 0, xMin - 1, 0, yMin - 1, 0, zMin - 1, 0.0));
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, xMax + 1, k_ImageDims[0] - 1, yMax + 1, k_ImageDims[1] - 1, zMax + 1, k_ImageDims[2] - 1, 0.0));
+      break;
```
mmarineBlueQuartz: Extract copy-pasted code to its own function unless these arguments really might change based on the data type.

src/Plugins/ComplexCore/test/InitializeImageGeomCellDataTest.cpp
```diff
@@ -0,0 +1,391 @@
...
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, 0, xMin - 1, 0, yMin - 1, 0, zMin - 1, 0.0));
+      REQUIRE(IsDataWithinInclusiveRange<float32>(dataStore, xMin, xMax, yMin, yMax, zMin, zMax, initRange));
+      REQUIRE(DoesRangeEqualValue<float32>(dataStore, xMax + 1, k_ImageDims[0] - 1, yMax + 1, k_ImageDims[1] - 1, zMax + 1, k_ImageDims[2] - 1, 0.0));
+      break;
+    }
+    case DataType::float64: {
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, 0, xMin - 1, 0, yMin - 1, 0, zMin - 1, 0.0));
+      REQUIRE(IsDataWithinInclusiveRange<float64>(dataStore, xMin, xMax, yMin, yMax, zMin, zMax, initRange));
+      REQUIRE(DoesRangeEqualValue<float64>(dataStore, xMax + 1, k_ImageDims[0] - 1, yMax + 1, k_ImageDims[1] - 1, zMax + 1, k_ImageDims[2] - 1, 0.0));
+      break;
```
mmarineBlueQuartz: Same as above.

mmarineBlueQuartz (changes requested):

src/complex/Utilities/DataArrayUtilities.hpp
nyoungbq: > What was the reason this was changed, and is this used somewhere that it could break assumptions should the input not have a clear conversion?

I had hopped on a call with Mike and he told me that he wants boolean values to always return true if they are not explicitly false and he approved the change in advance so long as all the affected classes tests still passed.

mmarineBlueQuartz: Alright. Sounds good to me.

src/Plugins/ComplexCore/src/ComplexCore/Filters/InitializeImageGeomCellData.cpp
nyoungbq: > Since this is templated, can you use DataArray? That also slightly simplifies line 123 so you can use dataArray.getDataStoreRef() to receive the AbstractDataStore reference.

This was the original Initialize Data filter that was renamed so I didn't really work on this one, hence the extensive repeated use of switches in the test for it, but I don't think this is possible with the API call to the ExecuteNeighborFunction. That being said I can turn this functor into a class and capture the typed DataArray as a member variable from the constructor via the RunTemplateClass API function if that is preferable.

nyoungbq: This is not possible without a switch to grab the array from the dataStrucutre or passing the whole datastructure by ref to the functor

mmarineBlueQuartz (approved):

PR #784, opened 2023-12-01
ITK: Update ITK to version 5.3.0

imikejackson:
Update ITK to version 5.3.0 through complex-registry and small changes in the source codes.

This will setup the ability to package DREAM3D-NX with Anaconda


imikejackson:
BEFORE THIS IS MERGED: The vcpkg-configure needs to be updated to point back to the main COMPLEX-REGISTRY repository. 

For now, this will test if everything works on all platforms.

vcpkg-configuration.json
```diff
@@ -7,7 +7,7 @@
   "registries": [
     {
       "kind": "git",
-      "repository": "https://github.com/bluequartzsoftware/complex-registry",
+      "repository": "https://github.com/imikejackson/complex-registry",
```
nyoungbq: Is this supposed to be set to your fork?

nyoungbq: I just saw your comment

nyoungbq (approved):
All looks good just make sure to point it back to bluequartz registry once its updated

vcpkg-configuration.json
imikejackson: At the moment yes. Before I commit the update to the complex-registry I want to make sure this works on all the platforms.

PR #789, opened 2023-12-05
ENH: Allow test data to be downloaded from a different URL.

imikejackson:
This is being put in place to allow internal build bots to utilize the local web server to download the data files.


nyoungbq (approved):
looks good to me, should be able speed up build bots with this

PR #801, opened 2023-12-20
ENH: Rename complex to simplnx

imikejackson:
The `complex` library is renamed to "SIMPLNX"


joeykleingers (approved):

jmarquisbq (approved):

joeykleingers (approved):

JDuffeyBQ (approved):

README.md
nyoungbq, line 1: Not an actual problem just a style warning:
Just a heads up on this file the all-contributors bot will post links with the uppercase `BlueQuartzSoftware`, since that is the way the URL is distributed. That being said I checked the github docs and organization names are case insensitive so it shouldn't matter either way (and URLS are obviously case insensitive when making requests outside the local OS filesystem). All this is to say that it isn't breaking and lowercase URLs conforms to typical conventions, but I fear uniformity may be broken when we add new contributors, since this is a niche styling decision that will likely not be considered/remembered when reviewing the new contributors PR.

nyoungbq (approved):

mmarineBlueQuartz (approved):

README.md
imikejackson: I'll get this cleaned up on the back side of this conversion.

PR #816, opened 2024-01-11
DOC: ConvertHexGridToSquareGrid. Update and add image to doc file

imikejackson:


nyoungbq (approved):

nyoungbq:
Might be worth adding the "ANG File, EDAX, Hex Grid" tags to the default tags as part of this PR too, but not a requirement. I didn't update the default tags in the original PR it only says "Conversion" and "Processing" currently (and the class name).

PR #818, opened 2024-01-11
ENH: Crop Image Geometry with Physical Coordinate Bounds

### LINKED ISSUE #794: CropImageGeometry: Allow user to set the crop region using lengh units instead of voxels
The current filter requires the user to crop using voxel index coordinates instead of physical unit coordinates. Adding an option to the filter to allow the user to state the min/max in physical units would help some use cases.

The internal implementation should just be simple and straight forward when doing the crop. Find the voxel that the min coordinate is located in and find the voxel that the max coordinate is located in and then just use those voxel coordinates.

In this version we should not worry about partial cell cropping.

nyoungbq:
Fixes #794

- Added ability to define crop with coordinates
- Added test case
- Updated documentation

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/Plugins/SimplnxCore/docs/CropImageGeometry.md
```diff
@@ -15,6 +15,32 @@ Bounds: 100 voxels * 0.25 micron/voxel = 25 microns
 
 *Note:* The input parameters are *inclusive* and begin at *0*, so in the above example *0-99* covers the entire range of **Cells** in a given dimension.
 
+*Note on Physical Bounds:* If the physical coordinate bounds are outside the image geometry's bounds the nearest valid bound will be used. Example:
```
imikejackson: I think we are going to need some more explicit documentation with images to fully explain what is going on.

imikejackson (changes requested):

src/Plugins/SimplnxCore/docs/CropImageGeometry.md
imikejackson: Documentation has been updated.

imikejackson (approved):

imikejackson (approved):

PR #838, opened 2024-01-30
FILTERS: Added CalculateHistogram, InterpolateGridData, and CliReader Python filters.

joeykleingers:
+ Adds the 3 filters mentioned in the title to a new plugin called DataAnalysisToolkit.
+ Exposed StringArray to Python so that it can be used in CliReader Python filter.
+ Fixes a bug where the IDataCreationAction subclasses were not correctly passing the geometry path member variable when being cloned.


imikejackson (approved):

imikejackson (changes requested):
Going to need a doc file for these filters.

imikejackson (approved):

PR #839, opened 2024-01-30
BUG: ReadStlFile-Fix error logic when reading an ASCII STL File.

imikejackson:
Add an experimental function to convert from ASCI file to a binary STL file. This could be used in the actual filter to add support for reading ASCII STL files.


nyoungbq (approved):
Looks good for the most part. The enum was a good choice and the new algorithm for conversion looks alright. That said, I think it should probably get a test case, but I understand if this is more of a utility function and it doesn't use stuff outside standard c++ so it might be okay to worry about testing later

PR #856, opened 2024-02-12
ENH: WriteAsciData now writes String Arrays

imikejackson:


nyoungbq (approved):
code looks good, and document looks great

PR #864, opened 2024-02-19
[WIP] BUG: Rework API for DataMap and DeepCopy in DataObject.

imikejackson:
We need to be able to report errors back up the call tree if possible.

### DIFF
--- a/src/simplnx/DataStructure/AttributeMatrix.cpp
+++ b/src/simplnx/DataStructure/AttributeMatrix.cpp
@@ -96,7 +96,10 @@ bool AttributeMatrix::canInsert(const DataObject* obj) const
 
   const usize totalTuples = std::accumulate(m_TupleShape.cbegin(), m_TupleShape.cend(), static_cast<usize>(1), std::multiplies<>());
   const usize incomingTupleCount = std::accumulate(arrayTupleShape.cbegin(), arrayTupleShape.cend(), static_cast<usize>(1), std::multiplies<>());
-
+  if(totalTuples != incomingTupleCount)
+  {
+    std::cout << "AttributeMatrix: CanInsert() Failed with object " << obj->getName() << ". totalTuples=" << totalTuples << "  incomingTupleCount=" << incomingTupleCount << "\n";
+  }
   return (totalTuples == incomingTupleCount);
 }
 
--- a/src/simplnx/DataStructure/AttributeMatrix.cpp
+++ b/src/simplnx/DataStructure/AttributeMatrix.cpp
@@ -78,18 +78,19 @@ std::string AttributeMatrix::getTypeName() const
   return k_TypeName;
 }
 
-bool AttributeMatrix::canInsert(const DataObject* obj) const
+Result<bool> AttributeMatrix::canInsert(const DataObject* obj) const
 {
-  if(!BaseGroup::canInsert(obj))
+  auto result = BaseGroup::canInsert(obj);
+  if(result.invalid())
   {
-    return false;
+    return result;
   }
 
   const auto* arrayObjectPtr = dynamic_cast<const IArray*>(obj);
 
   if(arrayObjectPtr == nullptr)
   {
-    return false;
+    return MakeErrorResult<bool>(-1673, "BaseGroup::canInsert() Error: DataObject being inserted is null");
   }
 
   const IArray::ShapeType arrayTupleShape = arrayObjectPtr->getTupleShape();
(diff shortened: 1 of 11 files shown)


src/simplnx/DataStructure/DataMap.cpp
```diff
@@ -30,6 +30,10 @@ DataMap DataMap::deepCopy(const DataPath& parentCopyPath) const
   {
     const auto& dataObj = m_Map.at(key);
     const auto copy = dataObj->deepCopy(parentCopyPath.createChildPath(dataObj->getName()));
+    if(copy == nullptr)
+    {
+      throw std::runtime_error(fmt::format("A deep copy request for parent '{}' failed on child object '{}' ", parentCopyPath.toString(), dataObj->getName()));
```
nyoungbq:
```suggestion
      throw std::runtime_error(fmt::format("{}({}): Function {}: A deep copy request for parent '{}' failed on child object '{}'", __FILE__, __LINE__, "DataMap::DeepCopy", parentCopyPath.toString(), dataObj->getName()));
```

nyoungbq (approved):

imikejackson:
@JDuffeyBQ @mmarineBlueQuartz @nyoungbq Did we want to actually do something more with this as in changing the API? Otherwise I'll rename the PR to something like "Updating error message"

nyoungbq:
@imikejackson I'm gonna be honest, it's been a while, so I don't really remember exactly what API this PR was aiming to fix in the first place. However, if the point is percolating errors up the call tree as the description suggests then I think it's a good idea to go further with this, since it will be easier to debug and more stable.

imikejackson:
I think there was some early discussion on slack about this. But, yes, I think the issue is that the error should be percolated up the call chain instead of just crashing at runtime. Either that or *every* call that uses this API should be put inside of a try-catch in order to not crash.

src/simplnx/DataStructure/BaseGroup.cpp
```diff
@@ -133,19 +135,21 @@ bool BaseGroup::isParentOf(const DataObject* dataObj) const
...
+  if(result.invalid())
   {
-    return false;
+    return result;
   }
   if(m_DataMap.insert(ptr))
   {
     ptr->addParent(this);
-    return true;
+    return {true};
```
nyoungbq:
```suggestion
    return {{true}};
```
Since we use std::expected (from extraneous nonstd library) I believe we need the second bracket to construct it inside the Result object to pass it to the default constructor. This ensures that the valid() and invalid() Result wrapper methods function correctly. @JDuffeyBQ should probably verify I am correct in this behavior since I'm shaky on how the result is intended to function.

src/simplnx/DataStructure/AttributeMatrix.cpp
```diff
@@ -78,26 +78,30 @@ std::string AttributeMatrix::getTypeName() const
...
 
   const usize totalTuples = std::accumulate(m_TupleShape.cbegin(), m_TupleShape.cend(), static_cast<usize>(1), std::multiplies<>());
   const usize incomingTupleCount = std::accumulate(arrayTupleShape.cbegin(), arrayTupleShape.cend(), static_cast<usize>(1), std::multiplies<>());
-
-  return (totalTuples == incomingTupleCount);
+  if(totalTuples != incomingTupleCount)
+  {
+    return MakeErrorResult<bool>(-1674, fmt::format("AttributeMatrix: CanInsert() Failed with object {}. totalTuples={} incomingTupleCount={}", obj->getName(), totalTuples, incomingTupleCount));
+  }
+  return {true};
```
nyoungbq:
```suggestion
  return {{true}};
```
Same justification

src/simplnx/DataStructure/BaseGroup.cpp
```diff
@@ -94,21 +95,22 @@ const DataObject& BaseGroup::at(const std::string& name) const
...
+    return MakeErrorResult<bool>(-1664, fmt::format("BaseGroup::canInsert() Error: DataObject with name='{}' and type='{}' already exists in the DataMap", obj->getName(), obj->getTypeName()));
   }
   if(const auto* objGroup = dynamic_cast<const BaseGroup*>(obj); objGroup != nullptr && objGroup->isParentOf(this))
   {
-    return false;
+    return MakeErrorResult<bool>(-1665, fmt::format("BaseGroup::canInsert() Error: DataObject with name='{}' and type='{}' is a parent of the current DataObject. A circular reference would occur.",
+                                                    obj->getName(), obj->getTypeName()));
   }
-  return true;
+  return {true};
```
nyoungbq:
```suggestion
  return {{true}};
```
Same justification

src/simplnx/DataStructure/BaseGroup.cpp
JDuffeyBQ: I think that issue was with a particular version of gcc. Testing on my Windows machine both ways compile and produce the same result.

Also in this particular case `Result<bool>` is redundant since the result already conveys the success or failure so it should just be `Result<void>`.

imikejackson: There is a lot of places that this is going to hit. I'm slowly going down this rabbit hole a bit at a time.

@JDuffeyBQ Ironically I started off with a Result<> but then changed to Result<bool> thinking that is what was needed. But you are correct in the fact that you return a "MakeErrorResult" will set the result to invalid, so that works.

PR #877, opened 2024-03-04
ENH: GeneratePythonFilter-Crashing fixed, Generation Fixed, headers split

imikejackson:
- Crash when adding a filter to an existing plugin
- Generation fixed when adding a filter to an existing plugin
- Split the constant strings from the header into their own header file which allows easier debugging of the sources


joeykleingers (approved):
Not sure if we care or not, but right now it is possible for someone to generate the same filter name twice into an existing plugin or new plugin.  I am approving this if we don't care, otherwise if we do, I can review and approve again when it's fixed.

imikejackson:
I also noticed the same thing. Since this solves the initial crashing issues we can merge this and put in another issue for the multiple same filter name issue.

PR #879, opened 2024-03-06
BUG: Fix resetting of the DataStructure::Id when importing from a file.

imikejackson:


src/simplnx/DataStructure/Geometry/INodeGeometry1D.cpp
```diff
@@ -224,37 +226,54 @@ void INodeGeometry1D::checkUpdatedIdsImpl(const std::vector<std::pair<IdType, Id
...
+      visitedMap[m_EdgeAttributeMatrixId] = true;
     }
 
-    if(m_ElementSizesId == updatedId.first)
+    if(m_ElementSizesId == updatedId.first && !visitedMap[m_EdgeAttributeMatrixId])
     {
       m_ElementSizesId = updatedId.second;
+      visitedMap[m_EdgeAttributeMatrixId] = true;
     }
+#endif
```
mmarineBlueQuartz: Might as well delete these lines if they are no longer relevant.

mmarineBlueQuartz (approved):
There is a code segment in INodeGeometry1D that should probably be deleted since it was disabled, but otherwise this looks ready to merge

src/simplnx/DataStructure/Geometry/INodeGeometry1D.cpp
imikejackson: Good catch. Thanks.

PR #891, opened 2024-03-19
BUG: Return invalid Result<> if size of directory is requested.

imikejackson:


src/simplnx/Utilities/FileUtilities.cpp
```diff
@@ -67,6 +67,10 @@ Result<> ValidateCSVFile(const std::string& filePath)
   {
     return MakeErrorResult(-300, fmt::format("File does not exist: {}", absPath.string()));
   }
+  if(fs::is_directory({absPath}))
+  {
+    return MakeErrorResult(-301, fmt::format("Cannot compute size of a directory: {}", absPath.string()));
```
JDuffeyBQ: Would it make more sense for the error message to be something like "Path '{}' is a directory not a CSV file"?

PR #903, opened 2024-04-02
ENH/BUG: Add DataStructure::validateGeometries() method

imikejackson:
This method is run by IFilter::execute() after the actual filter's executeImpl() is run. The code will validate that the geometry's Attribute Matrices are correctly sized to match the vertex/edge/face/cell shared lists.

Some filters and unit tests bugs were also found and fixed by this update.

### DIFF
--- a/src/simplnx/DataStructure/DataStructure.cpp
+++ b/src/simplnx/DataStructure/DataStructure.cpp
@@ -1035,4 +1035,22 @@ Result<> DataStructure::validateGeometries() const
   return result;
 }
 
+Result<> DataStructure::validateAttributeMatrices() const
+{
+  Result<> result;
+  for(const auto& dataObject : m_DataObjects)
+  {
+    auto dataObjectType = dataObject.second.lock()->getDataObjectType();
+    if(dataObjectType == DataObject::Type::AttributeMatrix)
+    {
+      auto* attrMatPtr = dynamic_cast<AttributeMatrix*>(dataObject.second.lock().get());
+      if(nullptr != attrMatPtr)
+      {
+        result = MergeResults(attrMatPtr->validate(), result);
+      }
+    }
+  }
+  return result;
+}
+
 } // namespace nx::core
--- a/src/simplnx/DataStructure/DataStructure.cpp
+++ b/src/simplnx/DataStructure/DataStructure.cpp
@@ -3,6 +3,7 @@
 #include "simplnx/Core/Application.hpp"
 #include "simplnx/DataStructure/BaseGroup.hpp"
 #include "simplnx/DataStructure/DataGroup.hpp"
+#include "simplnx/DataStructure/Geometry/IGeometry.hpp"
 #include "simplnx/DataStructure/IDataArray.hpp"
 #include "simplnx/DataStructure/INeighborList.hpp"
 #include "simplnx/DataStructure/LinkedPath.hpp"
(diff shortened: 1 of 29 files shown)


src/simplnx/DataStructure/DataStructure.cpp
```diff
@@ -1014,4 +1015,24 @@ Result<> DataStructure::transferDataArraysOoc()
...
+  {
+    auto dataObjectType = dataObject.second->getDataObjectType();
+    if(dataObjectType >= DataObject::Type::IGeometry && dataObjectType <= DataObject::Type::TetrahedralGeom)
+    {
+      auto* geomPtr = dynamic_cast<IGeometry*>(dataObject.second.get());
+      result = MergeResults(geomPtr->validateGeometry(), result);
+    }
+  }
+  return result;
+}
```
mmarineBlueQuartz: Geometries do not need to be in the top level of the DataStructure. Loop over m_DataObjects instead.

src/simplnx/Filter/IFilter.cpp
```diff
@@ -219,20 +219,22 @@ IFilter::ExecuteResult IFilter::execute(DataStructure& data, const Arguments& ar
...
-  Result<> executeImplResult = executeImpl(data, resolvedArgs, pipelineFilter, messageHandler, shouldCancel);
+  Result<> executeImplResult = executeImpl(dataStructure, resolvedArgs, pipelineFilter, messageHandler, shouldCancel);
   if(shouldCancel)
   {
     return {MakeErrorResult(-1, "Filter cancelled")};
   }
 
-  Result<> preflightActionsExecuteResult = MergeResults(std::move(preflightActionsResult), std::move(executeImplResult));
+  Result<> validGeometries = MergeResults(dataStructure.validateGeometries(), executeImplResult);
+
```
mmarineBlueQuartz:
```suggestion
  Result<> validGeometries = MergeResults(dataStructure.validateGeometries(), executeImplResult);
  Result<> validMatrices = MergeResults(dataStructure.validateMatrices(), executeImplResult);

```
validateGeometries() does not finish validating the AttributeMatrix and completely misses any matrix created outside of a geometry's initialization (i.e. CreateAttributeMatrixFilter). Considering that it is the AttributeMatrix that was causing import errors, this is the part that needs attention to solve that specific bug.

This will require an additional method in DataStructure and AttributeMatrix.

mmarineBlueQuartz (changes requested):
Of note, this does not actually check the AttributeMatrix validity (what prevents importing data) beyond that of a single DataArray that each AttributeMatrix that relates to a geometry contains. To solve this, each of those matrices would have to loop over their arrays and check tuple counts individually.

Similarly to any geometry not at the top-level of the DataStructure (montage groupings, perchance?), this will not find any AttributeMatrix not specifically created through a geometry. We have a CreateAttributeMatrix filter in Core, so the import dream3d error is not necessarily solved with just checking geometries. Add a `DataStructure::validateMatrices()` method as well, and both this and my previous point will be solved.

mmarineBlueQuartz (approved):

mmarineBlueQuartz (changes requested):
Fix unit tests

src/simplnx/DataStructure/DataStructure.cpp
imikejackson: Added AttributeMatrix::validate(). Worked it into the IFilter.cpp class for checks.

src/simplnx/Filter/IFilter.cpp
imikejackson: Done

mmarineBlueQuartz (approved):

PR #906, opened 2024-04-02
PY: Add python bindings to allow Node based geometries to resize correctly

imikejackson:


wrapping/python/plugins/debugging_helper.py
```diff
@@ -0,0 +1,99 @@
...
+import simplnx as nx
+
+# ------------------------------------------------------------------------------
+# This NEEDS to be executed here so that we can load the python based plugin
+#
+# You will need to REPLACE the name of the plugin and the name of the filter
+# that you are trying to debug in the next 3 lines of code
+# ------------------------------------------------------------------------------
+import DataAnalysisToolkit
+nx.load_python_plugin(DataAnalysisToolkit)
```
JDuffeyBQ, line 18: This should only be necessary if you are trying to load a pipeline with a python filter from this plugin. For just programmatically running a filter in python you could omit this though having it shouldn't harm anything.

src/nxrunner/src/nxrunner.cpp
```diff
@@ -481,8 +481,11 @@ std::vector<std::string> GetPythonPluginList()
   {
     return {};
   }
-
-  return StringUtilities::split(var, ';');
+#if defined(Q_OS_WIN)
```
JDuffeyBQ, line 484: I think I have a solution that will let just deal with one environment variable instead of both `PYTHONPATH` and `SIMPLNX_PYTHON_PLUGINS` for this that I'm testing on my reload branch.

src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
```diff
@@ -689,16 +689,44 @@ PYBIND11_MODULE(simplnx, mod)
   py::class_<RectGridGeom, IGridGeometry, std::shared_ptr<RectGridGeom>> rectGridGeom(mod, "RectGridGeom");
 
   py::class_<INodeGeometry0D, IGeometry, std::shared_ptr<INodeGeometry0D>> iNodeGeometry0D(mod, "INodeGeometry0D");
+  iNodeGeometry0D.def(
+      "resize_vertices",
+      [](INodeGeometry0D& nodeGeometry0D, int32 size) {
```
JDuffeyBQ: I tested with `usize` here and didn't have any issues. Do you have the code you were running and any error messages?

wrapping/python/plugins/debugging_helper.py
imikejackson: I thought we decided that it was needed?

src/nxrunner/src/nxrunner.cpp
imikejackson: OK. We can leave this in for now and then change over when you get your branch in.

wrapping/python/plugins/debugging_helper.py
JDuffeyBQ: It's needed for loading a pipeline from json for sure.

PR #912, opened 2024-04-12
ENH: RenameAction and Filter Add Overwrite Option

nyoungbq:
Added new overwrite functionality

Works by just deleting other object and renaming accordingly (no move semantics)

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/Algorithms/ITKImportFijiMontage.cpp
```diff
@@ -269,7 +269,7 @@ class IOHandler
         // rename grayscale array to reflect original
         {
           auto& gray = m_DataStructure.getDataRefAs<IDataArray>(imageDataPath.replaceName("gray" + imageDataPath.getTargetName()));
-          if(!gray.canRename(imageDataPath.getTargetName()))
+          if(gray.canRename(imageDataPath.getTargetName()) != 1)
```
mmarineBlueQuartz:
```suggestion
          if(gray.canRename(imageDataPath.getTargetName()) == false)
```
Keep the boolean typing.

src/simplnx/DataStructure/DataObject.hpp
```diff
@@ -205,7 +205,7 @@ class SIMPLNX_EXPORT DataObject
    * @brief Checks and returns if the DataObject can be renamed to the provided
    * value.
    * @param name
-   * @return bool
+   * @return int: 0 = false, 1 = true, 2 = duplicate object found
```
mmarineBlueQuartz:
```suggestion
   * @return bool
```

If you are going to return more than true or false, do not use bool return type but perhaps a uint8 or enum.

I would expect a duplicate object to default to failing since overwriting the existing value is not the default and never has been. I am fairly certain an int value of 2 is still true. Adding an optional parameter would add clarity on when this is expected.

However, since DataObject::canRename is not changed in the .cpp file, it would be best to just reset the `@return` description to what it was. There are no 0, 1, and 2 special return values, only true and false.

src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/Algorithms/ITKImportFijiMontage.cpp
mmarineBlueQuartz: I see the change in DataObject::canRename documentation that this is referencing, but the implementation in DataObject.cpp is unchanged. There are no special return codes to reference here.

mmarineBlueQuartz (changes requested):
Looks good, just have some minor things to tweak for readability around changes that did not seem to make it into the PR.

mmarineBlueQuartz (approved):

PR #924, opened 2024-04-22
BUG: Fix HDF5 implicit copy crashes.

### LINKED ISSUE #916: BUG: HDF5 classes prematurely close on implicit copy
### Is there an existing issue for this?

- [X] I have searched the existing issues, known issues in release notes, and documentation.

### Brief Description of the Issue and Expected Behavior

The HDF5 classes like `AttributeReader` are implicitly copy constructible since they don't have copy or move constructors or assignment operators. If a copy is created and then destructed, it will render that HDF5 id unusable for all other objects with that id. Therefore these classes should be move only and zero out the id on move.

### Version

DREAM3D NX (version 7)

### What operating system are you seeing the problem on? [Further details may be required during triage process]

Windows 10

### What hardware architecture are you using? [Further details may be required during triage process]

x86_64

### What section did you encounter the error in? [Further details may be required during triage process]

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JDuffeyBQ:
Some HDF5 classes didn't get the changes meant to fix #916. This PR fixes those ones and adds a unit test.

### DIFF
--- a/cmake/Utility.cmake
+++ b/cmake/Utility.cmake
@@ -73,12 +73,19 @@ function(simplnx_enable_warnings)
       set(SHADOW_WARNING "shadow-all")
     endif()
 
+    if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+      set(CLANG_WARNINGS
+        -Werror=implicit-exception-spec-mismatch # Wimplicit-exception-spec-mismatch: function previously declared with an explicit/implicit exception specification redeclared with an implicit/explicit exception specification
+      )
+    endif()
+
     target_compile_options(${ARG_TARGET}
       PRIVATE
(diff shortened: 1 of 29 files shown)


imikejackson (changes requested):
function previously declared with an explicit exception specification redeclared with an implicit exception specification [-Wimplicit-exception-spec-mismatch]

There are 3 warnings that should be cleaned up before merge.

JDuffeyBQ:
The CI and my machine only found one warning in `DatasetIO` (now fixed).

imikejackson (approved):

PR #960, opened 2024-05-14
ENH: CliReaderFilter-Add edge-bounding box intersection options

joeykleingers:
Adds a combo box that allows the user to set the behavior that occurs when the filter reads an edge where one vertex is inside the bounding box and the other is outside the bounding box.

Option 1: Ignore/Discard the edge
Option 2: Filter error
Option 3: Interpolate the outside vertex so that it moves closer to the inside vertex and ends up within the bounds.

Also moved all helper code from cli_tools.py script into the main filter script.


imikejackson:
Generally looks good.

- [ ] Tried to the usual UDRI file and I could not mask that model in the Z direction. I get the error:
```
Python exception: ValueError: could not broadcast input array from shape (0,) into shape (0,3)

At:
  /Users/mjackson/Workspace1/simplnx/wrapping/python/plugins/DataAnalysisToolkit/CliReaderFilter.py(237): execute_impl
```

- [ ] Out-of-Bounds behavior: I would put the order as: Ignore Edge, Interpolate Outside Vertex, Generate Filter Error. I would consider the "Generate Filter Error" as a destructive option and not have that embedded between the other 2 options. I would set the Default to "Interpolate Outside Vertex". I'm not sold on that terminology, but I don't have any better ideas

- [ ] The terminology on the "Mask X|Y|Z Dimension" isn't quite correct. 
    - Restrict X Dimension? Import X Sub-Volume?

- [ ] There are a lot of output. If the files stay small this probably isn't too bad. We may want to consider throttling the messages to no more than once a second? Thoughts?

imikejackson (changes requested):
See comments in the main conversation

imikejackson (approved):

PR #964, opened 2024-05-17
ENH: Factor Out EliminateDuplicateNodes and FillDataArray, Add Sub-Volume Reading, and Improve Error Reporting

joeykleingers:
+ Factored out the EliminateDuplicateNodes method from the ReadStlFile filter into GeometryUtilities and templated it so that it can be used with all node-based geometries.
+ Factored out the FillDataArray method from ReadHDF5Dataset filter into the HDF5 Support namespace, so that other filters can use it.
+ Added the ability to read a sub-volume of a dataset using the FillDataArray method.
+ Improved error reporting for DataStore::copyFrom and DatasetReader::readIntoSpan by returning a Result object instead of a boolean.


imikejackson (approved):

src/simplnx/Utilities/Parsing/HDF5/Readers/DatasetReader.hpp
```diff
@@ -113,12 +113,12 @@ class SIMPLNX_EXPORT DatasetReader : public ObjectReader
...
   /**
    * @brief Reads the dataset into the given span. Requires the span to be the
-   * correct size. Returns false if unable to read.
+   * correct size. Returns invalid result if unable to read.
    * @tparam T
    * @param data
    */
   template <class T>
-  bool readIntoSpan(nonstd::span<T> data) const;
+  Result<> readIntoSpan(nonstd::span<T> data, const std::optional<std::vector<hsize_t>>& start = std::nullopt, const std::optional<std::vector<hsize_t>>& count = std::nullopt) const;
```
JDuffeyBQ, line 124: The documentation should be updated for the new arguments. It's unclear what they're used for.

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #977, opened 2024-05-24
BUG: ReadCSVFileFilter-Replace illegal characters in headers.

imikejackson:
- The headers will eventually become DataObject names.

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadCSVFileFilter.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadCSVFileFilter.cpp
@@ -67,6 +67,22 @@ enum class IssueCodes
   IGNORED_TUPLE_DIMS = -200
 };
 
+StringVector RemoveIllegalCharacters(StringVector& headers)
+{
+  for(int i = 0; i < headers.size(); i++)
+  {
+    auto& headerName = headers[i];
+    // Replace all illegal characters with '_' character. The header names become array names which is the issue.
+    // This should have been taken care of in the GUI, but if someone is trying this from Python they will not have done that
+    // or if they are just reading it in through nxrunner.
+    headerName = StringUtilities::replace(headerName, "&", "_");
+    headerName = StringUtilities::replace(headerName, ":", "_");
+    headerName = StringUtilities::replace(headerName, "/", "_");
+    headerName = StringUtilities::replace(headerName, "\\", "_");
+    headerName = StringUtilities::replace(headerName, "\"", "");
+  }
+  return headers;
+}
 // -----------------------------------------------------------------------------
 Result<OutputActions> validateExistingGroup(const DataPath& groupPath, const DataStructure& dataStructure, const std::vector<std::string>& headers)
 {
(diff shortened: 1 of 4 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadCSVFileFilter.cpp
```diff
@@ -67,6 +67,22 @@ enum class IssueCodes
...
+  {
+    auto& headerName = headers[i];
+    // Replace all illegal characters with '_' character. The header names become array names which is the issue.
+    // This should have been taken care of in the GUI, but if someone is trying this from Python they will not have done that
+    // or if they are just reading it in through nxrunner.
+    headerName = StringUtilities::replace(headerName, "&", "_");
+    headerName = StringUtilities::replace(headerName, ":", "_");
+    headerName = StringUtilities::replace(headerName, "/", "_");
+    headerName = StringUtilities::replace(headerName, "\\", "_");
+    headerName = StringUtilities::replace(headerName, "\"", "");
```
joeykleingers, line 82:
```suggestion
    headerName = StringUtilities::replace(headerName, "\"", "_");
```

I think you missed an underscore here.

src/Plugins/SimplnxCore/docs/RequireMinimumSizeFeaturesFilter.md
```diff
@@ -10,6 +10,10 @@ This **Filter** removes **Features** that have a total number of **Cells** below
 
 The **Filter** can be run in a mode where the minimum number of neighbors is applied to a single **Ensemble**.  The user can select to apply the minimum to one specific **Ensemble**.
 
+## Warning: Cell Level Data Will Be Changed
+
+Cells that are part of features that are smaller than the minimum size threshold will have their DataArray values **replaced** with values from neighboring cells that are within a valid feature, i.e., features that are removed have their cell level data replaced with neighboring larger features.
+
```
joeykleingers: This is not related to the CSV bug fix and should probably be in its own PR.

src/simplnx/Utilities/StringUtilities.hpp
```diff
@@ -166,24 +166,24 @@ inline constexpr StringLiteral k_Whitespaces = " \t\f\v\n\r";
...
   }
-  while((startPos = str.find(from, startPos)) != std::string::npos)
+  while((startPos = inputString.find(from, startPos)) != std::string::npos)
   {
-    str.replace(startPos, from.length(), to);
+    inputString.replace(startPos, from.length(), to);
     startPos += to.length();
   }
-  return str;
+  return inputString;
```
joeykleingers, line 186: This is not related to the CSV bug fix and should probably be in its own PR.

src/simplnx/DataStructure/DataPath.cpp
```diff
@@ -22,7 +22,7 @@ DataPath::DataPath(std::vector<std::string> path)
   {
     if(!DataObject::IsValidName(item) && !item.empty())
     {
-      throw std::invalid_argument("DataPath: Invalid DataObject name");
+      throw std::invalid_argument(fmt::format("DataPath: Invalid DataObject name - [{}]. One of the DataObject names contains the '/' character.", fmt::join(m_Path, ",")));
```
joeykleingers, line 25: This is not related to the CSV bug fix and should probably be in its own PR.

joeykleingers (changes requested):

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadCSVFileFilter.cpp
imikejackson: @joeykleingers 
This was a conscious decision on my part. If you have a header that looks like
```
"Some:Header"
```

Then we end up with _Some_Header_ which really isn't what they user is going to want. I would expect almost every user to want to just remove those quotes.

src/simplnx/DataStructure/DataPath.cpp
imikejackson: It sort of is related since it is throwing an error about a DataObject name that is illegal. Something that the ReadCSVFileFilter could potentially throw, although not any more. But the error message is clearer as to what happened.

PR #981, opened 2024-05-29
PYTHON: Add helper methods to be able to easily create all the geometries.

joeykleingers:
Adds the following helper methods to be able to easily create all the geometries:

nx.CreateGeometry.create_image_geometry
nx.CreateGeometry.create_rect_grid_geometry
nx.CreateGeometry.create_vertex_geometry
nx.CreateGeometry.create_edge_geometry
nx.CreateGeometry.create_triangle_geometry
nx.CreateGeometry.create_quad_geometry
nx.CreateGeometry.create_tetrahedral_geometry
nx.CreateGeometry.create_hexahedral_geometry

Also adds parameter names to the ChoicesParameter binding.

### DIFF
--- a/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
+++ b/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
@@ -1525,11 +1525,9 @@ PYBIND11_MODULE(simplnx, mod)
   manualImportFinder.def("contains_module", &ManualImportFinder::containsModule, "mod_name"_a);
 
   // Geometry Helper Methods
-  py::module_ sub = mod.def_submodule("ArrayHandlingType", "How existing arrays will be handled when creating the new geometry.");
-  sub.attr("CopyArray") = CreateGeometryFilter::k_CopyArray;
-  sub.attr("MoveArray") = CreateGeometryFilter::k_MoveArray;
+  py::module_ sub = mod.def_submodule("CreateGeometry", "Submodule that contains the CreateGeometry utility methods.");
 
-  mod.def(
+  sub.def(
       "create_image_geometry",
       [](DataStructure& ds, const DataPath& geometryPath, const std::vector<uint64>& dims, const std::vector<float32>& origin, const std::vector<float32>& spacing,
          const std::string& cellAttrMatrixName) {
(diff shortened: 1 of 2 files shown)


imikejackson:
We talked about the API returning a Result<> object.

joeykleingers:
> We talked about the API returning a Result<> object.

They all return Result objects.

imikejackson (approved):

JDuffeyBQ:
Were we going to merge the various `ArrayHandlingType` enums/values for these bindings?

imikejackson:
@JDuffeyBQ Those changes should be on https://github.com/BlueQuartzSoftware/simplnx/pull/983/

JDuffeyBQ:
Ah then that enum should be bound using `py::enum_` here.

src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
```diff
@@ -1520,4 +1523,170 @@ PYBIND11_MODULE(simplnx, mod)
   manualImportFinder.def("clear", &ManualImportFinder::clear);
   manualImportFinder.def("contains_path", &ManualImportFinder::containsPath, "path"_a);
   manualImportFinder.def("contains_module", &ManualImportFinder::containsModule, "mod_name"_a);
+
+  // Geometry Helper Methods
+  py::module_ sub = mod.def_submodule("ArrayHandlingType", "How existing arrays will be handled when creating the new geometry.");
+  sub.attr("CopyArray") = CreateGeometryFilter::k_CopyArray;
+  sub.attr("MoveArray") = CreateGeometryFilter::k_MoveArray;
```
JDuffeyBQ: Can be replaced by the binding made in https://github.com/BlueQuartzSoftware/simplnx/pull/983 though it needs to be updated to include all values. And the functions after this should be updated as well.

JDuffeyBQ (changes requested):

src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
imikejackson: @joeykleingers Can you take care of this? If so we could get everything we need for a release.

joeykleingers:
This is ready for review again.

imikejackson (approved):

PR #998, opened 2024-06-20
Added DataStructure::getDataAsUnsafe functions

JDuffeyBQ:
- Added replaced recursive implementation of `DataStructure::getData` with an iterative one
- Added `DataObject::isGroup` to remove need for `dynamic_cast` in `DataStructure::getData`

### DIFF
--- a/src/simplnx/DataStructure/BaseGroup.cpp
+++ b/src/simplnx/DataStructure/BaseGroup.cpp
@@ -39,6 +39,11 @@ DataObject::Type BaseGroup::getDataObjectType() const
   return Type::BaseGroup;
 }
 
+bool BaseGroup::isGroup() const
+{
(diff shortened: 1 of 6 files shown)


imikejackson:
This is needed by https://github.com/BlueQuartzSoftware/simplnx/pull/991 and https://github.com/BlueQuartzSoftware/simplnx/pull/984

imikejackson (commented):
Is there a unit test to go with these new methods

JDuffeyBQ:
Not currently. They're really just convenience functions with the only thing different that the regular functions being the `static_cast` vs `dynamic_cast`. So the only way for them to fail in a way different from `getData` is from programmer error.

imikejackson (approved):

PR #1003, opened 2024-06-25
COMP: Fix compiler warnings in ComputeSurfaceFeaturesTest

imikejackson:


src/Plugins/SimplnxCore/test/ComputeSurfaceFeaturesTest.cpp
```diff
@@ -147,7 +147,8 @@ TEST_CASE("SimplnxCore::ComputeSurfaceFeaturesFilter: 3D", "[SimplnxCore][Comput
 
     for(usize i = 0; i < featureArrayExemplary.getSize(); i++)
     {
-      REQUIRE(featureArrayExemplary[i] == createdFeatureArray[i]);
+      uint8 value = featureArrayExemplary[i] == true ? 1 : 0;
```
JDuffeyBQ:
```suggestion
      uint8 value = featureArrayExemplary[i] ? 1 : 0;
```

JDuffeyBQ (changes requested):

PR #1004, opened 2024-06-25
ENH: Update required standard to C++20

JDuffeyBQ:


nyoungbq (approved):

nyoungbq:
Some notes:
Since this PR is mainly about getting compatibility up, I don't think these are necessary here I just wanted to mention them, since they are related to the topic of the PR.
- Once this goes in we should be able to change the `nonstd::span` to `std::span` since it is included in c++20
- Some of the `enable_if` statements in the template declarations should be reviewed to see if they can be changed to `concept` declarations for readability and more centralized type gating (such as in the types.hpp file)
- Other template functions should be evaluated to see if we can use `concept` to reduce the compilation scope of the template heavy files
- Unless I am misunderstanding what `BasicStringLiteral` does in the current code base, the `StringLiteral` alias can be switched to a `std::string_view`

JDuffeyBQ:
Yeah I was waiting until this gets merged to do any improvements to keep this PR minimal.

For `nonstd::span` it actually detects if `std::span` is available and typedefs it if it is. So it will be using it for the compile but to minimize our dependencies we can remove it in the future.

The issue with `std::string_view` is that it is not guaranteed to be null terminated since it can represent an arbitrary slice of a string. Our `StringLiteral` is a null terminated string basically a `std::string_view`-like wrapper around a true string literal. This is useful for third party APIs that take `char*` where they assume it is null terminated.

PR #1009, opened 2024-06-29
FILTERS: Adding MeshIO reader and writers.

joeykleingers:
Adds the following filters:

+ ReadMeshFile
+ WriteAbaqusFile
+ WriteAnsysFile
+ WriteGmshFile
+ WriteMedFile
+ WriteTetGenFile
+ WriteVtuFile

Also exposes a new method called numpy_dtype_to_datatype to Python that converts numpy dtypes to NX DataTypes.

### DIFF
--- a/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
+++ b/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
@@ -453,6 +453,57 @@ PYBIND11_MODULE(simplnx, mod)
   dataType.value("float64", DataType::float64);
   dataType.value("boolean", DataType::boolean);
 
+  mod.def(
+      "numpy_dtype_to_datatype",
+      [](const py::dtype& dtype) {
+        if(dtype.is(py::dtype::of<int8>()))
+        {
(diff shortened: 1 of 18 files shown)


src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
```diff
@@ -453,6 +453,57 @@ PYBIND11_MODULE(simplnx, mod)
   dataType.value("float64", DataType::float64);
   dataType.value("boolean", DataType::boolean);
 
+  mod.def(
+      "numpy_dtype_to_datatype",
```
JDuffeyBQ:
```suggestion
      "convert_np_dtype_to_datatype",
```
I think adding "convert" helps make it clear that this is a function. And shortening "numpy" to "np" helps to keep it from getting too long.

JDuffeyBQ (changes requested):

imikejackson (approved):

StopkaKris:
Very cool! I will definitely be testing this in the next release.

JDuffeyBQ (approved):

PR #1011, opened 2024-06-29
ENH: Add the ability to resample images as they are read in as an image stack.

joeykleingers:


nyoungbq (approved):
Looks good, I would merge this sooner rather than later because the namespaces PR hits this filter, so I will need rebase on your work before it goes in.

PR #1020, opened 2024-07-16
ENH: SampleSurfaceMesh and GeometryMath Speed Optimizations

### LINKED ISSUE #961: BUG: GeometryMath::IsPointInPolyhedron gives erroneous results
The algorithm can give bad results depending on the random ray direction that happens to have been selected.

### LINKED ISSUE #962: PERF: GeometryMath::IsPointInPolyhedron is very slow
This is due to the using the passed in Geometry's getFaceCoordinates function. Embedded in that function is the are few calls to dataStructure->getDataRefAs<T> which is known to be very costly in time. The short workaround is to bring the needed functions out of the NodeGeometry class so that the vertex ref and the face ref can be cached. 

I'm not sure if we want to attempt to implement a caching algorithm inside of the NodeGeometry classes so that we don't have to constantly get the reference from DataStructure? Or bring the IsPointInPolyHedron _inside_ the NodeGeometry class so that the algorithm itself can cache the needed references? 

```

static Point3D<float32> getVertexCoordinate(INodeGeometry0D::SharedVertexList& vertices, usize vertId)
{
  // auto& vertices = getVerticesRef();
  const usize offset = vertId * 3;
  Point3D<float32> coordinate;
  for(usize i = 0; i < 3; i++)
  {
    coordinate[i] = vertices.at(offset + i);
  }
  return coordinate;
}

static void getFacePointIds(INodeGeometry2D::SharedFaceList& cells, usize numVertsPerFace, usize faceId, nonstd::span<usize> vertexIds)
{
  const usize offset = faceId * numVertsPerFace;
  if(offset + numVertsPerFace > cells.getSize())
  {
    return;
  }
  for(usize i = 0; i < numVertsPerFace; i++)
  {
    vertexIds[i] = cells.at(offset + i);
  }
}

static void getFaceCoordinates(INodeGeometry2D::SharedFaceList& cells, INodeGeometry0D::SharedVertexList& vertices, usize numVertsPerFace, usize faceId, nonstd::span<Point3Df> coords)
{
  std::vector<usize> verts(numVertsPerFace);
  getFacePointIds(cells, numVertsPerFace, faceId, verts);
  for(usize index = 0; index < verts.size(); index++)
  {
    coords[index] = getVertexCoordinate(vertices, verts[index]);
  }
}

/**
 * @breif !!!Uses unseeded randomness, but only for validity checks [SHOULD not affect outcomes]!!! Determines if a point is in the polyhedron
 * @param triangleGeomRef the geometry to query
 * @param faceIds the list of feature ids
 * @param faceBBs the bounding boxes of each id in the geometry
 * @param point search point
 * @param bounds overarching bounding box for all of the geometry
 * @param radius length of ray
 * distToBoundary param can be added down the line (subtract intersect point from origin)
 * @return bool
 */
template <typename T>
char IsPointInPolyhedron(const nx::core::TriangleGeom& triangleGeomRef, const std::vector<int32>& faceIds, const std::vector<BoundingBox3D<T>>& faceBBs, const Point3D<T>& point,
                         const nx::core::BoundingBox3D<T>& bounds, T radius)
{
  usize iter = 0, crossings = 0;

  //* If query point is outside bounding box, finished. */
  if(!IsPointInBox(point, bounds))
  {
    return 'o';
  }

  std::random_device randomDevice;           // Will be used to obtain a seed for the random number engine
  std::mt19937_64 generator(randomDevice()); // Standard mersenne_twister_engine seeded with rd()
  std::mt19937_64::result_type seed = static_cast<std::mt19937_64::result_type>(std::chrono::steady_clock::now().time_since_epoch().count());
  generator.seed(seed);
  std::uniform_real_distribution<T> distribution(0.0, 1.0);

  auto verticesRef = triangleGeomRef.getVerticesRef();
  auto facesRef = triangleGeomRef.getFacesRef();
  const usize numVertsPerFace = triangleGeomRef.getNumberOfVerticesPerFace();
  std::vector<usize> verts(numVertsPerFace);

  usize numFaces = faceIds.size();
  while(iter++ < numFaces)
  {
    crossings = 0;

    std::array<T, 3> eulerAngles;
    float rand1 = distribution(generator);
    float rand2 = distribution(generator);

    eulerAngles[2] = (2.0f * rand1) - 1.0f;
    float t = Constants::k_2PiF * rand2;
    float w = std::sqrt(1.0f - (eulerAngles[2] * eulerAngles[2]));
    eulerAngles[0] = w * std::cos(t);
    eulerAngles[1] = w * std::sin(t);

    // Generate and add ray to point to find other end
    Ray<T> ray(point, ZXZEuler(eulerAngles.data()), radius);

    bool doNextCheck = false;
    for(usize face = 0; face < numFaces; face++)
    {
      char code = '?';
      if(!DoesRayIntersectBox(ray, faceBBs[faceIds[face]]))
      {
        code = '0';
      }
      else
      {
        std::array<Point3D<T>, 3> coords;
        getFaceCoordinates(facesRef, verticesRef, numVertsPerFace, faceIds[face], coords);
        code = RayIntersectsTriangle(ray, coords[0], coords[1], coords[2]);
      }

      /* If ray is degenerate, then goto outer while to generate another. */
      if(code == 'p' || code == 'v' || code == 'e' || code == '?')
      {
        doNextCheck = true;
        break;
      }

      /* If ray hits face at interior point, increment crossings. */
      else if(code == 'f')
      {
        crossings++;
      }

      /* If query endpoint q sits on a V/E/F, return that code. */
      else if(code == 'V' || code == 'E' || code == 'F')
      {
        return (code);
      }

    } /* End check each face */
    if(doNextCheck)
    {
      continue;
    }
    /* No degeneracies encountered: ray is generic, so finished. */
    break;

  } /* End while loop */

  /* q strictly interior to polyhedron if an odd number of crossings. */
  if((crossings % 2) == 1)
  {
    return 'i';
  }

  return 'o';
}
```

That will fix the problem at least and offer a substantial speedup.

nyoungbq:
Fixes #962
Fixes #961
Updated Documentation

| Version | Time | CPU | Iterations |
|---------|------|-----|------------|
| Baseline | 7232 ns | 7231 ns | 75982 |
| Post-Cache | 6249 ns | 6249 ns | 107611 |
| branch-reduction | 6234 ns | 6234 ns | 104425 |

Todo:

- [x] add caching routing
- [x] strip extraneous copies
- [x] extract all major tight loop constructions
- [x] extract extraneous if statements
- [x] consolidate calculations for coditionals in tight loops
- [x] Strip `if` from `DoesRayIntersectBox`
- [x] Strip `if` from `IsPointInTriangle`
- [x] Strip `if` from `RayCrossesTriangle`
- [x] `Ray.hpp` now uses a cached variant in the algorithm to avoid inserting Eigen (after discussion) while preserving API


JDuffeyBQ:
Do you have the code that you used for the timings?

nyoungbq:
> Do you have the code that you used for the timings?

I slacked it to you because it uses hardcoding to get around additional library linkage

src/simplnx/Utilities/Math/GeometryMath.cpp
```diff
@@ -54,67 +54,67 @@ BoundingBox3Df nx::core::GeometryMath::FindBoundingBoxOfVertices(INodeGeometry0D
...
-  {
-    ur[2] = points[2][2];
-  }
+  // Avoid Branch Prediction misses by exclusively doing multiplication operations
+  // C++ (§4.5/4):
+  // An r-value of type bool can be converted to an r-value of type int,
+  // with false becoming zero and true becoming one.
+  //
+  // The standard guarantees the same for float
+  bool swap = points[1][0] < ll[0];
```
JDuffeyBQ: I'm not sure that this actually contributes to the speed up. Switching between the two versions gives similar times, and looking at the assembly this version actually generates more assembly than the other.

https://godbolt.org/z/fvYcxMfaf

src/simplnx/Common/Ray.hpp
```diff
@@ -216,5 +240,26 @@ class Ray
   PointType m_Origin;
   ZXZEulerType m_Angle;
   LengthType m_Length;
+
+  // Optional caching for speed
+  PointType m_Endpoint = {};
```
JDuffeyBQ: Right now all `Ray` will use space for the endpoint even if they aren't caching the value. You can do something similar to `Result` to have different members depending on the template arguments.

src/simplnx/Utilities/Math/GeometryMath.hpp
```diff
@@ -23,6 +23,49 @@ class VertexGeom;
...
+  {
+  }
+
+  const AbstractDataStore<float32>& VerticesStoreRef;
+  const AbstractDataStore<IGeometry::MeshIndexType>& FacesStoreRef;
+  usize NumVertsPerFace;
+};
+
+template <typename T>
+concept FloatType = std::is_floating_point_v<T>;
```
JDuffeyBQ: The standard actually has this one: https://en.cppreference.com/w/cpp/concepts/floating_point

src/simplnx/Common/Ray.hpp
```diff
@@ -16,7 +17,7 @@ namespace nx::core
  * simplify and better describe values used in GeometryMath. The rays initial alignment
  * is assumed to be with the Y-axis.
  */
-template <typename T>
+template <typename T, bool CacheEndV = false>
```
JDuffeyBQ: Adding an alias would be helpful I think. Something like:
 ```cpp
template <class T>
using CachedRay = Ray<T, true>;
```

File: src/simplnx/Utilities/Math/GeometryMath.hpp [hunk 5]
```diff
@@ -536,18 +689,23 @@ char IsPointInPolyhedron(const nx::core::TriangleGeom& faces, const std::vector<
     return 'o';
   }
 
-  std::random_device randomDevice;           // Will be used to obtain a seed for the random number engine
-  std::mt19937_64 generator(randomDevice()); // Standard mersenne_twister_engine seeded with rd()
-  std::mt19937_64::result_type seed = static_cast<std::mt19937_64::result_type>(std::chrono::steady_clock::now().time_since_epoch().count());
-  generator.seed(seed);
+  // Standard mersenne_twister_engine random seed
+  std::mt19937_64 generator(std::mt19937_64::default_seed);
```
JDuffeyBQ, line 708: Does this need to actually be random? Because the default seed is fixed. So it will always return the same values.

src/simplnx/Utilities/Math/GeometryMath.hpp
```diff
@@ -605,7 +759,8 @@ char IsPointInPolyhedron(const nx::core::TriangleGeom& faces, const std::vector<
   } /* End while loop */
 
   /* q strictly interior to polyhedron if an odd number of crossings. */
-  if((crossings % 2) == 1)
+  // Bitwise & for speed
+  if(crossings & 1)
```
JDuffeyBQ: Since this is a common optimization, the compiler can generally figure out when to apply it. In this link, the compiler is able to optimize all three to same assembly for unsigned integers. Then it differs on `value % 2 == 1` for signed integers. `value % 2 != 0` gives the same assembly for signed and unsigned. I tend to think that `value % 2 != 0` is the most readable while also giving the same assembly. Alternatively we could put it in a macro like `#define IS_ODD(value) (value % 2 != 0)`. In any case this is only a minor note and not a big deal either way.

https://godbolt.org/z/7KWsEYzex

src/simplnx/Utilities/Math/GeometryMath.hpp
imikejackson: This gives the most correct output of the 3 options available.

JDuffeyBQ (approved):

PR #1022, opened 2024-07-18
BUG: ArraySelection Parameters linked to ChoiceParameter cannot be empty

imikejackson:
- If an Array Selection Parameter is linked to a choice parameter the validate() function is called regardless whether the proper choice that activates that parameter is selected. This commit fixes the bug and restores the intended behavior.


JDuffeyBQ:
Is there a branch with a filter that had this issue?

imikejackson:
Write Triangle Geometry as STL file. Select the "Single File" option. The filter will still make you fill in the other array paths.

imikejackson:
Closing as a clean build rectifies the problem.

PR #1026, opened 2024-07-29
ENH: Fix compile warnings in FilterUtilities. Init array in DataStore when resizing

imikejackson:
FilterUtilities had a function without a return statement

- DataStore did not attempt to initialize the array with any value. This leads to difficulty debugging when the values are random or happening randomly (if the re-allocation happens to be in the same place as a once-valid array.

This change attempts to allow for the initialization value to be used. In case *no* initialization value is passed into the constructor then a default 0xABAB byte pattern is used instead.

### DIFF
--- a/src/simplnx/Common/TypesUtility.hpp
+++ b/src/simplnx/Common/TypesUtility.hpp
@@ -4,7 +4,19 @@
 #include "simplnx/Common/TypeTraits.hpp"
 #include "simplnx/Common/Types.hpp"
 
+#if defined(__clang__) && defined(__clang_major__) && defined(__APPLE__)
+#if __clang_major__ > 14
 #include <bit>
+namespace bs = std;
+#else
+#include "Bit.hpp"
+namespace bs = nx::core;
+#endif
+#else
+#include <bit>
+namespace bs = std;
+#endif
+
 #include <optional>
 #include <stdexcept>
 #include <vector>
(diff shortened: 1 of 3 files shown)


src/simplnx/Utilities/FilterUtilities.hpp
```diff
@@ -94,6 +95,9 @@ auto RunTemplateClass(DataType dataType, ArgsT&&... args)
       return ClassT<float64>(std::forward<ArgsT>(args)...)();
     }
   }
+  std::stringstream ss;
+  ss << "FilterUtilities::RunTemplateClass<> Error: dataType did not match any known type. DataType was " << to_underlying(dataType);
+  throw std::runtime_error(ss.str());
```
JDuffeyBQ:
```suggestion
  throw std::runtime_error(fmt::format("FilterUtilities::RunTemplateClass<> Error: dataType did not match any known type. DataType was {}", DataTypeToString(dataType)));
```

src/simplnx/DataStructure/DataStore.hpp
```diff
@@ -82,6 +83,47 @@ class DataStore : public AbstractDataStore<T>
   , m_NumComponents(std::accumulate(m_ComponentShape.cbegin(), m_ComponentShape.cend(), static_cast<size_t>(1), std::multiplies<>()))
   , m_NumTuples(std::accumulate(m_TupleShape.cbegin(), m_TupleShape.cend(), static_cast<size_t>(1), std::multiplies<>()))
   {
+    // Because no init value is passed into the constructor, we will use a "mudflap" style value that is easy to debug.
+    if constexpr(sizeof(T) == 1)
```
JDuffeyBQ:
```cpp
template <class T>
constexpr T GetMudflap() noexcept
{
  if constexpr(sizeof(T) == 1)
  {
    return std::bit_cast<T>(static_cast<uint8>(0xAB));
  }
  if constexpr(sizeof(T) == 2)
  {
    return std::bit_cast<T>(static_cast<uint16>(0xABAB));
  }
  if constexpr(sizeof(T) == 4)
  {
    return std::bit_cast<T>(static_cast<uint32>(0xABABABAB));
  }
  if constexpr(sizeof(T) == 8)
  {
    return std::bit_cast<T>(static_cast<uint64>(0xABABABABABABABAB));
  }
}
```
Something like this would work.

JDuffeyBQ (changes requested):

imikejackson:
Requires https://github.com/BlueQuartzSoftware/simplnx/pull/1028

src/simplnx/Common/TypesUtility.hpp
```diff
@@ -10,6 +10,33 @@
```
nyoungbq, line 23:
```suggestion
#include <bit>

```

src/simplnx/Utilities/FilterUtilities.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include "ParallelAlgorithmUtilities.hpp"
 #include "simplnx/Common/Result.hpp"
+#include "simplnx/Common/TypeTraits.hpp"
```
nyoungbq, line 5:
```suggestion
#include "simplnx/Common/TypeTraits.hpp"
#include "simplnx/Common/TypeUtilities.hpp"
```

PR #1043, opened 2024-08-08
COMP: ITKThresholdMaximumConnectedComponentsImageFilter Fix value changed warning

imikejackson:


src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/ITKThresholdMaximumConnectedComponentsImageFilter.cpp
```diff
@@ -21,7 +21,7 @@ using FilterOutputType = uint8;
 struct ITKThresholdMaximumConnectedComponentsImageFilterFunctor
 {
   uint32 minimumObjectSizeInPixels = 0u;
-  float64 upperBoundary = 65536.0;
+  float64 upperBoundary = 65536.0f;
```
JDuffeyBQ:
```suggestion
  float64 upperBoundary = 65536.0;
```

src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/ITKThresholdMaximumConnectedComponentsImageFilter.cpp
```diff
@@ -31,7 +31,23 @@ struct ITKThresholdMaximumConnectedComponentsImageFilterFunctor
...
+    else if constexpr(std::is_same_v<typename InputImageT::PixelType, uint64>)
+    {
+      auto upBound = std::min<typename InputImageT::PixelType>(upperBoundary, itk::NumericTraits<typename InputImageT::PixelType>::max());
+      filter->SetUpperBoundary(upBound);
+    }
+    else
+    {
+      auto upBound = static_cast<typename InputImageT::PixelType>(std::min<double>(upperBoundary, itk::NumericTraits<typename InputImageT::PixelType>::max()));
+      filter->SetUpperBoundary(upBound);
+    }
```
JDuffeyBQ:
```suggestion
    using PixelType = typename InputImageT::PixelType;
    using MinType = std::conditional_t<std::is_same_v<PixelType, int64> || std::is_same_v<PixelType, uint64>, PixelType, float64>;
    auto upBound = static_cast<PixelType>(std::min<MinType>(upperBoundary, itk::NumericTraits<PixelType>::max()));
    filter->SetUpperBoundary(upBound);
```
Since the `int64` and `uint64` branches are the same, I combined them.

JDuffeyBQ (changes requested):

src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/ITKThresholdMaximumConnectedComponentsImageFilter.cpp
imikejackson: The compiler should complain about a conversion from float32 to float64. ?

JDuffeyBQ: But it's a double literal? I wouldn't think that it would give a warning there.

PR #1066, opened 2024-09-05
FILT: CreateDataArray Advanced (Create Data Array + Initialize Data Array)

### LINKED ISSUE #954: Array Operations: Create Array added functionality
- Allow the user to set the component dimension shape (Comma separated list of the slowest to fastest dimension)
    - The issue here is that the current component dimension is a single int32 value. We would now be changing that to a string or array. That will mess up existing pipeline files.
- Allow the user to initialize the array instead of having to run an additional filter.
    - Just run the "Initialize Array" underneath the covers

imikejackson:


imikejackson (commented):
- We probably need to rearrange the parameter order a bit to make the flow of filling out the parameters more cohesive.
- Need a default for the name of the Data Array, the number of tuples (1) and the number of components (1). I should be able to place the filter and have the basic bits filled out with sane defaults
- Update the original Create Data Array with these same changes
- Use some logic to summarize or cut down the preflight updated value when there are more than X number of components? Maybe past 10? I tried to create an array with an 80 x 60 component dimensions...
- If there isn't already, a short python script showing the use of this filter (since there are a bunch of parameters now). The script can be just the single filter. Nothing fancy.


Screenshot 2024-09-19 at 13 32 12
Can this be cleaned up (for both filters). There is a large amount of text explaining a bunch of things that as a user I'm not sure if I should be taking action of some sort. I feel like what is being presented should be in the documentation and in its place here we should have some sort of a "summary" of what the data might look like.

I am creating an array with multiple components, a suggestion:

Multicomponent Notes: Each tuple will contain the same values for all components: |2,2,2| 9,9,9|. 
or
Multicomponent Notes: Each tuple will contain different values for all components: |1,7,9|8,2,1|

Basically, just state what is going to happen. The documentation is there for more detail explanations.

See documentation for other parameter settings.

nyoungbq (approved):

PR #1072, opened 2024-09-16
FILTER: Concatenate Data Arrays

### LINKED ISSUE #1071: Array Operations: Concatenate Arrays
- User selects multiple arrays
- All input arrays must have same number of components
- Output Array will have the sum of the input arrays tuple counts

joeykleingers:
Adds the Concatenate Data Arrays filter with unit test and documentation.


imikejackson (changes requested):
I think we may have an issue with having to set the output tuple dimensions. What if we are concatenating arrays that we don't know the size of (think feature arrays). There is no way the user can properly fill out the filter.

joeykleingers:
We talked about this during our most recent meeting, and we decided to remove the ability to set the output tuple dimensions.

The output array will have a single value as its tuple dimensions.

imikejackson (approved):

PR #1073, opened 2024-09-17
BUG/API: Histogram Sync

### LINKED ISSUE #1044: Histogram Generation between 2 filters do not agree
The filters "Calculate Frequency Histogram" and "Compute Attribute Array Statistics" do not agree on a computed histogram.

Ideally, both filters should use the exact same function to generate the histogram.

You can use the Small IN100 feature data that is generated from the "(01) Small IN100 Morphological Statistics". Use both filters to generate histograms for the "Volume Sizes" array.

## Issue 0: Extract out the Histogram Generation Function into a Utility class

The DREAM3D-NX application should be able to call this utility function with an STL compatible container to generate the histogram.

## Issue 1: Compute Attribute Array Statistics needs to store the histogram bins

The frequencies are generated but not the actual bin ranges. This filter needs to be updated to include the bin ranges. The format of the output data should be the exact same as the "Calculate Frequency Histogram" filter produces.

## Issue 2: Compute Attribute Array Statistics needs more documentation for the histogram

The docs need to indicate what the histogram options do. Better yet would be to use the same options (or as close as possible) to the "Calculate Frequency Histogram"

## Issue 3: Compute Attribute Array Statistics Histogram options should be consistent with Calculate Frequency Histogram

These 2 filters are doing the same thing (for parts), they should have the same options worded in the same way for the filters.

## Issue 4: Additional Options (At least internally) when generating the Histogram

Currently the stand-alone filter saves the upper value of each bin but that is all. Maybe we should be saving both the lower and upper bounds for each bin in the array. This can come in handy when trying to create a plot (as in the DREAM3DNX capability) where we need to know the range of each bin. Just knowing the upper or lower probably isn't enough to work out the complete bin ranges.

## Example output from the Small IN100 Volume Sizes.

Screenshot 2024-08-08 at 11 37 50

nyoungbq:
fixes #1044 

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


imikejackson (approved):

src/simplnx/DataStructure/IDataStore.hpp
```diff
@@ -73,6 +73,15 @@ class SIMPLNX_EXPORT IDataStore
     return getNumberOfTuples() * getNumberOfComponents();
   }
 
+  /**
+   * @brief Returns the number of values stored within the DataStore.
+   * @return usize
+   */
+  usize size() const
+  {
+    return getNumberOfTuples() * getNumberOfComponents();
```
JDuffeyBQ:
```suggestion
    return getSize();
```
We can implement `size()` using `getSize()`.

src/simplnx/Utilities/HistogramUtilities.hpp
```diff
@@ -0,0 +1,347 @@
...
+  FillBinRanges(binRangesStore, rangeMinMax, numBins, increment);
+
+  for(usize i = 0; i < inputStore.size(); i++)
+  {
+    if(shouldCancel)
+    {
+      return MakeErrorResult(-23763, fmt::format("HistogramUtilities::{}: Signal Interrupt Received. {}:{}", __func__, __FILE__, __LINE__));
+    }
+    const auto bin = CalculateBin(inputStore[i], rangeMinMax.first, increment);
+    if((bin >= 0) && (bin < numBins))
```
JDuffeyBQ, line 134: `bool` histograms should probably be a special case.

JDuffeyBQ (changes requested):

imikejackson (approved):

JDuffeyBQ (approved):

PR #1084, opened 2024-09-25
Better preflight values for CreateDataArrayAdvanced & InitializeData.

joeykleingers:


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/InitializeData.hpp
```diff
@@ -25,6 +24,198 @@ enum StepType : uint64
   Subtraction
 };
 
+inline std::string CreateComponentValuesString(const std::vector<int64>& componentValues, usize numComps)
```
nyoungbq: I would remove the inline here and in the subsequent untemplated functions, and instead declare the functions here but define them in the source file. 

To be moved:
- [ ] CreateComponentValuesString(componentValues, numComps)
- [ ] CreateComponentValuesString(componentValuesStr, numComps)
- [ ] CreateFillValuesInitPreflightValues()
- [ ] CreateIncrementalValuesInitPreflightValues()
- [ ] CreateRandomValuesInitPreflightValues()

The compilers reserve the right to ignore the `inline` and these functions are too complex to be inlined implicitly, thus chances are it will be ignored on the modern compiler versions. This means the functions would be header defined for no reason, leading to unnecessary bloat in compiled objects containing this header.

nyoungbq: Optional: Cut down on the function naming, just very verbose. I would argue that details about functions should be in the code documentation, rather than encoded. This is just a personal preference though, no obligation to implement.

Suggestions:
- CreateCompValString()
- CreateFillValPreflight()
- CreateIncrementalValPreflight()
- CreateRandomValPreflight()

joeykleingers: I use these functions in two different filters (InitializeData and CreateDataArrayAdvanced).  Where should I be keeping functions that are hyper-specific to two filters?

nyoungbq (approved):

PR #1098, opened 2024-10-04
Removed clang-format off as it is no longer needed

JDuffeyBQ:


imikejackson:
What version of CLang Format do I need? This failed on my local machine

JDuffeyBQ:
The CI is using 10.

imikejackson:
I use a slightly newer version of ClangFormat and it still formats that line of code. I think we should just leave it in until we can figure out the differences.

JDuffeyBQ:
Sounds good.

PR #1099, opened 2024-10-04
Updated macos-12 to macos-14-large (x86)

JDuffeyBQ:
Updated to have parity with arm64 build.


imikejackson (approved):

imikejackson (changes requested):
Did we want macOS 13 or 14?

JDuffeyBQ:
Unfortunately x86 macOS runners past 12 are "large" runners which are paid runners.

PR #1107, opened 2024-10-16
FILT: Add 'Read GrainMapper3D' filter

imikejackson:
[https://xnovotech.com/3d-crystallographic-imaging-software/](https://xnovotech.com/3d-crystallographic-imaging-software/)


nyoungbq (approved):

nyoungbq (approved):
Looks good to me. A possible feature that can be added in a later version would be file caching, so it doesn't have to be re-read in execute after preflight. That said it may be a fairly fast format so parsing twice may not matter.

PR #1114, opened 2024-10-21
BUG: WriteASCIIDataFilter-Fix bug writing multi-component arrays to multiple files

imikejackson:
For the very first tuple, only the first component would be written in the first line, then the remaining components would be on the second line, in addition to the components for the second index.

- Added example pipeline to convert orientations.
- Added example pipeline to unit tests.


imikejackson:
I'm not sure how this got past our unit tests.

nyoungbq (approved):

PR #1126, opened 2024-11-03
FILT: Add Oxford Channel 5 Binary file reader.

imikejackson:
- Requires EbsdLib 1.0.34 or higher
- Filter requires BOTh .cpr and .crc files to be present
- Needs a unit test

- [x] Requires https://github.com/BlueQuartzSoftware/simplnx-registry/pull/18
- [x] Requires https://github.com/BlueQuartzSoftware/simplnx/pull/1122

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/ReadChannel5DataFilter.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/ReadChannel5DataFilter.cpp
@@ -1,0 +1,277 @@
+#include "ReadChannel5DataFilter.hpp"
+
+#include "OrientationAnalysis/Filters/Algorithms/ReadChannel5Data.hpp"
+
+#include "simplnx/DataStructure/DataPath.hpp"
+#include "simplnx/DataStructure/Geometry/ImageGeom.hpp"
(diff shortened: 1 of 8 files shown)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/ReadChannel5DataFilter.cpp
```diff
@@ -0,0 +1,288 @@
...
+namespace
+{
+namespace SIMPL
+{
+} // namespace SIMPL
+} // namespace
+
+Result<Arguments> ReadChannel5DataFilter::FromSIMPLJson(const nlohmann::json& json)
+{
+  return {};
```
JDuffeyBQ: Does this filter correspond to a SIMPL filter?

imikejackson: No.

JDuffeyBQ: Then we don't need to provide this function.

JDuffeyBQ (approved):

PR #1133, opened 2024-11-19
PYTHON: Expose methods to convert types.

joeykleingers:
+ Expose method to convert numpy types to NumericType.
+ Expose method to convert DataType to NumericType.

### DIFF
--- a/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
+++ b/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
@@ -561,7 +561,7 @@ PYBIND11_MODULE(simplnx, mod)
           return NumericType::float64;
         }
 
-        std::string dtypeStr = py::str(dtype);
+        std::string dtypeStr = py::str(static_cast<py::object>(dtype));
         throw std::invalid_argument(fmt::format("Unable to convert dtype to NumericType: Unsupported dtype '{}'.", dtypeStr));
       },
       "Convert numpy dtype to simplnx NumericType", "dtype"_a);
--- a/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
+++ b/src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
@@ -506,6 +506,55 @@ PYBIND11_MODULE(simplnx, mod)
       },
       "Convert numpy dtype to simplnx DataType", "dtype"_a);
 
+  mod.def(
+      "convert_np_dtype_to_numerictype",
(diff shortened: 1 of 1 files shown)


src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
```diff
@@ -506,6 +506,55 @@ PYBIND11_MODULE(simplnx, mod)
       },
       "Convert numpy dtype to simplnx DataType", "dtype"_a);
 
+  mod.def(
+      "convert_np_dtype_to_numerictype",
```
JDuffeyBQ:
```suggestion
      "convert_np_dtype_to_numeric_type",
```
For consistency with the other function name.

src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
```diff
@@ -506,6 +506,55 @@ PYBIND11_MODULE(simplnx, mod)
...
+        if((dtype.attr("__eq__")(py::dtype::of<float32>())).cast<bool>())
+        {
+          return NumericType::float32;
+        }
+        if((dtype.attr("__eq__")(py::dtype::of<float64>())).cast<bool>())
+        {
+          return NumericType::float64;
+        }
+
+        std::string dtypeStr = py::str(static_cast<py::object>(dtype));
```
JDuffeyBQ, line 564:
```suggestion
        std::string dtypeStr = py::str(dtype);
```

src/Plugins/SimplnxCore/wrapping/python/simplnxpy.cpp
```diff
@@ -506,6 +506,55 @@ PYBIND11_MODULE(simplnx, mod)
       },
       "Convert numpy dtype to simplnx DataType", "dtype"_a);
 
+  mod.def(
+      "convert_np_dtype_to_numerictype",
+      [](const py::dtype& dtype) {
+        if((dtype.attr("__eq__")(py::dtype::of<int8>())).cast<bool>())
```
JDuffeyBQ: I think since we're using this in multiple places now it makes sense to make a helper function.
```cpp
/**
 * @brief Equivalent to lhs.__eq__(rhs) in python
 * @param lhs
 * @param rhs
 * @return bool
 */
bool PyIsEqual(py::handle lhs, py::handle rhs)
{
  return (lhs.attr("__eq__")(rhs)).cast<bool>();
}
```

JDuffeyBQ (changes requested):

imikejackson (approved):

PR #1143, opened 2024-11-23
BUG: Fixes apps crashing due to nullptr in std::mutex

imikejackson:
Microsoft broke STL ABI compatibility with a recent VS update. Using this macro will prevent the crashing.


imikejackson:
Not needed. Will fix at the packaging level. If you need to package NXRunner by itself make sure you also ship the latest MSVC redistributable package with NXRunner.

PR #1146, opened 2024-11-28
ENH: Improve performance of the Surface Mesh to Regular Grid.

imikejackson:


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/SliceTriangleGeometry.hpp
```diff
@@ -53,7 +59,7 @@ class SIMPLNXCORE_EXPORT SliceTriangleGeometry
 protected:
   using TriStore = AbstractDataStore<INodeGeometry2D::SharedFaceList::value_type>;
   using VertsStore = AbstractDataStore<INodeGeometry0D::SharedVertexList::value_type>;
-  usize determineBoundsAndNumSlices(float32& minDim, float32& maxDim, usize numTris, TriStore& tris, VertsStore& triVerts);
+  // usize determineBoundsAndNumSlices(float32& minDim, float32& maxDim, usize numTris, TriStore& tris, VertsStore& triVerts);
```
nyoungbq:
```suggestion
```

src/simplnx/Utilities/GeometryUtilities.cpp
```diff
@@ -250,3 +250,436 @@ Result<> GeometryUtilities::ComputeTriangleNormals(const nx::core::TriangleGeom*
...
+      minDim = zStart;
+    }
+    // if(m_InputValues->Zend < maxDim)
+    {
+      maxDim = zEnd;
+    }
+  }
+  // TODO: Is this still correct? Why have the subtraction at all?
+  const auto numberOfSlices = static_cast<usize>((maxDim - minDim) / sliceResolution) + 1;
+  // const auto numberOfSlices = static_cast<usize>((maxDim - minDim) / m_InputValues->SliceResolution) + 1;
```
nyoungbq:
```suggestion
```

src/simplnx/Utilities/GeometryUtilities.cpp
```diff
@@ -250,3 +250,436 @@ Result<> GeometryUtilities::ComputeTriangleNormals(const nx::core::TriangleGeom*
...
+    {
+      minDim = zStart;
+    }
+    // if(m_InputValues->Zend < maxDim)
+    {
+      maxDim = zEnd;
+    }
+  }
+  // TODO: Is this still correct? Why have the subtraction at all?
+  const auto numberOfSlices = static_cast<usize>((maxDim - minDim) / sliceResolution) + 1;
```
nyoungbq: A little confused on what's going on here. Why are the if checks commented out? --- and if its because they are implicitly true, I would remove the code blocks so it doesn't seem accidental. What is the TODO about? The subtraction seems to be finding the diff which would be used to determine the total number of points, which is then divided by points per slice to determine the total number of slices.

src/simplnx/Utilities/GeometryUtilities.cpp
```diff
@@ -250,3 +250,436 @@ Result<> GeometryUtilities::ComputeTriangleNormals(const nx::core::TriangleGeom*
...
+          regionIds.push_back(regionId);
+        }
+        edgeCounter++;
+      }
+    }
+  }
+
+  return {std::move(slicedVerts), std::move(sliceIds), std::move(regionIds), numberOfSlices};
+}
+#endif
```
nyoungbq: Why is this macro'd out? If it is for debugging purposes or part of a TODO, document it with a comment. Else remove.

src/simplnx/Utilities/GeometryUtilities.cpp
imikejackson: Fixed.

src/simplnx/Utilities/GeometryUtilities.cpp
imikejackson: Removed.

nyoungbq (approved):

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/SliceTriangleGeometry.cpp
```diff
@@ -321,63 +88,51 @@ Result<> SliceTriangleGeometry::operator()()
...
-        maxDim = triVerts[3 * vert + 2];
+        (*triRegionIds)[currentEdgeListSize] = (*triRegionIds)[edgeIdx];
       }
+      currentEdgeListSize++;
     }
   }
-
-  // adjust sectioning range if user selected a specific range - check that user range is within actual range
-  if(m_InputValues->SliceRange == 1)
+  if(numEdges != uniqueEdges.size())
```
nyoungbq, line 131: Inside this if statement, you resize the edge list, but how can you guarantee the edges being chopped off the array are the duplicates? Are the edges pre-generated/pre-sorted such that the duplicates are on the end? Or is it that the edges aren't actually calculated here and you are just defining lengths?

I considered maybe one of the resize functions did it implicitly, but the `resizeEdgeList()` function doesn't recalculate the edges.

src/Plugins/SimplnxCore/test/CreateAMScanPathsTest.cpp
```diff
@@ -63,37 +63,38 @@ TEST_CASE("SimplnxCore::CreateAMScanPathsFilter: Valid Filter Execution", "[Simp
...
-    DataPath computedDataArray = k_ComputedScanVectorsPath.createChildPath(k_EdgeData).createChildPath(k_SliceIdsPath.getTargetName());
-    UnitTest::CompareArrays<int32>(dataStructure, exemplarDataArray, computedDataArray);
-  }
+// Write out the .dream3d file now
+#ifndef SIMPLNX_WRITE_TEST_OUTPUT
+  std::cout << "Writing File: " << fmt::format("{}/create_am_scan_paths_test.dream3d", unit_test::k_BinaryTestOutputDir) << "\n";
+  WriteTestDataStructure(dataStructure, fmt::format("{}/create_am_scan_paths_test.dream3d", unit_test::k_BinaryTestOutputDir));
+#endif
+
+  //  // Compare the exemplar and the computed outputs
```
nyoungbq: Are these comments here for debugging or artifacts of test updates? If the former put a comment mentioning the use case for future updaters

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/SliceTriangleGeometry.cpp
imikejackson: This is a bit non-obvious but in the loop above this section starting at line 115 we are effectively rewriting through the shared edges array. If we encounter an edge that has already been written (through the std::set<> variable) then we skip writing that edge which means we also skip incrementing the `currentEdgeListSize` variable.
   
 So when we finally hit that `if` statement on line 131 if the 2 sizes do NOT match we resize the edges array which by the algorithm the new size is always smaller in value than the previous size. The way that DataArray works when you resize, it will always copy what elements it can from the old array into the new array.

src/Plugins/SimplnxCore/test/CreateAMScanPathsTest.cpp
imikejackson: Well crap. I should have enabled some of these comparisons. I will get a new commit pushed up.

nyoungbq (approved):

PR #1157, opened 2024-12-06
ENH: Fixes algorithm to compute shape factors from Triangle geometry

### LINKED ISSUE #1139: BUG: Compute Feature Shapes for Triangle Geometries are broken
The filter does not produce correct results for simple shapes.

The archive has the necessary input files and pipeline file.

[Archive.zip](https://github.com/user-attachments/files/17871077/Archive.zip)

Here is a chart showing what the approximate Omega3 Values should be.

SG_Omega3_Chart

The output values for _all_ shapes = 1.0. That is not correct.

This image shows the example shapes that are being calculated. 

Screenshot 2024-11-22 at 09 48 32

imikejackson:


src/simplnx/Utilities/GeometryHelpers.hpp
```diff
@@ -32,6 +34,146 @@ SIMPLNX_EXPORT std::string GenerateGeometryInfo(const nx::core::SizeVec3& dims,
 
 namespace Connectivity
 {
+namespace detail
+{
+constexpr uint64 k_MaxOptimizedValue = static_cast<uint64>(std::numeric_limits<uint32>::max());
```
JDuffeyBQ:
```suggestion
inline constexpr uint64 k_MaxOptimizedValue = static_cast<uint64>(std::numeric_limits<uint32>::max());
```

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #1178, opened 2025-01-15
ENH: Add LLDB pretty printers for a few SIMPLNX classes.

imikejackson:
* More to come later

There is a bit of setup that one must do in order to get LLDB to use these pretty
   printers. Namely create a file called `.lldbinit` at the top level of your home
   directory and put the following code in it:

```
command script import /Users/${USER}/Workspace1/simplnx/scripts/lldb_pretty_printers/data_path_pretty_printer.py

command script import /Users/${USER}/Workspace1/simplnx/scripts/lldb_pretty_printers/nx_core_array_pretty_printer.py
```

Where you will need to adjust the path to the actual python files based on how you have your environment setup.

### DIFF
--- a/scripts/lldb_pretty_printers/data_path_pretty_printer.py
+++ b/scripts/lldb_pretty_printers/data_path_pretty_printer.py
@@ -1,0 +1,36 @@
+import lldb
+
+def data_path_summary(valobj, internal_dict):
+    """
+    Return a string summary for nx::core::DataPath.
+    We'll display the contents of m_Path (std::vector<std::string>)
+    as a list of unquoted string values.
+    """
+    path_valobj = valobj.GetChildMemberWithName('m_Path')
+
+    # Use LLDB's synthetic children for std::vector to retrieve each element.
+    count = path_valobj.GetNumChildren()
+    elements = []
+    for i in range(count):
+        element = path_valobj.GetChildAtIndex(i)
+        # For std::string, GetSummary() often returns the string in quotes, e.g. "\"Hello\""
+        str_val = element.GetSummary()
+        if str_val is None:
+            str_val = "<NULL>"
+        else:
+            # Strip surrounding double quotes if present
+            if len(str_val) >= 2 and str_val.startswith('"') and str_val.endswith('"'):
+                str_val = str_val[1:-1]
+        elements.append(str_val)
+
+    # Join all the strings into a single display
+    return f"\"{'/'.join(elements)}\""
+
+def __lldb_init_module(debugger, internal_dict):
+    """
+    LLDB will call this function automatically when the script is imported.
+    We register our pretty-printer (summary) for the C++ type nx::core::DataPath.
+    """
+    debugger.HandleCommand(
+        'type summary add -F data_path_pretty_printer.data_path_summary "nx::core::DataPath"'
+    )
--- a/scripts/lldb_pretty_printers/nx_core_array_pretty_printer.py
+++ b/scripts/lldb_pretty_printers/nx_core_array_pretty_printer.py
@@ -1,0 +1,83 @@
+import lldb
+
+def array_summary(valobj, internal_dict):
+    """
+    Custom summary function for nx::core::Array<T, Dimensions>.
+    Shows the elements of the private std::array<T, Dimensions> m_Array.
+    """
+
+    # 1) Retrieve the m_Array child.
+    std_array_elements = valobj.GetChildMemberWithName("m_Array").GetChildMemberWithName("__elems_")
+
+    # 2) For an std::array<T, N>, LLDB usually shows each element as a child.
+    #    We'll iterate through them and collect their values/summaries.
+    count = std_array_elements.GetNumChildren()
+
+    elements = []
+    for i in range(count):
+        elem = std_array_elements.GetChildAtIndex(i)
+        # Prefer GetValue() if it's a scalar; otherwise, fallback to GetSummary().
+        elem_value = elem.GetValue()
+        if elem_value is None:
+            elem_value = elem.GetSummary()
+        if elem_value is None:
+            elem_value = "<unavailable>"
+        elements.append(elem_value)
+
+    # 4) Return a concise summary.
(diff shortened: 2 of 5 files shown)


scripts/lldb_pretty_printers/nx_core_array_pretty_printer.py
```diff
@@ -0,0 +1,83 @@
+import lldb
+
+def array_summary(valobj, internal_dict):
+    """
+    Custom summary function for nx::core::Array<T, Dimensions>.
+    Shows the elements of the private std::array<T, Dimensions> m_Array.
+    """
+
+    # 1) Retrieve the m_Array child.
+    std_array_elements = valobj.GetChildMemberWithName("m_Array").GetChildMemberWithName("__elems_")
```
JDuffeyBQ: This uses the libc++ implementation which means it won't work with libstdc++ which is typical when using clang and lldb on linux.

scripts/lldb_pretty_printers/data_path_pretty_printer.py
```diff
@@ -0,0 +1,36 @@
...
+    Return a string summary for nx::core::DataPath.
+    We'll display the contents of m_Path (std::vector<std::string>)
+    as a list of unquoted string values.
+    """
+    path_valobj = valobj.GetChildMemberWithName('m_Path')
+
+    # Use LLDB's synthetic children for std::vector to retrieve each element.
+    count = path_valobj.GetNumChildren()
+    elements = []
+    for i in range(count):
```
JDuffeyBQ, line 14: I looked at the documentation and fortunately `lldb.SBValue` has `__iter__` so you can do `for element in path_valobj`.

scripts/lldb_pretty_printers/nx_core_array_pretty_printer.py
```diff
@@ -0,0 +1,83 @@
...
+    We register a summary for the templated type:
+    'nx::core::Array<*, *>' — the wildcard syntax tells LLDB to match
+    any T and Dimensions combination.
+    
+    Alternatively, you can use a regex to match all template instantiations
+    of nx::core::Array if the wildcard approach does not work in your LLDB version:
+        type summary add -x 'nx::core::Array<.*>' -F nx_core_array_pretty_printer.array_summary
+    """
+    debugger.HandleCommand(
+       'type summary add -F nx_core_array_pretty_printer.array_summary "nx::core::Array<*,*>"'
```
JDuffeyBQ:
```suggestion
       'type summary add -F nx_core_array_pretty_printer.array_summary -x "nx::core::Array<.*,.*>"'
```
The `-x` is what makes the string be interpreted as a regex so it's needed here but not on the others.

scripts/lldb_pretty_printers/data_path_pretty_printer.py
```diff
@@ -0,0 +1,36 @@
+import lldb
+
+def data_path_summary(valobj, internal_dict):
+    """
+    Return a string summary for nx::core::DataPath.
+    We'll display the contents of m_Path (std::vector<std::string>)
+    as a list of unquoted string values.
+    """
+    path_valobj = valobj.GetChildMemberWithName('m_Path')
```
JDuffeyBQ, line 9: We might consider a different approach since `m_Path` is a private variable and an implementation detail so if it changes then this breaks. It is possible to call expressions in lldb like so: `valobj.EvaluateExpression('toString()')`. I think it's worth looking into at least.

JDuffeyBQ (commented):
We could combine all pretty printers into one package so that you only have one import to do for lldb. It would also let use share utility code between printers.

scripts/lldb_pretty_printers/nx_core_array_pretty_printer.py
imikejackson: What is the other option?

scripts/lldb_pretty_printers/data_path_pretty_printer.py
imikejackson: Unfortunately that does not work on macOS 14. Will stay with the implementation details version.
Screenshot 2025-02-19 at 09 43 23

JDuffeyBQ:
I have this script for testing:

`nxprinter.py`
```python
import lldb

def data_path_summary(valobj: lldb.SBValue, internal_dict: dict) -> str:
  path_valobj: lldb.SBValue = valobj.GetChildMemberWithName('m_Path')

  elements = []
  element: lldb.SBValue
  for element in path_valobj:
    str_val: str = element.GetSummary()
    if str_val is None:
      str_val = '<NULL>'
    else:
      if len(str_val) >= 2 and str_val.startswith('"') and str_val.endswith('"'):
        str_val = str_val[1:-1]
    elements.append(str_val)

  return f"\"{'/'.join(elements)}\""

def array_summary(valobj: lldb.SBValue, internal_dict: dict) -> str:
  array_valobj: lldb.SBValue = valobj.GetChildMemberWithName('m_Array')

  # should "_M_elems" for libstdc++ and "__elems" for libc++
  std_array_elements: lldb.SBValue = array_valobj.child[0]

  if std_array_elements is None:
    return '<unavailable>'

  elements = []
  element: lldb.SBValue
  for element in std_array_elements:
    elem_value = element.GetValue()
    if elem_value is None:
      elem_value = element.GetSummary()
    if elem_value is None:
      elem_value = '<unavailable>'
    elements.append(elem_value)

  return f"({', '.join(elements)})"


def add_type_summary(debugger: lldb.SBDebugger, func_name: str, type_name: str, type_name_is_regex: bool = False) -> None:
  regex_flag = '-x' if type_name_is_regex else ''
  debugger.HandleCommand(
    f'type summary add -F nxprinter.{func_name} {regex_flag} "{type_name}"'
  )

def __lldb_init_module(debugger: lldb.SBDebugger, internal_dict: dict):
  add_type_summary(debugger, 'data_path_summary', 'nx::core::DataPath')
  add_type_summary(debugger, 'array_summary', 'nx::core::Array<.*,.*>', type_name_is_regex=True)
```

PR #1200, opened 2025-02-12
BUG: Disables data transfer from voxels to triangles for SurfaceNets

imikejackson:
The transfer function does not work correctly with the SurfaceNets codes. This should be a temporary situation until a fix is engineered.


JDuffeyBQ (approved):
Before the next release if the bug is not fixed, the parameters json version should be updated.

PR #1220, opened 2025-03-03
BUG: Fixes bugs with Trilinear Interpolation in Apply Transformation to Geometry

imikejackson:
Specifically for Image Geometry Transforms.

### DIFF
--- a/src/simplnx/Utilities/ImageRotationUtilities.cpp
+++ b/src/simplnx/Utilities/ImageRotationUtilities.cpp
@@ -1,11 +1,11 @@
-
 #include "ImageRotationUtilities.hpp"
+
+#include "simplnx/Utilities/Math/MatrixMath.hpp"
 
 #include <algorithm>
 
 namespace nx::core::ImageRotationUtilities
 {
-
 //------------------------------------------------------------------------------
 FloatVec6 DetermineMinMaxCoords(const BoundingBox3Df& imageGeomBoundingBox, const Matrix4fR& transformationMatrix)
 {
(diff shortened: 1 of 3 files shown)


src/simplnx/Utilities/ImageRotationUtilities.cpp
```diff
@@ -242,7 +242,15 @@ ImageRotationUtilities::Matrix4fR GenerateScaleTransformationMatrix(const Vector
   return transformationMatrix;
 }
 
-//------------------------------------------------------------------------------
+#define IR_SQUARE(value) ((value) * (value))
```
JDuffeyBQ:
```suggestion
template <class T>
constexpr T square(T value) noexcept
{
  return value * value;
}
```
We can pick a different name if we want.

src/simplnx/Utilities/ImageRotationUtilities.hpp
```diff
@@ -224,6 +224,10 @@ static const OctantOffsetArrayType k_IndexOffset7 = {Vector3i64{-1, 0, 0}, Vecto
                                                      Vector3i64{-1, 0, 1}, Vector3i64{0, 0, 1}, Vector3i64{0, 1, 1}, Vector3i64{-1, -1, 1}};
 static const std::array<OctantOffsetArrayType, 8> k_AllOctantOffsets{k_IndexOffset0, k_IndexOffset1, k_IndexOffset2, k_IndexOffset3, k_IndexOffset4, k_IndexOffset5, k_IndexOffset6, k_IndexOffset7};
 
+template <class T>
+// using MaxSizeType = std::conditional_t<std::is_floating_point_v<T>, float64, std::conditional_t<std::is_signed_v<T>, int64, uint64>>;
```
JDuffeyBQ:
```suggestion
```

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #1225, opened 2025-03-05
BUG: Fixes creating arrays when OoC preferences files says 'In-Memory'

### LINKED ISSUE #1035: BUG: Preferences can use wrong value for UseOoc and disable parallelization
```cpp
void Preferences::checkUseOoc()
{
  m_UseOoc = !value(k_PreferredLargeDataFormat_Key).get<std::string>().empty();
}
```
```cpp
IParallelAlgorithm::IParallelAlgorithm()
{
#ifdef SIMPLNX_ENABLE_MULTICORE
  // Do not run OOC data in parallel by default.
  m_RunParallel = !Application::GetOrCreateInstance()->getPreferences()->useOocData();
#endif
}
```
This will set `m_UseOoc` to `true` if the json contains any string value. The preferences could have `"In-Memory"` and `m_UseOoc` would still be `true` and disable parallelization.

imikejackson:
The logic to determine if out-of-core is used basically says that if we think the data structure will be larger than available ram AND we have an out-of-core format available then we use out-of-core otherwise we throw an error.

### DIFF
--- a/src/simplnx/Core/Preferences.cpp
+++ b/src/simplnx/Core/Preferences.cpp
@@ -247,7 +247,13 @@ Result<> Preferences::loadFromFile(const std::filesystem::path& filepath)
 
 void Preferences::checkUseOoc()
 {
-  m_UseOoc = !value(k_PreferredLargeDataFormat_Key).get<std::string>().empty();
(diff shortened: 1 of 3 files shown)


imikejackson:
@JDuffeyBQ has better solutions. Closing this PR.

PR #1227, opened 2025-03-06
Added fallback data store format

JDuffeyBQ:
Now `DataIOCollection::createDataStore` will fallback to a standard in memory `DataStore` if the request data format is not available.


imikejackson (approved):

imikejackson:
Tested with the troublesome preference file and all seems to work correctly now.

PR #1238, opened 2025-03-17
ENH: Added pipeline relative path support

JDuffeyBQ:
Added support for interpreting relative paths in pipelines as relative to the pipeline file itself. Filters and filter parameter need to have support manually added. Currently only `FileSystemPathParameter` is added.


imikejackson:
@JDuffeyBQ Do any of the other parameters such as Read HDF5 Data set and Read H5Ebsd need to be updated?

JDuffeyBQ:
Yes. Any parameter that contains a path needs to override the `construct` function like `FileSystemPathParameter`. I wanted to get to have this looked at before committing those changing in case we needed to make a change to the design.

imikejackson (approved):

PR #1246, opened 2025-03-19
ENH: Fixes empty exception messages.

imikejackson:


src/simplnx/DataStructure/DataStore.hpp
```diff
@@ -314,7 +314,7 @@ class DataStore : public AbstractDataStore<T>
   {
     if(index >= this->getSize())
     {
-      throw std::runtime_error("");
+      throw std::runtime_error(fmt::format("DataStore::at({}) is out of bounds. Size={}", index, this->getSize()));
```
JDuffeyBQ: We could make this `std::out_of_range` which would be consistent with `std::vector` and other standard classes.

imikejackson: Good Idea.

JDuffeyBQ (approved):

PR #1263, opened 2025-04-10
ENH: Show EBSD Phase Info in Preflight Updated Values for each EBSD readers

imikejackson:


nyoungbq (commented):
The code itself looks fine, I would recommend making it into a utility function. This would allow you to segment the 3rd party include into the cpp of the utility file (preventing bloating of the filter objects themselves) and make sure the code is much easier to maintain should changes need to be made/prevent copy bugs. If we don't have a utility file it fits into already I would recommend a new one called "PreflightUtilities", this would open the door for other common paradigms to be integrated later on. (eg displaying length unit / other geometry details in updated values.

imikejackson:
While they look similar there are subtle differences in each implementation. I will take another look and see if the code can be consolidated a bit.

nyoungbq (approved):

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/ReadGrainMapper3DFilter.cpp
```diff
@@ -188,7 +191,23 @@ IFilter::PreflightResult ReadGrainMapper3DFilter::preflightImpl(const DataStruct
...
+    // auto laueOps = LaueOps::GetAllOrientationOps();
+    // int phaseIndex = 1;
+    // for(const auto& phaseInfo : phases)
+    // {
+    //   preflightUpdatedValues.push_back(
+    //       {fmt::format("{}: ", phaseIndex++), fmt::format("Material Name: {}    |    Crystal Symmetry: {}    |    Universal Hermann Mauguin: {}", phaseInfo.Name,
+    //                                                       laueOps[GetLaueIndexFromSpaceGroup(phaseInfo.SpaceGroup)]->getSymmetryName(), phaseInfo.UniversalHermannMauguin)});
+    // }
+
+    auto phases = reader.getPhaseVector();
```
nyoungbq:
```suggestion
```

nyoungbq (approved):
Funnily enough this implementation is strikingly similar to the implementation I had in mind, code looks good. Suggestion is just to clean up some artifacts

PR #1267, opened 2025-04-15
ENH: Ensure all filters are checking for cancel and sending progress messages.

### LINKED ISSUE #1254: ENH: Ensure each filter is checking for cancel and sending progress feedback messages
## Missing 'cancel()' check

- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/AlignSectionsList.hpp:55
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/AlignSectionsFeatureCentroid.hpp:58
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadRawBinary.hpp:38
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/QuickSurfaceMesh.hpp:79
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMisorientation.hpp:65
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMutualInformation.hpp:59
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.hpp:49
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeIPFColors.hpp:55
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeMisorientations.hpp:60
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadAngData.hpp:74
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadChannel5Data.hpp:80
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/RotateEulerRefFrame.cpp:67

## Missing use of m_MessageHandler

- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/CreatePythonSkeleton.cpp:47:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/FillBadData.cpp:48:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ConvertData.cpp:42:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundaryCells.cpp:43:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/LabelTriangleGeometry.cpp:44:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeTriangleGeomVolumes.cpp:42:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/FlyingEdges3D.cpp:42:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/PartitionGeometry.cpp:83:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/CreateColorMap.cpp:40:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/NearestPointFuseRegularGrids.cpp:44:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadRawBinary.cpp:39:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadStringDataArray.cpp:48:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadVtkStructuredPoints.cpp:162:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/SplitAttributeArray.cpp:40:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/TriangleCentroid.cpp:42:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/UncertainRegularGridSampleSurfaceMesh.cpp:56:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/SurfaceNets.cpp:57:34: 
- [x] SimplnxCore/src/SimplnxCore/Filters/Algorithms/SharedFeatureFace.cpp:46:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeBoundaryStrengths.cpp:50:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp:48:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeCAxisLocations.cpp:45:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureFaceMisorientation.cpp:46:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFaceIPFColoring.cpp:46:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureReferenceMisorientations.cpp:66:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeQuaternionConjugate.cpp:55:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborMisorientations.cpp:46:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeIPFColors.cpp:54:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ConvertQuaternion.cpp:46:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/CreateEnsembleInfo.cpp:44:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeSchmids.cpp:73:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeMisorientations.cpp:61:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeSlipTransmissionMetrics.cpp:49:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadAngData.cpp:73:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadEnsembleInfo.cpp:52:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadCtfData.cpp:80:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/RodriguesConvertor.cpp:44:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadChannel5Data.cpp:79:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadGrainMapper3D.cpp:67:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/WriteGBCDTriangleData.cpp:45:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/WriteINLFile.cpp:48:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/RotateEulerRefFrame.cpp:38:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/WriteGBCDGMTFile.cpp:45:34: 
- [x] OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeShapes.cpp:49:34:

imikejackson:
Addreses https://github.com/BlueQuartzSoftware/simplnx/issues/1254

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMisorientation.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMisorientation.cpp
@@ -29,6 +29,10 @@ AlignSectionsMisorientation::~AlignSectionsMisorientation() noexcept = default;
 // -----------------------------------------------------------------------------
 Result<> AlignSectionsMisorientation::operator()()
 {
+  if(m_ShouldCancel)
+  {
+    return {};
+  }
   const auto& gridGeom = m_DataStructure.getDataRefAs<IGridGeometry>(m_InputValues->ImageGeometryPath);
 
   return execute(gridGeom.getDimensions(), m_InputValues->ImageGeometryPath);
@@ -86,6 +90,10 @@ Result<> AlignSectionsMisorientation::findShifts(std::vector<int64_t>& xShifts, 
     // Loop over the Z Direction
     for(int64_t iter = 1; iter < dims[2]; iter++)
     {
+      if(m_ShouldCancel)
+      {
+        return {};
+      }
       progInt = static_cast<float>(iter) / static_cast<float>(dims[2]) * 100.0f;
       auto now = std::chrono::steady_clock::now();
       // Only send updates every 1 second
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMutualInformation.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMutualInformation.cpp
@@ -31,6 +31,10 @@ AlignSectionsMutualInformation::~AlignSectionsMutualInformation() noexcept = def
 // -----------------------------------------------------------------------------
 Result<> AlignSectionsMutualInformation::operator()()
 {
+  if(m_ShouldCancel)
+  {
+    return {};
+  }
   const auto& gridGeom = m_DataStructure.getDataRefAs<IGridGeometry>(m_InputValues->ImageGeometryPath);
 
   return execute(gridGeom.getDimensions(), m_InputValues->ImageGeometryPath);
@@ -87,6 +91,10 @@ Result<> AlignSectionsMutualInformation::findShifts(std::vector<int64>& xShifts,
     auto& cumulativeShiftsStore = m_DataStructure.getDataAs<Int64Array>(m_InputValues->CumulativeShiftsArrayPath)->getDataStoreRef();
     for(int64 iter = 1; iter < dims[2]; iter++)
     {
+      if(m_ShouldCancel)
+      {
+        return {};
+      }
       m_MessageHandler(IFilter::Message::Type::Info, fmt::format("Determining Shifts: Slice {}/{} complete", iter, dims[2]));
 
       float32 minDisorientation = std::numeric_limits<float32>::max();
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp
@@ -24,7 +24,6 @@ ComputeAvgOrientations::~ComputeAvgOrientations() noexcept = default;
 // -----------------------------------------------------------------------------
 Result<> ComputeAvgOrientations::operator()()
 {
-
   std::vector<LaueOps::Pointer> orientationOps = LaueOps::GetAllOrientationOps();
 
   nx::core::Int32Array& featureIds = m_DataStructure.getDataRefAs<Int32Array>(m_InputValues->cellFeatureIdsArrayPath);
@@ -36,7 +35,7 @@ Result<> ComputeAvgOrientations::operator()()
   nx::core::Float32Array& avgQuats = m_DataStructure.getDataRefAs<Float32Array>(m_InputValues->avgQuatsArrayPath);
   nx::core::Float32Array& avgEuler = m_DataStructure.getDataRefAs<Float32Array>(m_InputValues->avgEulerAnglesArrayPath);
 
-  size_t totalPoints = featureIds.getNumberOfTuples();
+  const size_t totalPoints = featureIds.getNumberOfTuples();
 
   auto validateNumFeatResult = ValidateFeatureIdsToFeatureAttributeMatrixIndexing(m_DataStructure, m_InputValues->avgQuatsArrayPath, featureIds, m_MessageHandler);
   if(validateNumFeatResult.invalid())
@@ -65,6 +64,10 @@ Result<> ComputeAvgOrientations::operator()()
 
   for(size_t i = 0; i < totalPoints; i++)
   {
+    if(m_ShouldCancel)
+    {
+      return {};
+    }
     if(featureIds[i] > 0 && phases[i] > 0)
     {
       int32 phase = phases[i];
@@ -92,6 +95,10 @@ Result<> ComputeAvgOrientations::operator()()
 
   for(size_t featureId = 1; featureId < totalFeatures; featureId++)
   {
+    if(m_ShouldCancel)
+    {
+      return {};
+    }
     size_t featureIdOffset = featureId * 4;
     float32 count = counts[featureId];
 
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeIPFColors.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeIPFColors.cpp
@@ -57,6 +57,10 @@ public:
     size_t index = 0;
     for(size_t i = start; i < end; i++)
     {
+      if(m_Filter->shouldCancel())
+      {
+        return;
+      }
       phase = m_CellPhases[i];
       index = i * 3;
       m_CellIPFColors.setValue(index, 0);
@@ -195,4 +199,9 @@ This indicates a problem with the input cell phase data. DREAM3D-NX will give IN
 void ComputeIPFColors::incrementPhaseWarningCount()
 {
   ++m_PhaseWarningCount;
+}
+
+bool ComputeIPFColors::shouldCancel() const
+{
+  return m_ShouldCancel;
 }
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeIPFColors.hpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeIPFColors.hpp
@@ -48,7 +48,8 @@ public:
    */
   void incrementPhaseWarningCount();
 
-protected:
+  bool shouldCancel() const;
+
 private:
   DataStructure& m_DataStructure;
   const IFilter::MessageHandler& m_MessageHandler;
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeMisorientations.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeMisorientations.cpp
@@ -22,7 +22,7 @@ inline void ComputeMisorientation(const QuatD& q1, const QuatD& q2, Float32Array
   outputMisorientations[tupleIdx * 4 + 3] = axisAngle[3] * Constants::k_180OverPiD; // Convert the output Angle to Degrees.
 }
 
-Result<> ComputeUsingArrays(DataStructure& m_DataStructure, const ComputeMisorientationsInputValues* inputValues)
+Result<> ComputeUsingArrays(DataStructure& m_DataStructure, const ComputeMisorientationsInputValues* inputValues, const std::atomic_bool& m_ShouldCancel)
 {
   const auto& cellPhases = m_DataStructure.getDataRefAs<Int32Array>(inputValues->InputPhasesArrayPath);
   const auto& crystalStructures = m_DataStructure.getDataRefAs<UInt32Array>(inputValues->InputCrystalStructuresArrayPath);
@@ -38,6 +38,10 @@ Result<> ComputeUsingArrays(DataStructure& m_DataStructure, const ComputeMisorie
 
   for(int64_t tupleIdx = 0; tupleIdx < totalPoints; tupleIdx++)
   {
+    if(m_ShouldCancel)
+    {
+      return {};
+    }
     if(cellPhases[tupleIdx] > 0) // We must have a valid phase index.
     {
       size_t laueClass = static_cast<size_t>(crystalStructures[cellPhases[tupleIdx]]);
@@ -62,7 +66,7 @@ Result<> ComputeUsingArrays(DataStructure& m_DataStructure, const ComputeMisorie
   return {};
 }
 
-Result<> ComputeUsingReferenceOrientation(DataStructure& m_DataStructure, const ComputeMisorientationsInputValues* inputValues)
+Result<> ComputeUsingReferenceOrientation(DataStructure& m_DataStructure, const ComputeMisorientationsInputValues* inputValues, const std::atomic_bool& m_ShouldCancel)
 {
   const auto& cellPhases = m_DataStructure.getDataRefAs<Int32Array>(inputValues->InputPhasesArrayPath);
   const auto& crystalStructures = m_DataStructure.getDataRefAs<UInt32Array>(inputValues->InputCrystalStructuresArrayPath);
@@ -84,6 +88,10 @@ Result<> ComputeUsingReferenceOrientation(DataStructure& m_DataStructure, const 
 
   for(int64_t tupleIdx = 0; tupleIdx < totalPoints; tupleIdx++)
   {
+    if(m_ShouldCancel)
+    {
+      return {};
+    }
     if(cellPhases[tupleIdx] > 0) // We must have a valid phase index.
     {
       size_t phase1 = static_cast<size_t>(crystalStructures[cellPhases[tupleIdx]]);
@@ -127,11 +135,11 @@ Result<> ComputeMisorientations::operator()()
 
   if(m_InputValues->ComputationType == compute_misorientations_constants::k_UseArraysIndex)
   {
-    result = ComputeUsingArrays(m_DataStructure, m_InputValues);
+    result = ComputeUsingArrays(m_DataStructure, m_InputValues, m_ShouldCancel);
   }
   else if(m_InputValues->ComputationType == compute_misorientations_constants::k_UseReferenceAxesIndex)
   {
-    result = ComputeUsingReferenceOrientation(m_DataStructure, m_InputValues);
+    result = ComputeUsingReferenceOrientation(m_DataStructure, m_InputValues, m_ShouldCancel);
   }
 
   return result;
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadAngData.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ReadAngData.cpp
@@ -41,6 +41,10 @@ Result<> ReadAngData::operator()()
   {
     return MakeErrorResult(result.first, result.second);
   }
+  if(m_ShouldCancel)
+  {
+    return {};
+  }
 
   copyRawEbsdData(&reader);
 
@@ -50,6 +54,7 @@ Result<> ReadAngData::operator()()
 // -----------------------------------------------------------------------------
 std::pair<int32, std::string> ReadAngData::loadMaterialInfo(AngReader* reader) const
 {
+
   const std::vector<AngPhase::Pointer> phases = reader->getPhaseVector();
   if(phases.empty())
   {
@@ -110,6 +115,10 @@ void ReadAngData::copyRawEbsdData(AngReader* reader) const
 
   // Adjust the values of the 'phase' data to correct for invalid values and assign the read Phase Data into the actual DataArray
   {
+    if(m_ShouldCancel)
+    {
+      return;
+    }
     auto& targetArray = m_DataStructure.getDataRefAs<Int32Array>(CellAttributeMatrixPath.createChildPath(EbsdLib::AngFile::Phases));
     int* phasePtr = reinterpret_cast<int32_t*>(reader->getPointerByName(EbsdLib::Ang::PhaseData));
     for(size_t i = 0; i < totalCells; i++)
@@ -124,6 +133,10 @@ void ReadAngData::copyRawEbsdData(AngReader* reader) const
 
   // Condense the Euler Angles from 3 separate arrays into a single 1x3 array
   {
+    if(m_ShouldCancel)
+    {
+      return;
+    }
     const auto* fComp0 = reinterpret_cast<float*>(reader->getPointerByName(EbsdLib::Ang::Phi1));
     const auto* fComp1 = reinterpret_cast<float*>(reader->getPointerByName(EbsdLib::Ang::Phi));
     const auto* fComp2 = reinterpret_cast<float*>(reader->getPointerByName(EbsdLib::Ang::Phi2));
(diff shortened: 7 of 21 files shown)


nyoungbq (approved):

JDuffeyBQ:
```cpp
class TimedMessenger
{
public:
  TimedMessenger(const IFilter::MessageHandler& messageHandler, std::chrono::milliseconds interval)
  : m_MessageHandler(messageHandler)
  , m_Interval(interval)
  {
  }

  ~TimedMessenger() noexcept = default;

  TimedMessenger(const TimedMessenger&) = delete;
  TimedMessenger(TimedMessenger&&) noexcept = delete;
  TimedMessenger& operator=(const TimedMessenger&) = delete;
  TimedMessenger& operator=(TimedMessenger&&) noexcept = delete;

  const IFilter::MessageHandler& getMessageHandler() const
  {
    return m_MessageHandler;
  }

  void sendMessage(const std::string& message)
  {
    if(!shouldSend())
    {
      return;
    }

    m_MessageHandler(IFilter::Message::Type::Info, message);
    m_InitialPoint = std::chrono::steady_clock::now();
  }

  template <class FunctorT>
  void sendMessage(FunctorT func)
  {
    if(!shouldSend())
    {
      return;
    }

    m_MessageHandler(IFilter::Message::Type::Info, func());
    m_InitialPoint = std::chrono::steady_clock::now();
  }

  bool shouldSend() const
  {
    return std::chrono::steady_clock::now() - m_InitialPoint >= m_Interval;
  }

private:
  const IFilter::MessageHandler& m_MessageHandler;
  std::chrono::milliseconds m_Interval;
  std::chrono::steady_clock::time_point m_InitialPoint = std::chrono::steady_clock::now();
};

class ThreadSafeMessenger : public TimedMessenger
{
public:
  ThreadSafeMessenger(const IFilter::MessageHandler& messageHandler, std::chrono::milliseconds interval)
  : TimedMessenger(messageHandler, interval)
  {
  }

  ~ThreadSafeMessenger() noexcept = default;

  ThreadSafeMessenger(const ThreadSafeMessenger&) = delete;
  ThreadSafeMessenger(ThreadSafeMessenger&&) noexcept = delete;
  ThreadSafeMessenger& operator=(const ThreadSafeMessenger&) = delete;
  ThreadSafeMessenger& operator=(ThreadSafeMessenger&&) noexcept = delete;

  void sendThreadSafeInfoMessage(const std::string& message)
  {
    const std::lock_guard<std::mutex> guard(m_ProgressMessageMutex);
    sendMessage(message);
  }

protected:
  std::mutex m_ProgressMessageMutex;
};

class ThreadSafeProgressMessenger : public ThreadSafeMessenger
{
public:
  ThreadSafeProgressMessenger(const IFilter::MessageHandler& messageHandler, std::chrono::milliseconds interval)
  : ThreadSafeMessenger(messageHandler, interval)
  {
  }

  ~ThreadSafeProgressMessenger() noexcept = default;

  ThreadSafeProgressMessenger(const ThreadSafeProgressMessenger&) = delete;
  ThreadSafeProgressMessenger(ThreadSafeProgressMessenger&&) noexcept = delete;
  ThreadSafeProgressMessenger& operator=(const ThreadSafeProgressMessenger&) = delete;
  ThreadSafeProgressMessenger& operator=(ThreadSafeProgressMessenger&&) noexcept = delete;

  void sendThreadSafeProgressMessage(size_t counter)
  {
    std::lock_guard<std::mutex> guard(m_ProgressMessageMutex);

    m_ProgressCounter += counter;
    auto generateMessageFunc = [this]() {
      auto progressInt = static_cast<usize>((static_cast<float32>(m_ProgressCounter) / static_cast<float32>(m_TotalElements)) * 100.0f);
      return fmt::format("{} || {}%", m_Message, progressInt);
    };
    sendMessage(generateMessageFunc);
  }

  void setTotalElements(size_t totalElements)
  {
    m_TotalElements = totalElements;
  }

  void setMessage(std::string message)
  {
    m_Message = std::move(message);
  }

private:
  std::string m_Message;
  size_t m_TotalElements = 0;
  size_t m_ProgressCounter = 0;
};
```
I think it would be good allow for reuse with the single threaded messaging as well. So this modified code is a rough idea of what I'm thinking.

src/simplnx/Utilities/ThreadedAlgorithm.cpp
```diff
@@ -0,0 +1,66 @@
...
+// -----------------------------------------------------------------------------
+const std::atomic_bool& ThreadedAlgorithm::getCancel() const
+{
+  return m_ShouldCancel;
+}
+
+const IFilter::MessageHandler& ThreadedAlgorithm::getMessageHandler() const
+{
+  return m_MessageHandler;
+}
```
nyoungbq:
```suggestion
bool ThreadedAlgorithm::shouldCancel() const
{
  return m_ShouldCancel.load();
}
```
It is not a good to return a reference in parallel. I see the use for providing cancel call without having the threads store the variable, so I have provided a thread-safe solution that uses the proper access pattern for atomic bool and will not add overhead since bool values are trivially copyable.

See here for simple breakdown: https://stackoverflow.com/questions/15624010/c-shared-mutex-with-return-by-reference

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+  ThreadedAlgorithm& operator=(const ThreadedAlgorithm&) = delete;
+  ThreadedAlgorithm& operator=(ThreadedAlgorithm&&) noexcept = delete;
+
+  const std::atomic_bool& getCancel() const;
+
+  void sendThreadSafeProgressMessage(size_t counter);
+
+  void sendThreadSafeInfoMessage(const std::string& message);
+
+  const IFilter::MessageHandler& getMessageHandler() const;
```
nyoungbq:
```suggestion
```
Returning objects by ref in parallel is a bad idea see comment on cpp file

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+public:
+  ThreadedAlgorithm(const IFilter::MessageHandler& mesgHandler, const std::atomic_bool& shouldCancel);
+  virtual ~ThreadedAlgorithm() noexcept;
+
+  ThreadedAlgorithm(const ThreadedAlgorithm&) = delete;
+  ThreadedAlgorithm(ThreadedAlgorithm&&) noexcept = delete;
+  ThreadedAlgorithm& operator=(const ThreadedAlgorithm&) = delete;
+  ThreadedAlgorithm& operator=(ThreadedAlgorithm&&) noexcept = delete;
+
+  const std::atomic_bool& getCancel() const;
```
nyoungbq:
```suggestion
  bool shouldCancel() const;
```
A thread-safe fetch function

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  void sendThreadSafeProgressMessage(size_t counter);
```
nyoungbq:
```suggestion
/**
   * @brief This function accepts a number to increment the counter by and handles all of the message 
   * assembly and send off
   * [!!Utilizes Mutex Locks!!] be wary calling this excessively in parallel as it can slow the program down
   * @param counter the number of elements the function has processed since last update progress call
   */
  void sendThreadSafeProgressMessage(size_t counter);
```

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  void sendThreadSafeInfoMessage(const std::string& message);
```
nyoungbq:
```suggestion
 /**
   * @brief sets the message to be displayed
   * [!!Utilizes Mutex Locks!!] be wary calling this excessively in parallel as it can slow the program down
   * @param progressMessage display message
   */
  void sendThreadSafeInfoMessage(std::string&& message);
```
Documentation and swap to r-value, since the string should not be modified by calling function after it has been handed off. Move statements can be used to convert it if need be.

src/simplnx/Utilities/ThreadedAlgorithm.cpp
```diff
@@ -0,0 +1,66 @@
...
+  auto progressInt = static_cast<usize>((static_cast<float32>(m_ProgressCounter) / static_cast<float32>(m_TotalElements)) * 100.0f);
+  std::string ss = fmt::format("Calculating Closest Vertices || {}%", progressInt);
+  m_MessageHandler(IFilter::Message::Type::Info, ss);
+
+  m_LastProgressInt = progressInt;
+  m_InitialPoint = std::chrono::steady_clock::now();
+}
+
+// -----------------------------------------------------------------------------
+void ThreadedAlgorithm::sendThreadSafeInfoMessage(const std::string& message)
```
nyoungbq:
```suggestion
void ThreadedAlgorithm::sendThreadSafeInfoMessage(std::string&& message)
```

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  const IFilter::MessageHandler& getMessageHandler() const;
+
+  void setTotalElements(size_t totalElements);
```
nyoungbq: There's nothing wrong with this method in particular, but its visibility is concerning since it modifies the object without a lock. This should never be called in parallel, but in its current state it can be done.

To get around this we could make this an arg that must be passed to the ctor.

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace nx::core
+{
+
+class SIMPLNX_EXPORT ThreadedAlgorithm
+{
+
+public:
+  ThreadedAlgorithm(const IFilter::MessageHandler& mesgHandler, const std::atomic_bool& shouldCancel);
+  virtual ~ThreadedAlgorithm() noexcept;
```
nyoungbq: Why is this dtor marked as virtual?

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include "simplnx/Filter/IFilter.hpp"
+#include "simplnx/simplnx_export.hpp"
+
+#include <chrono>
+#include <cstdint>
+
+namespace nx::core
+{
+
+class SIMPLNX_EXPORT ThreadedAlgorithm
```
nyoungbq: Integrating the time counts into this is a really bad idea. I think the threads calling this should be responsible for determining when to update. The check for 1000 milliseconds is an expensive one because of all the casts and it must be locked for that entire duration. This will significantly slow down output, rather than just locking to calculate and assemble the string output. By outsourcing the time cast checks to the threads at least we leverage the parallel nature. That said I will have to verify that the chrono library is actually threadsafe.

Edit: steady_clock is threadsafe: https://stackoverflow.com/questions/10894504/is-there-any-stdchrono-thread-safety-guarantee-even-with-multicore-context

src/simplnx/Utilities/ThreadedAlgorithm.cpp
JDuffeyBQ: In the case of `std::atomic<T>` the `operator T()` is equivalent to calling `load()`. See https://en.cppreference.com/w/cpp/atomic/atomic/operator_T. Also I would say that the cancel variable doesn't need to be in this class anyways since the class is intended to support the messaging aspect.

src/simplnx/Utilities/ThreadedAlgorithm.hpp
JDuffeyBQ: The `std::string` will be copied into `IFilter::Message` so no changes will propagate. Additionally in the current implementation control flow will invoke the `MessageHandler::Callback` with no delay on that thread so there is no opportunity for the value to change before being used. If we eventually change the implementation we may need to be more explicit about the lifetime/copying.

src/simplnx/Utilities/ThreadedAlgorithm.hpp
JDuffeyBQ: The current implementation has algorithm classes inheriting from `ThreadedAlgorithm` since that keeps the same API has the algorithm classes that already had a version of the thread safe message function. This is also why the `setTotalElements()` function exists since the number of elements is calculated in the call operator not the constructor. In a future implementation, we could move away from inheritance.

src/simplnx/Utilities/ThreadedAlgorithm.hpp
JDuffeyBQ: This implementation is the same as what some filters were already doing. I'd say that for this PR getting more filters to output and unifying the method of output is the goal. Then it should be easier to update the implementation to something with improved performance later.

nyoungbq (approved):

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/BadDataNeighborOrientationCheck.cpp
```diff
@@ -81,16 +74,12 @@ Result<> BadDataNeighborOrientationCheck::operator()()
   std::vector<int32_t> neighborCount(totalPoints, 0);
 
   int64_t progressInt = 0;
-  auto start = std::chrono::steady_clock::now();
+
   for(size_t i = 0; i < totalPoints; i++)
   {
-    auto now = std::chrono::steady_clock::now();
-    if(std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() > 1000)
+    if(shouldSend())
```
JDuffeyBQ: `sendMessage` calls `shouldSend()` already so you don't need this extra check. If you want to avoid having to call `fmt::format()` if a message isn't sent, then you can use the overload that takes a functor: e.g. `sendMessage([progressInt](){ return fmt::format("Processing Data '{}'% completed", progressInt); })`

This isn't the only spot for this as well.

src/simplnx/Utilities/ThreadedAlgorithm.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * @brief
+ */
+class SIMPLNX_EXPORT FilterAlgorithm
+{
+public:
+  FilterAlgorithm(const IFilter::MessageHandler& messageHandler, const std::atomic_bool& shouldCancel);
+
+  virtual ~FilterAlgorithm() noexcept = default;
```
JDuffeyBQ: We don't actually have to mark the destructors as `virtual` because these classes aren't intended to be used polymorphically. Not a big deal either way though.

JDuffeyBQ (changes requested):

src/simplnx/Utilities/ThreadedAlgorithm.hpp
imikejackson: The subclasses implement the destructor, even though they are "default" implementations, so core guidelines say that you should mark them override. And when you mark them override, then it complains that the base class isn't marked virtual. Thoughts?

JDuffeyBQ: Do you have the specific guideline that comes up?

imikejackson: https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-override.html

JDuffeyBQ: Is it just this one or are there two?

imikejackson:
@JDuffeyBQ 

I started profiling this branch and I'm not sure we want this design. Running a specific pipeline (that I got from a user) I was spending 87% of the time just doing `auto now = std::chrono::steady_clock::now();`. The time for the filter was around 3 minutes. When I removed this kind of check it went down to 30 seconds. Release mode. 

It is like we need a whole separate thread that just wakes up every second and looks for a message that has been added to it. It sends that message off with some sort of deferred execution of the fmt::format() and goes back to sleep.

imikejackson:
@JDuffeyBQ @nyoungbq 

The below is pretty rough and also generated by our favorite AI. I tried using std::jthread, which is C++20, but I guess Apple clang isn't going to support it at all. https://en.cppreference.com/w/cpp/compiler_support#cpp20

```
class StringChecker
{
public:
  explicit StringChecker(const IFilter::MessageHandler& messageHandler)
  : m_MessageHandler(messageHandler)
  , m_StopRequested(false)
  {
    m_Thread = std::thread(&StringChecker::run, this);
  }

  ~StringChecker()
  {
    {
      std::lock_guard lock(m_Mutex);
      m_StopRequested = true;
    }
    m_CV.notify_all(); // Wake up the thread if it's sleeping
    if(m_Thread.joinable())
    {
      m_Thread.join(); // Wait for the thread to finish
    }
  }

  // Sets the value in a thread-safe way
  void setValue(std::string value)
  {
    {
      std::lock_guard lock(m_Mutex);
      m_Value = std::move(value);
    }
  }

private:
  void run()
  {
    while(true)
    {
      std::unique_lock lock(m_Mutex);
      m_CV.wait_for(lock, std::chrono::seconds(1), [this] { return m_StopRequested || !m_Value.empty(); });

      if(m_StopRequested)
        break;

      if(!m_Value.empty())
      {
        std::string temp = std::move(m_Value);
        m_Value.clear(); // Reset before unlocking
        lock.unlock();
        m_MessageHandler({nx::core::IFilter::Message::Type::Info, temp});
      }
    }
  }

  std::string m_Value;
  std::mutex m_Mutex;
  std::condition_variable m_CV;

  std::atomic_bool m_StopRequested;
  std::thread m_Thread;

  const IFilter::MessageHandler& m_MessageHandler;
};

```
I profiled this and it barely took any extra time at all. In fact it didn't really show up on the profile trace in release mode.

nyoungbq:
> @JDuffeyBQ
> 
> I started profiling this branch and I'm not sure we want this design. Running a specific pipeline (that I got from a user) I was spending 87% of the time just doing `auto now = std::chrono::steady_clock::now();`. The time for the filter was around 3 minutes. When I removed this kind of check it went down to 30 seconds. Release mode.
> 
> It is like we need a whole separate thread that just wakes up every second and looks for a message that has been added to it. It sends that message off with some sort of deferred execution of the fmt::format() and goes back to sleep.

@imikejackson This is exactly what me and Jared were discussing in the Slack group. The most efficient way to handle messaging is with a thread that just manages messaging with a lock free message queue. There are even libraries that do this, (and would be able to do logging) that we could integrate such as [g3log](https://github.com/KjellKod/g3log), but the consensus was that it was outside the scope of this PR. As for the locking cost, I mentioned this in my feedback:

> Integrating the time counts into this is a really bad idea. I think the threads calling this should be responsible for determining 
> when to update. The check for 1000 milliseconds is an expensive one because of all the casts and it must be locked for that 
> entire duration. This will significantly slow down output, rather than just locking to calculate and assemble the string output. 
> By outsourcing the time cast checks to the threads at least we leverage the parallel nature.

That said I think that the implementation you provided for putting it on a separate thread is a good change. It still uses locks, but they are clearly being hit much less often (likely due to the interactive method `setValue()` being far less computationally complex) which is a big boon for performance. Offloading thread waking with time counts to the scheduler will always be much faster than the cast checks we run, so that is also cutting out several expensive casts. My vote is Yes for integrating the `StringChecker` implementation and looking to make a lock free implementation or library integration on the todo list.

imikejackson:
I am not wondering how much we are handicapping some of the filters where we do the check for 1000 milliseconds. 

@nyoungbq @JDuffeyBQ I'm tempted to just close this PR and start over although the cancel checking is minimally needed.

nyoungbq:
> I am not wondering how much we are handicapping some of the filters where we do the check for 1000 milliseconds.
> 
> I'm tempted to just close this PR and start over although the cancel checking is minimally needed.

@imikejackson Yes, we are definitely losing performance by doing the check, that said not having feedback is not an option so doing the check inside the threads and not in a serial place is the best solution without having it on a separate thread for the time being IMO. We can mitigate the cost by:
- putting cheaper prerequisite conditional with an `&&` (since lazy evaluation will lower cost) 
- ensuring the part that needs a lock as least complex as possible (ideally a wrapper for output with lock)
- making the threads do as much of the work as possible before making the serial call

This makes it hard to centralize it, but its the best way to cut down on cost.

As far as taking out the parallel messaging and just doing the cancel, that's up to you. I think you can put in just parallel feedback on the filters that need it and move the centralization/ separate thread messaging to a new branch, that way not all of it goes to waste and we have a temporary feedback solution until we can properly address an ideal solution

nyoungbq (approved):

JDuffeyBQ (approved):

PR #1286, opened 2025-05-01
FILTER: Sample Scan Vectors

joeykleingers:


imikejackson (changes requested):
@joeykleingers  Being a bit late to the game I think I would have liked to see this filter be able to take _any_ data from the edge geometry and copy/create it on the vertex geometry.

- [x] I think the filter should be called "Sample Edge Geometry"
- [x] I think we should allow the user to copy any Edge Array, not just "Power and Layer". I see this filter having uses outside of our current customers.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/PointSampleEdgeGeometryFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+  inputs.CalculateCumulativeSampleDistance = filterArgs.value<bool>(k_CalculateCumulativeSampleDistance_Key);
+  inputs.pSelectedDataArrayPaths = filterArgs.value<MultiArraySelectionParameter::ValueType>(k_SelectedDataArrayPaths_Key);
+  auto pVertexGroupDataName = filterArgs.value<std::string>(k_VertexDataGroupName_Key);
+  inputs.pVertexGroupDataPath = inputs.ScanVectorGeometryPath.createChildPath(pVertexGroupDataName);
+
+  MultiArraySelectionParameter::ValueType createdDataPaths;
+  for(const auto& selectedDataPath : inputs.pSelectedDataArrayPaths)
+  {
+    DataPath createdDataPath = inputs.pVertexGroupDataPath.createChildPath(selectedDataPath.getTargetName());
+    createdDataPaths.push_back(createdDataPath);
```
nyoungbq:
```suggestion
  createdDataPaths.push_back(inputs.pVertexGroupDataPath.createChildPath(selectedDataPath.getTargetName()));
```
or
```suggestion
    DataPath createdDataPath = inputs.pVertexGroupDataPath.createChildPath(selectedDataPath.getTargetName());
    createdDataPaths.push_back(std::move(createdDataPath));
```

nyoungbq (approved):

nyoungbq (approved):

nyoungbq (approved):

PR #1288, opened 2025-05-02
ENH: Compute Array Statistics - Range Gating and Histogram Migration

### LINKED ISSUE #1191: Compute Array Statistics Enhancemets
- [ ] Add option to also output an array that has the Feature id in it.

- [ ] Do not fill in from zero to min feature id. That could go bad if someone has feature Ids that start at 1 billion.

- [ ] The created group needs to be an Attribute Matrix and not a Group. This will break a unit test because computing the histogram for an array does not produce a matching tuple and component shape.

- [ ] Option to set the "Feature Id" range.
This would allow a user to set a range of feature ids to calculate statistics for.
Allows the user to "pad out the feature ids" to a specific range

- [ ] Option to Ignore Feature Id Zero.
- [ ] The creation of the attribute Matrix will pad from zero to the minimum feature id that is encountered in the Feature ID Array. This could have interesting side effects of blowing up the Attribute Matrix if the minimum Feature Id is something like 2 billion

### LINKED ISSUE #1270: Compute Array Statistics Updates/Changes
- [x] Add option to also output an array that has the Feature id in it.
- [x] The created group needs to be an Attribute Matrix and not a Group
    - I wonder if there was a technical reason for this?
    - Yes there is: The Histogram computation generates an array with the number of tuples = number of bins.
- [x] Option to set the "Feature Id" range.
    - Allows the user to "pad out the feature ids" to a specific range
    - Allows the user to only compute stats for specific feature Ids.
- [x] Option to Ignore Feature Id Zero.
- [x] Do not fill in from zero to min feature id. That could go bad if someone has feature Ids that start at 1 billion.
    - The creation of the attribute Matrix will pad from zero to the minimum feature id that is encountered in the Feature ID Array. This could have interesting side effects of blowing up the Attribute Matrix if the minimum Feature Id is something like 2 billion

I started on some of these items on the imikejackson/simplnx/topic/compute_array_statistics_fixes branch.

There probably needs to be a longer discussion on some of these issues as changes to one (like the creation of the Attribute Matrix) will stop others from working (Histogram).

@nyoungbq is the primary
@imikejackson to provide support
@joeykleingers backup

nyoungbq:
Waiting on @joeykleingers updates for reintegrating histogram

Fixes: #1270 

Added all feature related updates, documentation, and 5 new tests cases for new functionality.
Removed Histogram functionality.
Changed DataGroup output to Attribute Matrix. 
Extensive cleanup to algorithm class code.

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


joeykleingers:
Histogram feature has been migrated to ComputeArrayHistogram and ComputeArrayHistogramByFeature filters.

src/simplnx/DataStructure/IDataStore.hpp
```diff
@@ -83,6 +83,12 @@ class SIMPLNX_EXPORT IDataStore
     return getSize();
   }
 
+  /**
+   * @brief Returns if there are any elements in the array object
+   * @return bool, true if the DataArray has a size() == 0
+   */
+  virtual bool empty() const = 0;
```
JDuffeyBQ: We have access to `getNumberOfTuples()` here so we could just implement the function here instead of overriding in `AbstractDataStore`.

JDuffeyBQ (approved):

PR #1290, opened 2025-05-02
ENH: Update conda and vcpkg to use EbsdLib 1.0.38

imikejackson:


conda/meta.yaml
```diff
@@ -74,5 +75,5 @@ test:
 
 about:
   home: https://www.dream3d.io/
-  license: BSD
+  license: Aferro GPL
```
JDuffeyBQ:
```suggestion
  license: Affero GPL
```

JDuffeyBQ (changes requested):

PR #1292, opened 2025-05-02
ENH: New design for sending filter progress messages at a reasonable rate

imikejackson:
This is rough starting design. There are several directions that could be taken. Hierarchy based, stand alone? This is just a few ideas and a few examples of 2 of the use cases:

Single Threaded: Just send a message once a second
MultiThreaded: The individual threads send back how many items they have completed. The Filter Algorithm class determines when to send the actual message.


imikejackson:
work is being finished at https://github.com/BlueQuartzSoftware/simplnx/pull/1340

PR #1305, opened 2025-05-20
ENH: Clarify Kernel Average Misorientation algorithm with proper casting.

imikejackson:
Added proper casting as the previous code was relying on the compiler to do the proper casting at the proper time. This explicitly sets the types and adds defensive checks to ensure the calculated index is >= 0.

Updated the documentation to state that the KAM is only calculated within a grain.

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
@@ -96,26 +96,27 @@ public:
             uint32_t phase1 = crystalStructures[cellPhases[point]];
             for(int32_t j = -kernelSize[2]; j < kernelSize[2] + 1; j++)
             {
-              size_t jStride = j * xPoints * yPoints;
+
               if(plane + j < 0 || plane + j > zPoints - 1)
               {
                 continue;
               }
+              const int64_t jStride = j * xPoints * yPoints;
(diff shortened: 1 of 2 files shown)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
```diff
@@ -96,34 +96,35 @@ class FindKernelAvgMisorientationsImpl
...
                 for(int32_t l = -kernelSize[0]; l < kernelSize[0] + 1; l++)
                 {
                   if(col + l < 0 || col + l > xPoints - 1)
                   {
                     continue;
                   }
-                  const size_t neighbor = point + (jStride) + (kStride) + (l);
-                  if(featureIds[point] == featureIds[neighbor])
+                  const int64_t neighbor = static_cast<int64_t>(point) + jStride + kStride + l;
+                  if(neighbor >= 0 && featureIds[point] == featureIds[static_cast<size_t>(neighbor)])
```
JDuffeyBQ, line 119: My only comment is whether we should report an error if `neighbor` is less than 0.

imikejackson: I don't think so. If the neighbor is 0 then that means that we ended up grabbing a voxel that we should not be using anyways so ignoring it is fine for the calculation.

It might be interesting for us as developers but it would not add any useful information to report it to the user.

JDuffeyBQ (approved):

PR #1306, opened 2025-05-20
ENH: Add RandomizeFeatureIds utility and filter

### LINKED ISSUE #1179: Generic "FeatureId" randomization
Create a new filter which will randomize "Feature Ids" arrays (user defined) and also update any attached "Feature Attribute Matrix".

While some segmentation filters have this, other filers do not or if the user imports segmented data. This is here primarily for visualization so when visualizing data the user would not "see" a gradient along certain directions.

mmarineBlueQuartz:
* Fix DataStoreUtilities.hpp not included in CmakeLists.txt
* Add RandomizeFeatureIds utility function to nx::core.
* Updated existing randomize feature IDs calls to use the utility function.
* Add RandomizeFeatureIdsFilter to SimplnxCore

fixes #1179

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/RandomizeFeatureIdsFilter.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/RandomizeFeatureIdsFilter.cpp
@@ -73,14 +73,6 @@ IFilter::UniquePointer RandomizeFeatureIdsFilter::clone() const
 IFilter::PreflightResult RandomizeFeatureIdsFilter::preflightImpl(const DataStructure& dataStructure, const Arguments& filterArgs, const MessageHandler& messageHandler,
                                                                   const std::atomic_bool& shouldCancel, const ExecutionContext& executionContext) const
 {
-  auto featureIdsPath = filterArgs.value<DataPath>(k_FeatureIds_Key);
-
-  const Int32Array* featureIdsArray = dataStructure.getDataAs<Int32Array>(featureIdsPath);
-  if(featureIdsArray == nullptr)
-  {
-    return MakePreflightErrorResult(-72601, fmt::format("The Feature IDs array was not found at path '{}'", featureIdsPath.toString()));
-  }
-
   return {};
 }
 
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/RandomizeFeatureIdsFilter.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/RandomizeFeatureIdsFilter.cpp
@@ -90,8 +90,8 @@ Result<> RandomizeFeatureIdsFilter::executeImpl(DataStructure& dataStructure, co
 {
   auto featureIdsPath = args.value<DataPath>(k_FeatureIds_Key);
 
-  Int32Array* featureIdsArray = dataStructure.getDataAs<Int32Array>(featureIdsPath);
-  auto& featureIdsStore = featureIdsArray->getDataStoreRef();
+  Int32Array& featureIdsArray = dataStructure.getDataRefAs<Int32Array>(featureIdsPath);
+  auto& featureIdsStore = featureIdsArray.getDataStoreRef();
   usize totalFeatures = *std::max_element(featureIdsStore.begin(), featureIdsStore.end());
 
   ClusterUtilities::RandomizeFeatureIds(featureIdsStore, totalFeatures);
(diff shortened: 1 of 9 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/RandomizeFeatureIdsFilter.cpp
```diff
@@ -0,0 +1,101 @@
...
+IFilter::PreflightResult RandomizeFeatureIdsFilter::preflightImpl(const DataStructure& dataStructure, const Arguments& filterArgs, const MessageHandler& messageHandler,
+                                                                  const std::atomic_bool& shouldCancel, const ExecutionContext& executionContext) const
+{
+  auto featureIdsPath = filterArgs.value<DataPath>(k_FeatureIds_Key);
+
+  const Int32Array* featureIdsArray = dataStructure.getDataAs<Int32Array>(featureIdsPath);
+  if(featureIdsArray == nullptr)
+  {
+    return MakePreflightErrorResult(-72601, fmt::format("The Feature IDs array was not found at path '{}'", featureIdsPath.toString()));
+  }
```
JDuffeyBQ: This check is already covered by the parameter.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/RandomizeFeatureIdsFilter.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+//------------------------------------------------------------------------------
+Result<> RandomizeFeatureIdsFilter::executeImpl(DataStructure& dataStructure, const Arguments& args, const PipelineFilter* pipelineNode, const MessageHandler& messageHandler,
+                                                const std::atomic_bool& shouldCancel, const ExecutionContext& executionContext) const
+{
+  auto featureIdsPath = args.value<DataPath>(k_FeatureIds_Key);
+
+  Int32Array* featureIdsArray = dataStructure.getDataAs<Int32Array>(featureIdsPath);
+  auto& featureIdsStore = featureIdsArray->getDataStoreRef();
```
JDuffeyBQ:
```suggestion
  Int32Array& featureIdsArray = dataStructure.getDataRefAs<Int32Array>(featureIdsPath);
  auto& featureIdsStore = featureIdsArray.getDataStoreRef();
```

src/simplnx/Utilities/ClusteringUtilities.cpp
```diff
@@ -0,0 +1,44 @@
...
+      continue;
+    }
+
+    int32 randId_i = randomIds[i];
+    randomIds[i] = randomIds[r];
+    randomIds[r] = randId_i;
+  }
+
+  // Now adjust all the Grain ID values for each Voxel
+  // instead of taking total points as an input just extract the size, so we don't walk of
```
JDuffeyBQ:
```suggestion
  // instead of taking total points as an input just extract the size, so we don't walk off
```

src/simplnx/Utilities/ClusteringUtilities.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+  const usize rangeMin = 1;
+  const usize rangeMax = totalFeatures - 1;
+  auto gen = std::mt19937_64(std::mt19937_64::default_seed);
+  std::uniform_real_distribution<float64> dist(0, 1);
+
+  IDataStore::ShapeType tupleShape{totalFeatures};
+  IDataStore::ShapeType componentShape{1};
+  std::shared_ptr<AbstractDataStore<int32>> randomIdsPtr = nx::core::DataStoreUtilities::CreateDataStore<int32>(tupleShape, componentShape, IDataAction::Mode::Execute);
+  Int32AbstractDataStore& randomIds = *randomIdsPtr.get();
```
JDuffeyBQ, line 19: The previous implementation used `std::vector`. Since the data isn't exposed outside of this function I think `std::vector` is still probably the best choice.

JDuffeyBQ (changes requested):

src/simplnx/Utilities/ClusteringUtilities.cpp
mmarineBlueQuartz: Not if the feature IDs array is out of core. If the array is larger than the total memory, the size of the created vector could be larger than the available RAM. `std::vector` is only safe for in-memory DataArrays.

JDuffeyBQ: Ah. Can `DataStoreUtilities::CreateDataStore` create out of core arrays?

mmarineBlueQuartz: It can and checks the memory requirement to see if it should unless specifically told to do so.

imikejackson (approved):

imikejackson (approved):

JDuffeyBQ (approved):

PR #1307, opened 2025-05-23
ENH: ComputeFeatureShapesTriangleGeom algorithm parallelization

imikejackson:
Parallelizes the ComputeFeatureShapesTriangleGeom algorithm class.

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeShapesTriangleGeom.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeShapesTriangleGeom.cpp
@@ -6,6 +6,7 @@
 #include "simplnx/DataStructure/Geometry/TriangleGeom.hpp"
 #include "simplnx/Utilities/GeometryHelpers.hpp"
 #include "simplnx/Utilities/IntersectionUtilities.hpp"
+#include "simplnx/Utilities/ParallelDataAlgorithm.hpp"
 
 #include "EbsdLib/Core/Orientation.hpp"
 #include "EbsdLib/Core/OrientationTransformation.hpp"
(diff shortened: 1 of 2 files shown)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeShapesTriangleGeom.cpp
```diff
@@ -411,49 +386,136 @@ Result<> ComputeShapesTriangleGeom::operator()()
...
+  axisEulerAnglesRef[3 * featureId] = static_cast<float32>(axisEulerAngles[0]);
+  axisEulerAnglesRef[(3 * featureId) + 1] = static_cast<float32>(axisEulerAngles[1]);
+  axisEulerAnglesRef[(3 * featureId) + 2] = static_cast<float32>(axisEulerAngles[2]);
+
+  axisLengthsRef[3 * featureId] = axisLengths[0];
+  axisLengthsRef[(3 * featureId) + 1] = axisLengths[1];
+  axisLengthsRef[(3 * featureId) + 2] = axisLengths[2];
+
+  aspectRatiosRef[2 * featureId] = aspectRatios[0];
+  aspectRatiosRef[(2 * featureId) + 1] = aspectRatios[1];
```
nyoungbq: This shouldn't have to be in a lock since you are explicitly avoiding data races by updating on feature index and they are never read from in the loop either. I would move these ahead of the lock to help cut down on lock time or even move it back into the end of the parallel feature loop and just call this function for updates.

imikejackson: I don't want to write to one of our DataStores from different threads because I am fairly sure that would cause issues especially considering out-of-core.

nyoungbq: > I don't want to write to one of our DataStores from different threads because I am fairly sure that would cause issues especially considering out-of-core.

@imikejackson The out-of-core should be okay as long as the get/set methods are called (instead of bracket operator) as matthew suggested. If the concern is chunk/cache flushing/miss rates in multithreading, I think we should test it and potentially implement a custom access pattern like this for if out of core is enabled. The locks will cripple performance and needlessly so for regular computation.

imikejackson: @nyoungbq I reworked the algorithm a bit. Now I collect all results for each thread, then pass that group of results down to the filter to be written into the final arrays. This should significantly lessen the number of times that the mutex_lock has to be acquired for each thread.

nyoungbq: I will approve the changes since the lock detriment is lower now, but I still believe that updating the datastores in parallel with get and set is safe to do so. I will ask at meeting to clarify my understanding of out of core handling, since if out of core is not parallel safe there will need to be widespread algorithm reworks when we do the July fixes for OoC. This is good to go in since our progress feedback would need the locks anyway in the current iteration, but when the new lock free messaging comes in this should definitely be revisited.

Attached is a profiler run of this new commit, in it the locks cost as much as the algorithm calculations themselves. The second column from the left is the actual algorithm call and the futex column (4th column from the left) is the wait time from locks. I think their weight in the burn graph will be slightly reduced on a larger dataset, but due to the partitioner tending towards smaller sections this will likely still be a significant cost. 

image

nyoungbq (approved):

PR #1314, opened 2025-05-29
Added NeighborList python bindings

JDuffeyBQ:
Documentation needs to be updated


imikejackson (changes requested):
Could we add some more to the example script that shows looping over a neighbor list. I know it sounds basic but always nice to have a copy/paste example.

imikejackson (approved):

PR #1319, opened 2025-06-06
FILTER: Compute Coordinate Threshold

### LINKED ISSUE #1208: FILTER: Set Array Values based on Physical Location
I would like a filter for Image Geometries that allows the user to set values of a cell based on the cell's physical location. For example the filter could take in a lower and upper XYZ corner and everything INSIDE or OUTSIDE of that bounding box would have a specific value set (set by the user).

In theory we could do this for any geometry so we should design the first iteration with that in mind. I think tackling the ImageGeometry first is a great first step to add this functionality.

nyoungbq:
Fixes: #1208

This filter offers the ability to produce a mask for cells that fall within a given bounding geometric shape for most geometries. See documentation provided for further details.

Discussion was had to expand the bounding definitions down the line, so the filter was designed around executing lambdas that determine whether a point is in or out of bounds. These lambdas are run by a parallel wrapper class that handles parsing the different geometry types to pass an xyz point to the lambda. The idea being all that would be necessary to add new functionality would be:

- [ ] Updating input parameters/args/preflight for the filter
- [ ] Updating the `BoundsType` enum in the algorithm header
- [ ] Defining a new lambda to be passed to the parallel algorithm in the switch inside the `operator()` function of the algorithm class
- [ ] Adding 5 new test cases; one for each supported geometry (with 2 sections; one regular one inverted)

This PR adds the following functionality:
- New filter providing a way to mask off cells with supplied bounds in a variety of geometries
- Extensive supporting documentation
- 20 new test cases testing every algorithm code path (10 unique, 2 sections in each)

### DIFF
--- a/src/Plugins/SimplnxCore/CMakeLists.txt
+++ b/src/Plugins/SimplnxCore/CMakeLists.txt
@@ -26,6 +26,7 @@ set(FilterList
   ComputeBoundaryCellsFilter
   ComputeBoundaryElementFractionsFilter
   ComputeCoordinatesImageGeomFilter
+  ComputeCoordinateThresholdFilter
   ComputeDifferencesMapFilter
   ComputeEuclideanDistMapFilter
   ComputeFeatureBoundsFilter
@@ -174,6 +175,7 @@ set(AlgorithmList
   ComputeBiasedFeatures
   ComputeBoundaryCells
   ComputeCoordinatesImageGeom
+  ComputeCoordinateThreshold
   ComputeEuclideanDistMap
   ComputeFeatureBounds
   ComputeFeatureCentroids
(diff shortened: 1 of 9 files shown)


imikejackson (changes requested):
- We should check to make sure at least 1 cell was flagged true or false. This way if they set a threshold range that does not include the geometry we could emit a warning for that during pipeline execution
- I'm toiling with the idea of at least sanity checking ImageGeometries to make sure the user is actually going to create mask, see above first point. with Image Geometry, we know everything to check that condition up front. With the other geometries we can't know that information.

nyoungbq:
Change image geom to work max and min points defining a voxel instead of a centroid

imikejackson (approved):

PR #1320, opened 2025-06-09
BUG: RequireMinNumNeighbors DataArray Update fixes

imikejackson:
- [x] Remove all debugging code

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinNumNeighbors.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinNumNeighbors.cpp
@@ -8,6 +8,39 @@
 
 using namespace nx::core;
 
+namespace
+{
+
+struct UpdateDataFunctor
+{
+  template <typename T>
+  Result<> operator()(const DataStructure& dataStructure, const DataPath& dataArrayPath, const std::vector<usize>& badFeatureIdIndexes, const AbstractDataStore<int32_t>& featureIds,
+                      const std::vector<int32>& neighbors, const IFilter::MessageHandler& mesgHandler)
+  {
+    auto voxelArray = dataStructure.getDataRefAs<DataArray<T>>(dataArrayPath);
+    auto arraySize = voxelArray.getSize();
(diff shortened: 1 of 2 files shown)


JDuffeyBQ:
What is the bug that this fixes?

imikejackson:
I'm still trying to clean up the code.. .but... 

The issue is that the DataArrays are stored in alphabetical order in the std::vector<DataPath> so what happens is that if a data array has name that lexigraphically comes before the "FeatureIds" array then it will get updated correctly, if it comes after "FeatureIds" then the data array never gets updated because the "FeatureIds" array just got updated and so none of the conditions to perform the updates ever get hit (FeatureId < 0). The same issue exists in DREAM3D version 6.5 and 6.6 which is why this does not show up in the unit test

I actually need to come up with a proper testing data set.

I have a Merge Request for DREAM3D on our GitLab server that will fix the issue there also.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinNumNeighbors.cpp
```diff
@@ -292,38 +307,28 @@ Result<> RequireMinNumNeighbors::operator()()
...
-          m_MessageHandler(nx::core::IFilter::Message{nx::core::IFilter::Message::Type::Info, message});
-          return MakeErrorResult(-55568, message);
-        }
-
-        if(featureName < 0 && neighbor >= 0 && featureIds[neighbor] >= 0)
-        {
-          voxelArray->copyTuple(neighbor, featureIdIndex);
-        }
-      }
+      ExecuteDataFunction(UpdateDataFunctor{}, voxelArray->getDataType(), m_DataStructure, cellArrayPath, badFeatureIdIndexes, featureIds, neighbors, m_MessageHandler);
```
JDuffeyBQ: Was there a reason for changing this to a template function?

imikejackson: I guess we don't technically need it to be a templated function.

JDuffeyBQ (approved):

PR #1335, opened 2025-06-19
[WIP] BUG: Created arrays should have the same tuple shape as parent attribute matrix

### LINKED ISSUE #1334: BUG: Created Arrays do not inherit tuple dimensions from parent Attribute Matrix
Some filters do not correctly set the output array's tuple shape but instead just use the total number of tuples.

There are probably more filters that have this same kind of issue. They need to be found.

This issue can be broken into a PR for each of the Plugins

## SIMPLNX Core Plugin

- [ ] Array Calculator Filter
- [ ] Compute CoordinateImageGeomFilter

imikejackson:
This is for issue https://github.com/BlueQuartzSoftware/simplnx/issues/1334


nyoungbq (approved):

imikejackson:
@joeykleingers is going to work this on another branch.

PR #1340, opened 2025-06-23
ENH: Improved threadsafe messaging

JDuffeyBQ:
Switched to spdlog for processing messages from filters on a separate thread

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
@@ -70,10 +70,7 @@ public:
 
       if(counter > increment)
       {
-        progressMessenger.sendProgressMessage(counter, [&](usize currentProgress, usize maxProgress) {
-          int32 percentComplete = CalculatePercentCompleteAsInt(currentProgress, maxProgress);
-          return fmt::format(fmt::runtime(progressMessenger.getProgressMessageData().m_MessageTemplate), percentComplete);
-        });
+        progressMessenger.sendProgressMessage(counter);
         counter = 0;
       }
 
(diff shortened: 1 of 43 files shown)


imikejackson:
```
ProgressMessenger progressMessenger = m_ProgressMessageHelper.createProgressMessenger();
progressMessenger.sendProgressMessage(counter);
```
Is that chunk thread safe? Because it is getting called from multiple threads.

src/simplnx/Utilities/MessageHelper.hpp
imikejackson, line 1: There is no documentation for this file which I think is the main class that we should be using?

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
```diff
@@ -68,13 +70,11 @@ class FindKernelAvgMisorientationsImpl
...
       if(counter > increment)
       {
-        auto now = std::chrono::steady_clock::now();
-        if(std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() > 1000)
-        {
-          m_Filter->sendThreadSafeProgressMessage(counter);
-          counter = 0;
-          start = std::chrono::steady_clock::now();
-        }
+        progressMessenger.sendProgressMessage(counter, [&](usize currentProgress, usize maxProgress) {
```
imikejackson: The original code funneled the messages through the filter class in order to obtain an accurate count of how much has been computed and to throttle how fast messages can be sent. This implementation does not retain that functionality. I now get a large number of messages to the output. And while these messages do not show up on a time trace, the shear number of messages is too much.

imikejackson (changes requested):

src/simplnx/Utilities/MessageHelper.hpp
JDuffeyBQ: Done.

JDuffeyBQ:
> ```
> ProgressMessenger progressMessenger = m_ProgressMessageHelper.createProgressMessenger();
> progressMessenger.sendProgressMessage(counter);
> ```
> 
> Is that chunk thread safe? Because it is getting called from multiple threads.

Yes. An atomic variable is used to store the total progress, and the actual messaging is thread safe using spdlog. And the other data stored inside ProgressMessenger is not shared so there is not issue.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeKernelAvgMisorientations.cpp
JDuffeyBQ: I should be able to reduce it. And for comparison, what filter and data were you using?

imikejackson: @JDuffeyBQ I Slacked the Google Drive link to you

JDuffeyBQ: The current version should be reduced now.

imikejackson (approved):

PR #1341, opened 2025-06-24
BUG: Fix Import DREAM3D Memory Usage

mmarineBlueQuartz:
Broke up importing DREAM3D files into two stages so that only DataObjects requested are fully imported. Other data is imported using empty data stores to serve as placeholders. Geometries will always import the corresponding arrays completely during preflight due to other filters using those arrays during their calculations.

Added ability to import HDF5 dataset as an AbstractDataStore with or without predetermined tuple and component shapes. This will use the appropriate type of data store based on the required memory size and user preferences. Reading as a vector would not work when using out-of-core data.

### DIFF
--- a/src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
+++ b/src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
@@ -331,7 +331,7 @@ TEST_CASE("DREAM3DFileTest:Import/Export DREAM3D Filter Test")
     REQUIRE(importDataStructure.getData(DataPath({DataNames::k_Group1Name})) != nullptr);
     auto* dataArray = importDataStructure.getDataAs<DataArray<int8>>(DataPath({DataNames::k_ArrayName}));
     REQUIRE(dataArray != nullptr);
-    REQUIRE(dataArray->template getIDataStoreAs<EmptyDataStore<int8>>() == nullptr);
+    REQUIRE(dataArray->template getIDataStoreAs<EmptyDataStore<int8>>() != nullptr);
   }
 }
 
--- a/src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
+++ b/src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
@@ -331,7 +331,7 @@ TEST_CASE("DREAM3DFileTest:Import/Export DREAM3D Filter Test")
     REQUIRE(importDataStructure.getData(DataPath({DataNames::k_Group1Name})) != nullptr);
     auto* dataArray = importDataStructure.getDataAs<DataArray<int8>>(DataPath({DataNames::k_ArrayName}));
     REQUIRE(dataArray != nullptr);
-    REQUIRE(dataArray->template getIDataStoreAs<EmptyDataStore<int8>>() != nullptr);
+    REQUIRE(dataArray->template getIDataStoreAs<EmptyDataStore<int8>>() == nullptr);
   }
 }
 
--- a/src/simplnx/DataStructure/IO/HDF5/DataArrayIO.hpp
+++ b/src/simplnx/DataStructure/IO/HDF5/DataArrayIO.hpp
@@ -47,6 +47,82 @@ public:
     err = (data == nullptr) ? -400 : 0;
   }
 
+  template <typename K>
+  static Result<> importDataStore(data_type* dataArray, const DataPath& dataPath, const nx::core::HDF5::DatasetIO& datasetReader)
+  {
+    std::shared_ptr<AbstractDataStore<T>> dataStore = DataStoreIO::ReadDataStore<T>(datasetReader);
+    if(dataStore == nullptr)
+    {
+      return MakeErrorResult(-150202, fmt::format("Failed to import DataArray data at path '{}'.", dataPath.toString()));
+    }
+    dataArray->setDataStore(dataStore);
+    return {};
+  }
+
+  /**
+   * @brief Replaces the AbstractDataStore using data from the HDF5 dataset.
+   * @param dataStructure
+   * @param dataPath
+   * @param dataStructureReader
+   * @return Result<>
+   */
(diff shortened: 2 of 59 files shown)


imikejackson:
@mmarineBlueQuartz  Initial testing on a 75 GB .dream3d file allocated over 100GB of RAM just trying to add the "Read DREAm3D" filter to the pipeline. If you need this test file let me know and I can post it to Google Drive

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -331,7 +331,7 @@ TEST_CASE("DREAM3DFileTest:Import/Export DREAM3D Filter Test")
     REQUIRE(importDataStructure.getData(DataPath({DataNames::k_Group1Name})) != nullptr);
     auto* dataArray = importDataStructure.getDataAs<DataArray<int8>>(DataPath({DataNames::k_ArrayName}));
     REQUIRE(dataArray != nullptr);
-    REQUIRE(dataArray->template getIDataStoreAs<EmptyDataStore<int8>>() != nullptr);
+    REQUIRE(dataArray->template getIDataStoreAs<EmptyDataStore<int8>>() == nullptr);
```
JDuffeyBQ: This part of the test is checking the preflight data structure so shouldn't it contain an `EmptyDataStore`?

JDuffeyBQ: Testing in the GUI preflight is allocating the entire array.

mmarineBlueQuartz: The previous implementation of importing DREAM3D files did not have an option for creating EmptyDataStores during preflight.

src/simplnx/DataStructure/ListStore.hpp
```diff
@@ -258,7 +259,30 @@ class ListStore : public AbstractListStore<T>
     m_Array = lists;
   }
 
+  void readHdf5(const HDF5::DatasetIO& datasetReader) override
+  {
+    throw std::runtime_error("EmptyListStore cannot read from HDF5");
```
JDuffeyBQ:
```suggestion
    throw std::runtime_error("ListStore cannot read from HDF5");
```

src/simplnx/DataStructure/ListStore.hpp
```diff
@@ -258,7 +259,30 @@ class ListStore : public AbstractListStore<T>
...
+  }
+
+  void writeHdf5(HDF5::DatasetIO& datasetReader) override
+  {
+    throw std::runtime_error("EmptyListStore cannot write to HDF5");
```
JDuffeyBQ:
```suggestion
    throw std::runtime_error("ListStore cannot write to HDF5");
```

src/simplnx/DataStructure/IO/HDF5/DataArrayIO.hpp
```diff
@@ -47,6 +47,82 @@ class DataArrayIO : public IDataIO
...
+    if(dataArray == nullptr)
+    {
+      return MakeErrorResult(-150201, fmt::format("Imported DataStructure Object at path '{}' is not of the expected type.", dataPath.toString()));
+    }
+
+    auto datasetReader = parentGroupReader.openDataset(dataPath.getTargetName());
+    std::string dataTypeStr;
+    auto dataTypeStrResult = datasetReader.readStringAttribute(Constants::k_ObjectTypeTag);
+    dataTypeStr = std::move(dataTypeStrResult.value());
+    const bool isBoolArray = (dataTypeStr.compare("DataArray<bool>") == 0);
```
JDuffeyBQ:
```suggestion
    auto dataTypeStrResult = datasetReader.readStringAttribute(Constants::k_ObjectTypeTag);
    std::string dataTypeStr = std::move(dataTypeStrResult.value());
    const bool isBoolArray = dataTypeStr == "DataArray<bool>";
```

src/simplnx/Filter/Actions/ImportH5ObjectPathsAction.cpp
```diff
@@ -67,14 +67,16 @@ Result<> ImportH5ObjectPathsAction::apply(DataStructure& dataStructure, Mode mod
       importGroup->clear();
     }
 
-    if(dataStructure.getDataAs<DataObject>(targetPath) != nullptr)
+    if(!dataStructure.insert(importData, targetPath.getParent()))
     {
-      return MakeErrorResult(-6203, fmt::format("{}Unable to import DataObject at '{}' because an object already exists there. Consider a rename of existing object.", prefix, targetPath.toString()));
+      // return MakeErrorResult(k_InsertFailureError, fmt::format("{}Unable to import DataObject at '{}'", prefix, targetPath.toString()));
```
JDuffeyBQ: Should this be uncommented?

src/simplnx/Utilities/Parsing/DREAM3D/Dream3dIO.cpp
```diff
@@ -1517,6 +1997,70 @@ Result<nlohmann::json> DREAM3D::ImportPipelineJsonFromFile(const std::filesystem
...
+  const auto fileVersion = GetFileVersion(fileReader);
+  if(fileVersion == k_CurrentFileVersion)
+  {
+    return HDF5::DataStructureReader::FinishImportingObject(dataStructure, fileReader, dataPath);
+  }
+  else if(fileVersion == k_LegacyFileVersion)
+  {
+    auto dataStructureReader = fileReader.openGroup(k_LegacyDataStructureGroupTag);
+    // std::string parentPathStr = dataPath.getParent().toString();
+    // auto parentGroup = dataStructureReader.openGroup(dataPath.getParent().toString());
```
JDuffeyBQ:
```suggestion
```
Unused

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #1344, opened 2025-07-01
ENH: Allow existing geometry in RegularGridSampleSampleSurfaceMesh

### LINKED ISSUE #1311: ENH: RegularGridSampleSurfaceMeshFilter - allow use of existing image geometry
The filter RegularGridSampleSurfaceMeshFilter should allow the user to use an existing Image Geometry.

Maybe use a Choice Parameter to gather from the user if they have an existing geometry. That will inform the filter which other parameters also need to be shown.

If they are creating the Image Geomtery then you need to show the existing "Image Geometry", "Cell Attribute Matrix" and "Featuer Ids" parameters, otherwise, if they are bringing an Image Geometry, then you just need to show the "Feature Ids".

So the "Feature Ids" gets shown either way but in the base where they bring the Image Geometry, the output array needs to be placed into the "Cell Attribute Matrix" so you might have to entice that out of the geometry that is selected? Or you just end up with a set of 3 parameters for each case. Not sure which one is better.

JDuffeyBQ:
Resolves #1311


imikejackson (changes requested):
Need to update the documentation

imikejackson (approved):

PR #1347, opened 2025-07-01
ENH: Add 'swap' method to the various data array classes.

### LINKED ISSUE #1346: Add a 'swap' method to IDataArray/DataArray/NeighborList/StringArray classes
This will allow higher level manipulations inside a loop.

https://github.com/BlueQuartzSoftware/simplnx/pull/1347

imikejackson:


JDuffeyBQ (commented):
We could move the virtual function to `IArray` since `DataArray`, `StringArray`, and `NeighborList` all derive from it.

JDuffeyBQ (approved):

PR #1355, opened 2025-07-03
ENH: Add option to keep input geometry origin for RotateSampleRefFrame Filter

### LINKED ISSUE #1353: ENH: Rotate Sample Reference Frame add option to preserve origin
The rotation is a true rotation, but allowing the use to preserve the original origin would allow an effective axis swap of the data.

imikejackson:
* Also fix unit test so the test is actually testing the code
* Add another unit test section to test the new option.
* Documentation updated


nyoungbq (approved):
The code looks good, most of the changes outside the test seem straightforward. I noticed the PR purges a bunch of unused code from the Filter's anonymous namespace. I would add a description to the commit message/PR message that mentions removal of deprecated code, its deletion threw me off for a minute or two because it wasn't mentioned.

PR #1365, opened 2025-07-11
Fix Inconsistent Data Array and Attribute Matrix Tuple Shapes

### LINKED ISSUE #1334: BUG: Created Arrays do not inherit tuple dimensions from parent Attribute Matrix

joeykleingers:
Added method to UnitTestCommon that recursively searches a data structure for attribute matrices, then checks that all child data arrays have tuple shapes that match the attribute matrix tuple shape.  Inserted a call to this method at the bottom of every TEST_CASE that we have.  Several filters/tests were flagged and I fixed them.


joeykleingers:
Added CheckArraysInheritTupleDims method to UnitTestCommon that recursively searches a data structure for attribute matrices, then checks that all child data arrays have tuple shapes that match the attribute matrix tuple shape. Inserted a call to this method at the bottom of every TEST_CASE.  Fixed all tests that failed this check.

test/UnitTestCommon/include/simplnx/UnitTest/UnitTestCommon.hpp
```diff
@@ -1339,6 +1340,27 @@ inline DataStructure CreateComplexMultiLevelDataGraph()
...
+    REQUIRE(daPathsOpt.has_value());
+    auto daPaths = daPathsOpt.value();
+    //    REQUIRE(daPaths.size() > 0);
+    for(const auto& daPath : daPaths)
+    {
+      auto arr = dataStructure.getDataAs<IArray>(daPath);
+      REQUIRE(attrMatrix.getShape() == arr->getTupleShape());
+    }
+  }
+}
```
nyoungbq:
```suggestion
inline void CheckArraysInheritTupleDims(const DataStructure& dataStructure)
{
  std::optional<std::vector<DataPath>> amPathsOpt = GetAllChildDataPathsRecursive(dataStructure, {}, DataObject::Type::AttributeMatrix);
  REQUIRE(amPathsOpt.has_value());
  for(const auto& amPath : amPathsOpt.value())
  {
    const auto& attrMatrix = dataStructure.getDataRefAs<AttributeMatrix>(amPath);
    std::optional<std::vector<DataPath>> daPathsOpt = GetAllChildDataPaths(dataStructure, amPath, DataObject::Type::DataArray);
    REQUIRE(daPathsOpt.has_value());
    for(const auto& daPath : daPathsOpt.value())
    {
     const auto& arr = dataStructure.getDataAs<IArray>(daPath);
      REQUIRE(attrMatrix.getShape() == arr->getTupleShape());
    }
  }
}
```

nyoungbq: This change simplifies the code by extracting needless variables, better expressing the types of variables, and removing the unused code. Mostly stylistic.

Real quick I wanted to explain the methodology for the `auto` change. For best practice in C++, `auto` should only be used in cases where the type is already clearly expressed in the same function/block. Typical use cases include:
- Where the type is expressed in the assignment (think `getDataAs<>` where the value in the chevrons is the type; in these cases you should still explicitly specify modifiers and/or `&`/`*`) 
- In lambda variables where the typing is convoluted/redundant since lambdas typically have types laid out in definition
- In templates where the typing is derived implicitly
- In range for loops where the type of objects in container are already expressed elsewhere

In this case, you encoded the typing in the name with the `Opt` but the underlying types are still obscured, so it makes more sense to specify the types explicitly.

nyoungbq (approved):

nyoungbq (approved):

PR #1397, opened 2025-08-17
BUG: Fixes bug in BadDataNeighborOrientationCheck when finding the LaueOps instance

### LINKED ISSUE #1395: BUG: BadDataNeighborOrientationCheck crash
The variable `phase1` is shared between separate loops and if it's not set in the first loop it is invalid for the 2nd. This bug is not present in old DREAM3D. The solution is to separately acquire the value of `phase1` in each loop with no sharing.

imikejackson:
- Also brings consistency to all uses of the LaueOps classes to use 'laueClass' as a variable.
- Fixes bug where the phase value was indexing directly into orientation ops instead of going through crystal structures first in ComputeFeatureFaceMisorientation

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureFaceMisorientation.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureFaceMisorientation.cpp
@@ -76,7 +76,7 @@ public:
           quat2 = m_Quats[feature2 * 4 + 2];
           quat3 = m_Quats[feature2 * 4 + 3];
           QuatD q2(quat0, quat1, quat2, quat3);
-          OrientationD axisAngle = m_OrientationOps[phase1]->calculateMisorientation(q1, q2);
+          OrientationD axisAngle = m_OrientationOps[m_CrystalStructures[phase1]]->calculateMisorientation(q1, q2);
 
           m_Colors[3 * i + 0] = axisAngle[0] * (axisAngle[3] * Constants::k_180OverPiD);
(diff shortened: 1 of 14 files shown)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureFaceMisorientation.cpp
```diff
@@ -76,7 +76,7 @@ class CalculateFaceMisorientationColorsImpl
           quat2 = m_Quats[feature2 * 4 + 2];
           quat3 = m_Quats[feature2 * 4 + 3];
           QuatD q2(quat0, quat1, quat2, quat3);
-          OrientationD axisAngle = m_OrientationOps[phase1]->calculateMisorientation(q1, q2);
+          OrientationD axisAngle = m_OrientationOps[m_CrystalStructures[phase1]]->calculateMisorientation(q1, q2);
```
JDuffeyBQ, line 79: Is this also a bug fix?

imikejackson: yes

imikejackson: yes

JDuffeyBQ: I'll update the description to also mention this one.

JDuffeyBQ (approved):

PR #1409, opened 2025-08-27
ENH: AbstractDataStore now uses proxies instead of references

JDuffeyBQ:
`AbstractDataStore` now returns a proxy class for accessing elements instead of direct references to better support stores not in memory.


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgCAxes.cpp
```diff
@@ -145,11 +145,25 @@ Result<> ComputeAvgCAxes::operator()()
...
+      float32 a = avgCAxes.getValue(tupleIndex);
+      float32 b = avgCAxes.getValue(tupleIndex + 1);
+      float32 c = avgCAxes.getValue(tupleIndex + 2);
 
-      MatrixMath::Normalize3x1(avgCAxes[tupleIndex], avgCAxes[tupleIndex + 1], avgCAxes[tupleIndex + 2]);
+      a /= counter[i];
+      b /= counter[i];
+      c /= counter[i];
+
+      float32 denom = std::sqrt(((a * a) + (b * b) + (c * c)));
```
imikejackson: Why did we inline the MatrixMath::Normalize3x1() function here? Couldn't we just do
`MatrixMath::Normalize3x1(a, b, c);
   avgCAxes.setValue(tupleIndex, a);
      avgCAxes.setValue(tupleIndex + 1, b);
      avgCAxes.setValue(tupleIndex + 2, c);
`

imikejackson (changes requested):

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgCAxes.cpp
JDuffeyBQ: I think this was the result of two separate passes I did after some changes. With the current code I can change it back to `MatrixMath::Normalize3x1`.

imikejackson (approved):

PR #1417, opened 2025-09-03
MEM: Fix memory leaks and unnecessary memory allocations during preflight

imikejackson:


imikejackson:
closing this as https://github.com/BlueQuartzSoftware/simplnx/pull/1430 is a better solution

PR #1421, opened 2025-09-04
ENH: Rewrites the DBSCAN filter for performance and correctness

### LINKED ISSUE #1338: DBSCAN is very slow to execute
Current implementation is slow due to O(n^2) algorithm being used.

Updated implementation is needed.

nyoungbq:
Fixes: #1338

1.2 million 3D points in release in under 2 seconds. Benchmarks faster than Sci-Kit learn for 2D by a good amount.

The slowest part of the filter is now ClusterUtilities::GetDistance() - 60% of runtime, which is already fairly fast and is used though-out the code base.


imikejackson (changes requested):
Generally nicely done. I would like to see an example pipeline added including any needed data set.

imikejackson (approved):

PR #1429, opened 2025-09-09
ENH: Add function to DataArrayUtilities for finding the min & max for each component in a data array

jmarquisbq:


imikejackson (approved):

imikejackson (changes requested):
I think the return should be the same as the type. If the caller needs to then cast that to another type, the caller should take care of that.

imikejackson (approved):

PR #1440, opened 2025-09-22
BUG: Adds more checks to the HDFType of a data set and returns better error messages

imikejackson:
When attempting to read HDF5 H5T_COMPOUND types no error was thrown leaving the user to wonder what was happening. We now return an error message indicating that simplnx does not import H5T_COMPOUND datasets. Currently.

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadHDF5DatasetFilter.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadHDF5DatasetFilter.cpp
@@ -291,13 +291,13 @@ IFilter::PreflightResult ReadHDF5DatasetFilter::preflightImpl(const DataStructur
     Result<DataType> typeResult = datasetReader.getDataType();
     if(typeResult.invalid())
     {
-      return {nonstd::make_unexpected(typeResult.errors())};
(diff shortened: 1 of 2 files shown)


mmarineBlueQuartz (approved):

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadHDF5DatasetFilter.cpp
```diff
@@ -287,13 +287,16 @@ IFilter::PreflightResult ReadHDF5DatasetFilter::preflightImpl(const DataStructur
...
-    else
+
+    Result<DataType> typeResult = datasetReader.getDataType();
+    if(typeResult.invalid())
     {
-      Result<DataType> typeResult = datasetReader.getDataType();
-      DataType dataType = std::move(typeResult.value());
-      auto action = std::make_unique<CreateArrayAction>(dataType, tDims, cDims, dataArrayPath);
-      resultOutputActions.value().appendAction(std::move(action));
+      return {nonstd::make_unexpected(typeResult.errors())};
```
JDuffeyBQ:
```suggestion
      return {ConvertInvalidResult<OutputActions>(typeResult)};
```

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #1449, opened 2025-09-30
ENH: Crop Geometry Parameter

joeykleingers:
+ Adds a new Crop Geometry parameter that can be used to provide image geometry cropping inputs to filters.
+ Added the new Crop Geometry parameter to the image stack reader filter.
+ Added a new unit test to the image stack reader test to test the 270 unique cropping, resampling, flipping, and grayscale combinations.

NOTE: Make sure to merge in https://github.com/BlueQuartzSoftware/DREAM3DNX/pull/983 after this one.


src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/ITKImportImageStackFilter.cpp
```diff
@@ -491,89 +584,235 @@ IFilter::PreflightResult ITKImportImageStackFilter::preflightImpl(const DataStru
...
     {
-      return MakePreflightErrorResult(-23502, "Color to GrayScale conversion is disabled because the 'Color to GrayScale' filter is missing from the SimplnxCore plugin.");
+      auto errorResult = MakePreflightErrorResult(-23502, "Color to GrayScale conversion is disabled because the 'Color to GrayScale' filter is missing from the SimplnxCore plugin.");
+      return errorResult;
+    }
+
+    Arguments grayscaleImageGeomArgs;
+
+    //    static inline constexpr StringLiteral k_ConversionAlgorithm_Key = "conversion_algorithm_index";
+    //    static inline constexpr StringLiteral k_ColorChannel_Key = "color_channel";
```
nyoungbq:
```suggestion
```

nyoungbq (approved):
Code logic looks good. The only proposed change is about commented out code. 

I did notice there are a lot of `auto` type declarations in the filter source file. Some of them are valid, in that the type is expressed elsewhere in the definition (think `const auto& someVariable = dataStructure.getDataRefAs<some_type_here>(path)`). But there is more that one instance where I am not sure what the type of the variable is. I know that IDEs will show the type if you hover over it, but I think the readability suffers when you have to use additional tools to discern information (its the reason typescript exists and why python has type hints). It doesn't have to be fixed in this PR, but in the future try to be more conservative with its usage. A good rule of thumb is to only use when type is overly ownerous (think iterators/lambdas), in definition (getDataRefAs), or in element iteration loops (`for(auto& x : y)`).

imikejackson (changes requested):
I agree with NYoung about the over use of 'auto'. Please update the code accordingly.

imikejackson (changes requested):
The Python dev docs need to be updated to review how this parameter is used. Also if reasonably possible, an example in the python example scripts should be written to show a concrete use of the parameter in a working python script.

imikejackson (approved):

PR #1451, opened 2025-10-03
BUG: Use a Static Boolean to Register ITK Factories Once

### LINKED ISSUE #1447: BUG: ITKImageProcessingPlugin does not check if a factory has already been registered before registering it again.
ITKImageProcessingPlugin does not check if a factory has already been registered before registering it again, resulting in multiple registrations of the same factories and gradual slowdowns when executing a unit test many times in a row.  

Example: unit test that loads the plugins and then generates 270 SECTIONs (resulting in 270 separate runs of the entire test and subsequently 270 calls to load the plugins).

Also double check if this issue is affecting the other plugins and fix those too.

nyoungbq:
Fixes: #1447

### DIFF
--- a/src/Plugins/ITKImageProcessing/src/ITKImageProcessing/ITKImageProcessingPlugin.cpp
+++ b/src/Plugins/ITKImageProcessing/src/ITKImageProcessing/ITKImageProcessingPlugin.cpp
@@ -35,10 +35,11 @@ ITKImageProcessingPlugin::ITKImageProcessingPlugin()
     addFilter(filterFunc);
   }
 
-  if(!isRegistered)
+  static bool s_IsRegistered = false;
+  if(!s_IsRegistered)
   {
     RegisterITKImageIO();
-    isRegistered = true;
+    s_IsRegistered = true;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/Plugins/ITKImageProcessing/src/ITKImageProcessing/ITKImageProcessingPlugin.cpp
```diff
@@ -34,7 +34,12 @@ ITKImageProcessingPlugin::ITKImageProcessingPlugin()
   {
     addFilter(filterFunc);
   }
-  RegisterITKImageIO();
+
+  if(!isRegistered)
+  {
+    RegisterITKImageIO();
+    isRegistered = true;
+  }
```
JDuffeyBQ:
```suggestion
  static bool s_IsRegistered = false;
  if(!s_IsRegistered)
  {
    RegisterITKImageIO();
    s_IsRegistered = true;
  }
```
I would suggest moving the bool inside the function since it is only relevant to this spot.

JDuffeyBQ (changes requested):

JDuffeyBQ (approved):

PR #1457, opened 2025-10-21
STY: Clean up 'static inline' from filter headers

imikejackson:


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/InterpolatePointCloudToRegularGridFilter.hpp
nyoungbq, line 1: Two missed `inline`s here

nyoungbq (approved):

nyoungbq (approved):

PR #1464, opened 2025-10-29
BUG: When reading a legacy NeighborList, the incorrect group was passed to the reader

imikejackson:
Reading a 6.5.171 .dream3d file with neighbor lists would give warnings to the console about not being able to read the data sets with the number of neighbors. This will fix that. But I am wondering how reading the neighborlist actually has been working.


JDuffeyBQ (approved):
It's probably a good idea to go over the rest of the legacy import code to check for similar errors.

PR #1474, opened 2025-11-25
COMP: Fix MSVC template warnings

imikejackson:
MSVC Compiler warnings due to possible template type usage and converting between them. This PR attempts to fix the warnings by fixing the code.

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborCAxisMisalignments.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborCAxisMisalignments.cpp
@@ -71,7 +71,7 @@ Result<> ComputeFeatureNeighborCAxisMisalignments::operator()()
   const usize totalFeatures = featurePhases.getNumberOfTuples();
   const usize numQuatComps = featureAvgQuat.getNumberOfComponents();
 
-  std::vector<std::vector<double>> misalignmentLists(totalFeatures);
+  std::vector<std::vector<float32>> misalignmentLists(totalFeatures);
 
   const Eigen::Vector3d cAxis{0.0, 0.0, 1.0};
   usize hexNeighborListSize = 0;
@@ -129,7 +129,7 @@ Result<> ComputeFeatureNeighborCAxisMisalignments::operator()()
         }
 
         // Convert the misorientation to Degrees and store the value
-        currentMisalignmentList[j] = w * Constants::k_180OverPiD;
+        currentMisalignmentList[j] = static_cast<float32>(w * Constants::k_180OverPiD);
 
         // If we are finding the average misorientation, then start accumulating those values
         if(m_InputValues->FindAvgMisals)
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeArrayStatistics.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeArrayStatistics.cpp
@@ -542,11 +542,11 @@ public:
         {
           if(m_FindMedian)
           {
-            values.push_back(m_Source[i]);
+            values.push_back(static_cast<float>(m_Source[i]));
           }
           if(m_FindNumUniqueValues)
           {
-            valuesSet.emplace(m_Source[i]);
+            valuesSet.emplace(static_cast<int32>(m_Source[i]));
           }
         }
       }
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
@@ -19,6 +19,9 @@ constexpr usize k_MaxXIndex = 3;
 constexpr usize k_MaxYIndex = 4;
 constexpr usize k_MaxZIndex = 5;
 
+template <class T>
+concept ArithmeticNotBool = std::is_arithmetic_v<T> && !std::is_same_v<T, bool>;
+
 std::array<usize, 6> GetVoxelIndices(const Float32AbstractDataStore& unifiedBounds, usize targetBoundsIndex, const ImageGeom& image)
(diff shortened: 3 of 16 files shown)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborCAxisMisalignments.cpp
```diff
@@ -129,7 +129,7 @@ Result<> ComputeFeatureNeighborCAxisMisalignments::operator()()
         }
 
         // Convert the misorientation to Degrees and store the value
-        currentMisalignmentList[j] = w * Constants::k_180OverPiD;
+        currentMisalignmentList[j] = static_cast<float32>(w * Constants::k_180OverPiF);
```
JDuffeyBQ: Here `k_180OverPiF` gets promoted to `float64` due to `w` so I think keeping it as `k_180OverPiD` is ok.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeArrayStatistics.cpp
```diff
@@ -526,12 +526,12 @@ class MedianByFeatureRangeImpl
   {
   }
 
-  void compute(usize start, usize end) const
+  void compute(size_t start, size_t end) const
```
JDuffeyBQ: `usize` should be the same as `size_t`. Was there a warning here?

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
```diff
@@ -144,7 +144,14 @@ class ComputeBaseStatsImpl
             count++;
             minValue = std::min(minValue, value);
             maxValue = std::max(maxValue, value);
-            summationValue += value;
+            if constexpr(std::is_same_v<T, bool>)
+            {
+              summationValue = value == true ? true : false;
```
JDuffeyBQ: This is the same as `summationValue = value`. But is this what we want to do here?

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeArrayStatistics.cpp
imikejackson: Yes. there was a warning.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
imikejackson: Just throw a runtime exception if somehow we make it in there with a boolean.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
```diff
@@ -111,6 +111,10 @@ class ComputeBaseStatsImpl
   , m_UnifiedBounds(unifiedBounds)
   , m_StatsVector(statsVector)
   {
+    if constexpr(std::is_same_v<T, bool>)
```
JDuffeyBQ: I think we should use a modified version of `ExecuteDataFunction` so that `bool` doesn't get used and compiled. And also check that the filter disallows `bool`.

imikejackson: Fixed

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
```diff
@@ -372,7 +372,14 @@ class ComputeBasicAndFrequencyStatsImpl
             count++;
             minValue = std::min(minValue, value);
             maxValue = std::max(maxValue, value);
-            summationValue += value;
+            if constexpr(std::is_same_v<T, bool>)
```
JDuffeyBQ: Now that we've disallowed bool from the template we don't need these `if constexpr` blocks anymore. We could also add an additional check and add a concept to the template parameter of the struct to prevent bool from working.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
```diff
@@ -490,7 +492,8 @@ class ComputeStdDevImpl
 
       // We are working with primitives here for their trivially copyable nature, this lets us cut accesses to output vector
       float64 sumOfDiffs = 0.0f;
-      float32 meanValue = m_StatsVector[targetBoundsIndex].summationValue / static_cast<float32>(m_StatsVector[targetBoundsIndex].count);
+      float32 meanValue = 0.0f;
+      meanValue = m_StatsVector[targetBoundsIndex].summationValue / static_cast<float32>(m_StatsVector[targetBoundsIndex].count);
```
JDuffeyBQ:
```suggestion
      float32 meanValue = m_StatsVector[targetBoundsIndex].summationValue / static_cast<float32>(m_StatsVector[targetBoundsIndex].count);
```

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeBoundingBoxStats.cpp
```diff
@@ -19,6 +19,9 @@ constexpr usize k_MaxXIndex = 3;
 constexpr usize k_MaxYIndex = 4;
 constexpr usize k_MaxZIndex = 5;
 
+template <class T>
+concept ArithmeticNotBool = std::is_arithmetic_v<T> && !std::is_same_v<T, bool>;
```
JDuffeyBQ: Needs to be used.

mmarineBlueQuartz (approved):

PR #1476, opened 2025-12-03
BUG/ENH: Fix Backwards Pipeline Compatibility and Add Testing

nyoungbq:
Conversion updates
Test Added
Has a sibling branch on `SimplnxReview`
The tool to generate the test data is on the SIMPL repo

Testing Methodology:

- For `6.6` a "mega-pipeline" was generated with values according to parameter type. This not only verifies that the values are read in, but that they are read in correctly. Any exclusions (ie new filter parameters) must be explicitly excluded. Any new filter registered to the Plugin UUID conversion maps will be tested automatically.
- For `6.5.171` every default pipeline was updated to the `6.5.171` release standard (except synthetics, noted in `README` in test data). These are validated that the values are read in and that they match the expected type.  Any new filter registered to the Plugin UUID conversion maps will be tested automatically, SO LONG AS IT APPEARS IN THE DEFAULT `6.5.171` PIPELINES.

**Note any filter that is not in a plugin-level conversion UUID map will _NOT_ be tested.**

**Also note that just because it reads them in properly does not mean the pipeline cannot be run without errors**. Example: consider a real case where an AttributeMatrix was hard-coded to the name `vertex data` in SIMPL, thus filters that follow in the pipeline may have the path `../vertex data/..`. However, in `simplnx` the filter was made more general and thus uses `cell data` as the hard-coded AM name, thus filters that follow in the pipeline may have the path `../cell data/..`. When conversion occurs the `../vertex data/..` is read in properly, but the array actually exists at `../cell data/..` causing a disparity.


imikejackson:
@nyoungbq Just tried a quick test and I copied the (08) SmallIN100 Full Reconstruction from DREAM.3D version 6.5.171 into DREAM3D-NX and looks like it is mostly correct but I got an error on filter 13 about a missing data array path and the issue was in filter 11 "Compute Feature Average Orientations", the "Feature Attribute Matrix" was not set correctly or transferred correctly from the old pipeline.

nyoungbq:
@imikejackson okay I will pull all the default pipelines from 6.5.171 and test/patch for them. The mega-pipeline generated for this was from 6.6

imikejackson:
I took a look and generally looks very well done considering all the caveats that we have to deal with. Something that I think needs to be addressed is when we update a filter's parameters, and thus updating the version of the parameter the unit test will probably start to fail. I had this happen with the ITKImageWriter filter after I added a few new parameters into the SIMPLNX version. I had to go into the unit test and update the "ignore_keys" mapping.

If this is something that will need to happen each time with modify an existing filter that had a SIMPL version of the filter, then I think we need something written up somewhere that can help the developer underestand what went wrong and what to update in this new unit test. Possibly a section down in simplnx/docs/Porting_Filters.md or possibly some comments within the unit test file since that source file might be the first stop when debugging the failure.

imikejackson (approved):

PR #1479, opened 2025-12-05
FILTER: Write LAMMPS File

nyoungbq:
Ported Write LAMMPS File

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/WriteLAMMPSFile.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/WriteLAMMPSFile.cpp
@@ -1,0 +1,154 @@
+#include "WriteLAMMPSFile.hpp"
+
+#include "simplnx/Common/Constants.hpp"
+#include "simplnx/DataStructure/DataArray.hpp"
+#include "simplnx/DataStructure/Geometry/VertexGeom.hpp"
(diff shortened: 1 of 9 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/WriteLAMMPSFile.cpp
```diff
@@ -0,0 +1,154 @@
...
+  ThrottledMessenger throttledMessenger = messageHelper.createThrottledMessenger();
+  // Write the Atom positions (Vertices)
+  usize numVerts = verts.getNumberOfTuples();
+  for(usize i = 0; i < numVerts; i++)
+  {
+    if(m_ShouldCancel)
+    {
+      return {};
+    }
+    throttledMessenger.sendThrottledMessage([&]() { return fmt::format("Exporting Data {:.2f}% completed", CalculatePercentComplete(i, numVerts)); });
```
imikejackson: How expensive is this call. Maybe we should just check every "Nth" vert. 

Also I think we should build up a buffer or set the buffer on the "file" object to be like 4MB at the minimum.

Create a geometry that is 1000 x 1000 x 1000 and write the file and measure how long it takes. We should try to maximize that through put best we can.

imikejackson (changes requested):
Check code comments

imikejackson (approved):

PR #1488, opened 2025-12-10
[WIP] PY: Update to Pybind11 v3.0.1 so that Python 3.13 can be used.

imikejackson:


wrapping/python/CxPybind/CxPybind/CxPybind.hpp
```diff
@@ -116,10 +116,13 @@ std::string GetFullPythonName()
   {
     return GetRegisteredPythonTypeName<T>();
   }
-  else
-  {
-    return GetFixedPythonTypeCasterName<T>();
-  }
+  // NOTE: With Pybind11 v3.0.1 we drop into this branch which when called from
```
JDuffeyBQ:
```cpp
#if PYBIND11_HAS_NATIVE_ENUM
#include <pybind11/native_enum.h>
#endif

template <class T>
std::string GetFullPythonName()
{
  using CasterT = py::detail::make_caster<T>;
  // Use py::type::of for registered C++ types otherwise use py::detail::type_caster::name
  if constexpr(std::is_base_of_v<py::detail::type_caster_generic, CasterT>)
  {
    return GetRegisteredPythonTypeName<T>();
  }
#if PYBIND11_HAS_NATIVE_ENUM
  else if constexpr(std::is_base_of_v<py::detail::type_caster_enum_type<T>, CasterT>) // Is registered enum
  {
    const std::type_info& typeInfo = typeid(T);
    py::handle nativeEnumHandle = py::detail::global_internals_native_enum_type_map_get_item(typeInfo);
    if(nativeEnumHandle)
    {
      return GetFullPythonNameFromType(nativeEnumHandle);
    }
    py::handle handle = py::detail::get_type_handle(typeInfo, true);
    return GetFullPythonNameFromType(handle);
  }
#endif
  else
  {
    return GetFixedPythonTypeCasterName<T>();
  }
}
```
I think this should fix the issue. Pybind11 3 added `pybind11::native_enum` and with it changed some of the associated type caster code. Previously enums would fall under the first branch, but with the changes they incorrectly fall into the last branch. This adds a new branch just for handling enums.

imikejackson: Done

PR #1496, opened 2026-01-06
ENH: Compute Array Statistics error messages and documentation updates

imikejackson:
- Update CreateAttributeMatrixAction error messages to clarify errors
- Add check inside of CreateAttributeMatrixAction to ensure the parent is NOT already an AttributeMatrix
- Update Compute Array Statistics documentation to clarify where the destination AttributeMatrix can be created
- Adjust one of the separator parameters to clarify what the enclosed parameters are used for.


src/Plugins/SimplnxCore/docs/ComputeArrayStatisticsFilter.md
```diff
@@ -28,9 +27,17 @@ The user may optionally use a mask to specify points to be ignored when computin
...
+## Destination Attribute Matrix 
+
+The user must create a destination **Attribute Matrix** in which the computed statistics will be stored. DREAM3D-NX enforces a rule where any Attribute Matrix cannot containe another AttributeMatrix. With this in mind the user should select a destination that is not itself an Attribute Matrix such as the top level of a Geometry or the top level of the Data Structure itself. The user could have also created a group (using a previous filter) and use that group as the destination.
```
nyoungbq:
```suggestion
The user must create a destination **Attribute Matrix** in which the computed statistics will be stored. DREAM3D-NX enforces a rule where any Attribute Matrix cannot contain another Attribute Matrix. With this in mind, the user should select a destination that is not itself an Attribute Matrix, such as the top level of a Geometry or the top level of the Data Structure itself. The user could have also created a group (using a previous filter) and use that group as the destination.
```

src/Plugins/SimplnxCore/docs/ComputeArrayStatisticsFilter.md
```diff
@@ -28,9 +27,17 @@ The user may optionally use a mask to specify points to be ignored when computin
...
+Special operations occur for certain statistics if the supplied array is of type *bool* (for example, a mask array produced when thresholding).  The length, minimum, maximum, median, mode,and summation are computed as normal (although the resulting values may be platform dependent).  The mean and standard deviation for a boolean array will be true if there are more instances of true in the array than false.  If *Standardize Data* is chosen for a boolean array, no actual modifications will be made to the input.  These operations for boolean inputs are chosen as a basic convention, and are not intended be representative of true boolean logic.
```
nyoungbq:
```suggestion
Special operations occur for certain statistics if the supplied array is of type *bool* (for example, a mask array produced from threshold filters).  The *length*, *minimum*, *maximum*, *median*, *mode*, and *summation* are computed as normal (although the resulting values may be platform dependent).  The *mean* and *standard deviation* for a boolean array will be true if there are more instances of true in the array than false.  If *Standardize Data* is chosen for a boolean array, no actual modifications will be made to the input.  These operations for boolean inputs are chosen as a basic convention, and are not intended be representative of true boolean logic.
```

nyoungbq (approved):

PR #1497, opened 2026-01-07
BUG: MergeTwins 'Active' array not initialized correctly plus other e…

imikejackson:
…nhancements

- User facing parameter to optionally randomize parentIds
- getSeed() function updated to actually produce a new random value each time
- Copied from SIMPL the GroupFeatures::Execute() function so that this algorithm behaves more closely to DREAM3D 6.5.172
- The 'Active' array is not properly initialized with "true" values

NOTE: In order to reproduce the test data, the developer will need to build a version of DREAM.3D from the branch v6_5_172 or newer.


jmarquisbq (approved):
I'm not sure why the builds keep failing but the changes seem to work and produce the expected results now from my side.

PR #1499, opened 2026-01-08
REV: Bad Data Neighbor Orientation Check

nyoungbq:
- Cleaned up algorithm,filter, and test case code
- Removed preflight checks that are already preformed by parameters
- Added Test case using real world data compared against 6.5 output
- Added extensive readme about data and testing methodology in new test data file
- Verified and documented the identical results between 6.5 and NX


nyoungbq:
Here is the documentation files zipped into the data repo
[README.md](https://github.com/user-attachments/files/24938929/README.md)
[test_design.md](https://github.com/user-attachments/files/24938933/test_design.md)

imikejackson (approved):

PR #1502, opened 2026-01-14
ENH: Defaulting Multi-ThresholdObjects output array to uint8.

joeykleingers:


imikejackson:
@joeykleingers Why did the 2 AlignSections unit tests need to change?

joeykleingers:
> @joeykleingers Why did the 2 AlignSections unit tests need to change?

They use the MultiThresholdObjects filter inside the tests and then compare with an exemplary, and the exemplary output array was still boolean, so it would fail.  So I updated the exemplary data to have a uint8 mask array instead.

imikejackson:
Or. you could just set the output for the filter to use boolean and not have to update any tests?

imikejackson (approved):

PR #1504, opened 2026-01-14
BUG: Compute Feature Reference Misorientation algorithm bug fixes

imikejackson:
The previous version was using the average quaternion everywhere instead of switching on the type of reference misorientation to use: Average Quaternion or Cell level misorientation data


nyoungbq (approved):

nyoungbq:
The fix for the broken test is adding the new parameter keys to the backwards compatibility ignore key map. if the filter is not in the list already add it, otherwise you can just update the existing entry with the new param keys

PR #1506, opened 2026-01-16
BUG: Compute Euclidean Dist Map now does not create extra arrays.

### LINKED ISSUE #1505: Compute Euclidean Distance Map creates arrays that the user did not ask for
Image

The extra arrays should not be created.

The unit tests may need to be updated to reflect this.

joeykleingers:


imikejackson:
@joeykleingers Do you think we need an updated unit test for this? Not sure if we need new data files but do we need variations on the parameters to ensure the filter is at least producing the expected results?

imikejackson (approved):

PR #1510, opened 2026-01-21
FILT: Extract Feature Boundaries 2D filter implemented

imikejackson:
Claude Code (Sonnet 4.5) was used to write this filter, documentation, unit test and test files.

- Documentation Updated
- Unit test implemented
- New unit test data archive file uploaded to GitHub and Sync'ed

## Code Cleanup
- [x] No commented-out code (rare exceptions to this is allowed)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.hpp
```diff
@@ -0,0 +1,49 @@
...
+{
+
+struct SIMPLNXCORE_EXPORT ExtractFeatureBoundaries2DInputValues
+{
+  DataPath InputImageGeometryPath;
+  DataPath FeatureIdsArrayPath;
+  DataPath OutputEdgeGeometryPath;
+  uint64 ZValueChoice;
+  float32 CustomZValue;
+  bool ExtractVirtualSampleEdges;
```
JDuffeyBQ:
```suggestion
  uint64 ZValueChoice = 0;
  float32 CustomZValue = 0.0f;
  bool ExtractVirtualSampleEdges = false;
```

src/Plugins/SimplnxCore/test/SliceTriangleGeometryTest.cpp
```diff
@@ -12,7 +12,7 @@
 
 using namespace nx::core;
 
-namespace
+namespace slice_triangle_geometry_test
```
JDuffeyBQ: Why is this no longer in an anonymous namespace?

src/Plugins/SimplnxCore/test/CreateAMScanPathsTest.cpp
```diff
@@ -14,7 +14,7 @@ using namespace nx::core;
 using namespace nx::core::Constants;
 using namespace nx::core::UnitTest;
 
-namespace
+namespace create_am_scan_paths_test
```
JDuffeyBQ: Why is this no longer in an anonymous namespace?

src/Plugins/SimplnxCore/test/ExtractFeatureBoundaries2DTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+      // Write the DataStructure out to the file system
+#ifdef SIMPLNX_WRITE_TEST_OUTPUT
+      UnitTest::WriteTestDataStructure(dataStructure, fs::path(fmt::format("{}/test_output_{}", unit_test::k_BinaryTestOutputDir, fileName)));
+#endif
+
+      // Verify the output edge geometry was created
+      const auto* edgeGeom = dataStructure.getDataAs<EdgeGeom>(extract_features_boundaries_2d_test::k_OutputEdgeGeometryPath);
+      REQUIRE(edgeGeom != nullptr);
+      REQUIRE(edgeGeom->getNumberOfEdges() > 0);
+      REQUIRE(edgeGeom->getNumberOfVertices() > 0);
```
JDuffeyBQ: Do we know what the actual expected values for these are?

src/Plugins/SimplnxCore/test/ExtractFeatureBoundaries2DTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+namespace extract_features_boundaries_2d_test
+{
+const DataPath k_ImageGeometryPath({"ImageGeometry"});
+const DataPath k_FeatureIdsPath({"ImageGeometry", "Cell Data", "FeatureIds"});
+const DataPath k_OutputEdgeGeometryPath({"Computed Feature Boundaries"});
+const DataPath k_ExemplarEdgeGeometryPath({"Feature Boundaries"});
+
+} // namespace extract_features_boundaries_2d_test
+
+TEST_CASE("SimplnxCore::ExtractFeatureBoundaries2DFilter: 01_simple_adjacent", "[SimplnxCore][ExtractFeatureBoundaries2DFilter]")
```
JDuffeyBQ: Could some of the code in these test cases be factored out?

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ExtractFeatureBoundaries2DFilter.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  // Estimate the maximum number of edges:
+  // Vertical internal edges: (dimX - 1) * dimY
+  // Horizontal internal edges: dimX * (dimY - 1)
+  // Outer boundary edges (if enabled): 2 * dimX + 2 * dimY
+  usize dimX = dims.getX();
+  usize dimY = dims.getY();
+  usize maxEdges = (dimX > 0 && dimY > 0) ? ((dimX - 1) * dimY + dimX * (dimY - 1)) : 0;
+  if(pExtractVirtualSampleEdges && dimX > 0 && dimY > 0)
```
JDuffeyBQ:
```suggestion
  usize maxEdges = (dimX - 1) * dimY + dimX * (dimY - 1);
  if(pExtractVirtualSampleEdges)
```
`dimX` and `dimY` are `usize` so they are always greater than zero.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
```diff
@@ -0,0 +1,646 @@
...
+        }
+      }
+    }
+    m_EdgeCount.fetch_add(localCount, std::memory_order_relaxed);
+  }
+
+private:
+  const AbstractDataStore<T>& m_FeatureIds;
+  usize m_DimX;
+  usize m_DimY;
```
JDuffeyBQ:
```suggestion
  usize m_DimX = 0;
  usize m_DimY = 0;
```

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
```diff
@@ -0,0 +1,646 @@
...
```
JDuffeyBQ:
```suggestion
  usize m_DimX = 0;
  usize m_DimY = 0;
```

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
```diff
@@ -0,0 +1,646 @@
...
+ *   +---+---+---+---+
+ *   | 0 | 1 | 2 | 3 |  y=2
+ *   +---+---+---+---+  <- Horizontal edges checked here (y=1 to y=2)
+ *   | 0 | 1 | 2 | 3 |  y=1
+ *   +---+---+---+---+  <- Horizontal edges checked here (y=0 to y=1)
+ *   | 0 | 1 | 2 | 3 |  y=0
+ *   +---+---+---+---+
+ */
+template <typename T>
+class CountHorizontalEdgesImpl
```
JDuffeyBQ: These two classes are almost exactly the same. I think we could combine them to reduce code. Then alias then to the appropriate vertical or horizontal one.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
```diff
@@ -0,0 +1,646 @@
...
+    float32 originX = origin.getX();
+    float32 originY = origin.getY();
+    float32 spacingX = spacing.getX();
+    float32 spacingY = spacing.getY();
+
+    // =========================================================================
+    // Z VALUE CALCULATION: Determine the Z coordinate for all generated vertices
+    // =========================================================================
+    float32 zValue = 0.0f;
+    switch(zValueChoice)
```
JDuffeyBQ, line 244: We have an enum for this so we should use that.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
```diff
@@ -0,0 +1,646 @@
...
+    // =========================================================================
+    // This two-pass approach (count then populate) allows us to allocate the
+    // exact amount of memory needed upfront, avoiding dynamic resizing.
+    std::atomic<usize> verticalEdgeCount{0};
+    std::atomic<usize> horizontalEdgeCount{0};
+
+    // Count vertical edges (between horizontally adjacent cells)
+    {
+      CountVerticalEdgesImpl<T>(featureIdsStore, dimX, dimY, verticalEdgeCount, shouldCancel).operator()({0, dimY});
+    }
```
JDuffeyBQ:
```suggestion
    {
      CountVerticalEdgesImpl<T>(featureIdsStore, dimX, dimY, verticalEdgeCount, shouldCancel)({0, dimY});
    }
```
Why is this in a new scope? Also no reason to use `operator()` here. Also the way these classes are setup look like they are meant to be parallelized but they are not?

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
```diff
@@ -0,0 +1,646 @@
...
+  const auto& featureIdsArray = m_DataStructure.getDataRefAs<IDataArray>(m_InputValues->FeatureIdsArrayPath);
+
+  // Get the data type to dispatch to the correct template instantiation
+  DataType dataType = featureIdsArray.getDataType();
+
+  m_MessageHandler(IFilter::Message::Type::Info, "Extracting feature boundaries...");
+
+  // ExecuteDataFunction dispatches to the templated functor based on dataType.
+  // This allows the algorithm to work with any integer type for FeatureIds
+  // (int8, int16, int32, int64, uint8, uint16, uint32, uint64).
```
JDuffeyBQ: Some comments seem to be extraneous.

src/Plugins/SimplnxCore/test/SliceTriangleGeometryTest.cpp
imikejackson: The tests were failing because the variable names conflicted with those declared in UnitTestCommon.hpp

With these tests there are constants that are used in all the TEST_CASE() which are up in typically an anonymous namespace. Conflicts with either the naming of the variable or the actual value of the constant are starting to happen. There are a few ways to fixt this:

## Sledge Hammer Approach

The **only** way we would be able to fix that is to go back to **every** single test file and ensure that top level objects are **always** named the same: DataContainer, ImageGeometry, TriangleGeometry, EdgeGeometry. Then we go back and regenerate each and every single input test and exemplar file to use that naming scheme. Upload all those to GitHub. Also update every single unit test.

## Keep Updating UnitTestCommon

We can just keep adding values with variations on naming to UnitTestCommon.hpp so that we can keep an anonymous namespace? At some point we are going to end up with names like `slice_triangle_geometry_test_k_InputTriangleGeometryPath` which seems silly at best.

## Keep using Anonymous Namespaces

Sort of the same as what is in UnitTestCommon BUT we declare those variables in the unit test file itself. Again, you are going to run into naming conflicts and the names just become really long and unwieldly.

## Remove the `using namespace nx::core::Constants`

Remove the declaration of `using namespace nx::core::Constants`  from the top of **every** unit test file and then update the codes within the unit test file to prefix all constants with "nx::core::Constants" so that there are no conflicts.

## Use unit test specific namespaces

Just declare a namespace that is specifically for the unit test where constants are kept. Clean. Simple. This is my vote and what I am going to do.

src/Plugins/SimplnxCore/test/CreateAMScanPathsTest.cpp
imikejackson: See other reply about namespaces

src/Plugins/SimplnxCore/test/ExtractFeatureBoundaries2DTest.cpp
imikejackson: Looking at the code deeper, we compare both the shared vertex list and the shared edge list between the computed and the exemplar so we do not need these comparisons. They have been removed.

src/Plugins/SimplnxCore/test/ExtractFeatureBoundaries2DTest.cpp
imikejackson: Yep.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
imikejackson: Not quite exactly but I added some template parameters to make the necessary changes.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
imikejackson: Fixed

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
imikejackson: Nope. Typically `FeatureIds` are int32, in this case the filter will allow a `FeatureIds` array to be any integer type so I think the comment is appropriate.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractFeatureBoundaries2D.cpp
imikejackson: Claude code generated the infrastructure of the class. It was using our other filters and algorithms to patterns itself and to mimic how we write other filters. I don't believe that the algorithm can run in parallel. I will have to analyze the source code deeper to ensure that it could infact run in parallel.

The other suggestions have been addressed.

JDuffeyBQ (approved):

PR #1514, opened 2026-01-28
ENH: IArray to/from string api

jmarquisbq:


src/simplnx/DataStructure/IArray.hpp
```diff
@@ -103,6 +103,26 @@ class SIMPLNX_EXPORT IArray : public DataObject
...
 
+  /**
+   * @brief Returns the value at the tuple and component index as a std::string.
+   *        NOTE: This function is slow and should be used sparingly and avoided inside of a tight loop!
+   * @param tupleIndex
+   * @param compIndex
+   * @param format Optional fmt formatting, this will be ignored for integer types and a simple {} will always be used
+   * @return std::string
+   */
+  virtual std::string toString(const usize tupleIndex, const usize compIndex, const std::string& format = "{}") const = 0;
```
JDuffeyBQ: Is there a particular reason for parameters passed by value as const? It's not incorrect but most of our functions aren't defined that way.

JDuffeyBQ (commented):
Generally looks good. I think I have a change that can reduce the amount of code for the definitions. I can just push that if we want.

src/simplnx/DataStructure/IArray.hpp
jmarquisbq: @JDuffeyBQ No good catch. That was just an oversight on my part. I can go ahead and update that unless this already gets hit with the optimization you were referring to?

JDuffeyBQ: Nope I didn't change that so you are good to push.

imikejackson (approved):

PR #1515, opened 2026-01-28
ENH: Fill Bad Data Out-of-Core Optimization

imikejackson:
Used Claude-Code (Sonnet 4.5) to reimplement this filter's algorithm to better optimize for out-of-core performance without hugely impacting the in-core performance.

Both out-of-core builds and in-core builds pass the same unit test indicating that for the given input data set we are producing the same results. This code should be actually verified with smaller, more concise data sets to completely validate the algorithm is producing what we expect it to produce.

Claude-Code generated more unit test files that better exercise the edge cases of the algorithm. The Small IN100 test case exercises the out-of-core performance.


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/FillBadData.cpp
```diff
@@ -60,200 +165,278 @@ FillBadData::FillBadData(DataStructure& dataStructure, const IFilter::MessageHan
...
-  // int64 plane = 0;
-  //  int64 neighborPoint = 0;
-  //  int32 featureName = 0;
-  //  int32 feature = 0;
-  size_t numFeatures = 0;
-  size_t maxPhase = 0;
+// -----------------------------------------------------------------------------
+void FillBadData::phaseOneCCL(Int32AbstractDataStore& featureIdsStore, ChunkAwareUnionFind& unionFind, std::unordered_map<usize, int64_t>& provisionalLabels, const std::array<int64_t, 3>& dims)
+{
+  // const auto neighborOffsets = getNeighborOffsets(dims);
```
nyoungbq: The `getNeighborOffsets()` call is commented out but you have inlined the calculations it would be preforming below in this function. I fell this should either be removed (and the function with it) in favor of inlining or replace the inlined checks below with the precached values from function. I would assume precaching is the more efficient route but without benchmarking I can't be sure.

nyoungbq (approved):
Approving, suggestion is mostly for the sake of cleanup. The Union-Find structure looks to be all good. From the few times I have used Union-Find, I am not sure whether compressing the paths should be done at every merge. If your doing frequent merges you are wasting cycles updating values that will be updated again if it gets merged again, rather than one pass through at flatten for future searches. I understand it cuts down the depths of the trees though, so perhaps that provides an optimization when searching the trees during construction. Also worth noting the types sometimes use our type declarations `uint64` and others use the standard `uint64_t`, I would bring them all in line one way or the other. The rest looks good as far as I can tell.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/FillBadData.cpp
imikejackson: The updates you suggested have been integrated into the codes. Great suggestions.

test/UnitTestCommon/include/simplnx/UnitTest/UnitTestCommon.hpp
```diff
@@ -278,6 +278,37 @@ class TestFileSentinel
...
+
+  PreferencesSentinel(const PreferencesSentinel&) = delete;            // Copy Constructor Not Implemented
+  PreferencesSentinel(PreferencesSentinel&&) = delete;                 // Move Constructor Not Implemented
+  PreferencesSentinel& operator=(const PreferencesSentinel&) = delete; // Copy Assignment Not Implemented
+  PreferencesSentinel& operator=(PreferencesSentinel&&) = delete;      // Move Assignment Not Implemented
+
+private:
+  std::string m_OriginalFormat;
+  int64 m_OriginalSize;
+  bool m_OriginalForceOoc;
```
JDuffeyBQ:
```suggestion
  int64 m_OriginalSize = 0;
  bool m_OriginalForceOoc = false;
```

PR #1519, opened 2026-01-30
[WIP] STY: Update filter code to implement a standard naming convention

imikejackson:
Using claude-code to update filter source codes to use a more consistent naming convention which should more closely match the clang-tidy file but also become easier to read.

Rules such as:
pointer variables should have `Ptr` suffix
reference variables should have `Ref` suffix
Variables that point to or reference a Geometry type should end with `Geom`
When looping over every element in an ImageGeometry, we should be using "TotalVoxels" and "voxelIdx" instead of `i` in those loops.

Will update with more rules as they are implemented.


imikejackson:
This was primarily an experiment. Closing.

PR #1522, opened 2026-02-04
BUG: Fix int32 overflow bug and code refactoring in Require Minimum Feature Size

imikejackson:
Due to casting there could be an int32 overflow issue that occurs during the running of this filter. 

The filter was also heavily refactored in the process:
* Variable names were updated to reflect their actual usage
* One parameter type was changed from an int64 to an int32
* Functions were moved around and also renamed to conform to the current naming standards.

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
@@ -120,10 +120,17 @@ Result<> RequireMinimumSizeFeatures::operator()()
   {
     return {nonstd::make_unexpected(std::vector<Error>{errorReturn})};
   }
+  if(m_ShouldCancel)
+  {
+    return {};
+  }
 
   auto& imageGeom = m_DataStructure.getDataRefAs<ImageGeom>(m_InputValues->InputImageGeometryPath);
   assignBadVoxels(imageGeom.getDimensions(), featureNumCellsStoreRef);
-
+  if(m_ShouldCancel)
+  {
+    return {};
+  }
   DataPath cellFeatureGroupPath = m_InputValues->FeatureNumCellsPath.getParent();
   usize currentFeatureCount = featureNumCellsStoreRef.getNumberOfTuples();
 
@@ -147,6 +154,7 @@ void RequireMinimumSizeFeatures::assignBadVoxels(SizeVec3 dimensions, const Int3
 {
   MessageHelper messageHelper(m_MessageHandler);
 
+  messageHelper.sendMessage(fmt::format("RequireMinimumSizeFeatures::assignBadVoxels Starting"));
   Int32AbstractDataStore& featureIds = m_DataStructure.getDataAs<Int32Array>(m_InputValues->FeatureIdsPath)->getDataStoreRef();
   usize totalPoints = featureIds.getNumberOfTuples();
 
@@ -180,6 +188,10 @@ void RequireMinimumSizeFeatures::assignBadVoxels(SizeVec3 dimensions, const Int3
     counter = 0;
     for(int64 zIdx = 0; zIdx < dims[2]; zIdx++)
     {
+      if(m_ShouldCancel)
+      {
+        return;
+      }
       kstride = dims[0] * dims[1] * zIdx;
       for(int64 yIdx = 0; yIdx < dims[1]; yIdx++)
       {
@@ -262,6 +274,10 @@ std::vector<bool> RequireMinimumSizeFeatures::removeSmallFeatures(Int32AbstractD
 
   for(usize i = 1; i < totalFeatures; i++)
   {
+    if(m_ShouldCancel)
+    {
+      return {};
+    }
     if(!applyToSinglePhase)
     {
       if(featureNumCellsStoreRef.getValue(i) >= minAllowedFeatureSize)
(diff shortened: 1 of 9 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
```diff
@@ -1,10 +1,74 @@
...
+
+  ~RequireMinimumSizeFeaturesTransferDataImpl() = default;
+
+  void operator()() const
+  {
+    ThrottledMessenger throttledMessenger = m_MessageHelper.createThrottledMessenger();
+    std::string arrayName = m_DataArrayPtr->getName();
+    for(usize i = 0; i < m_TotalPoints; i++)
+    {
+      if(m_TotalPoints % 100 == 0)
```
nyoungbq: I would highly highly recommend benchmarking this. The branching and modulo cost is incredibly expensive in this I suspect. The throttled messenger already gates it to every one second. I have benchmarked similar checks and at best the time cost is probably equivalent without the if (in which case removing it would still be more efficient). if you find the if is more efficient after benchmark, change the number `100` to a single bit in each byte (64, 128, 256) so the compiler can convert the modulo to a bit check to make it more efficient

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
```diff
@@ -22,5 +86,240 @@ RequireMinimumSizeFeatures::~RequireMinimumSizeFeatures() noexcept = default;
...
+
+
+void RequireMinimumSizeFeatures::assignBadVoxels( SizeVec3 dimensions, const Int32AbstractDataStore& featureNumCellsStoreRef)
+{
+  MessageHelper messageHelper(m_MessageHandler);
+
+  Int32AbstractDataStore& featureIds = m_DataStructure.getDataAs<Int32Array>(m_InputValues->FeatureIdsPath)->getDataStoreRef();
+  usize totalPoints = featureIds.getNumberOfTuples();
+
+  std::array<int64_t, 3> dims = {
```
nyoungbq: int64_t -> int64

Search and replace because its in multiple locations

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
```diff
@@ -22,5 +86,240 @@ RequireMinimumSizeFeatures::~RequireMinimumSizeFeatures() noexcept = default;
...
+              if(l == 2 && i == 0)
+              {
+                good = 0;
+              }
+              if(l == 3 && i == (dims[0] - 1))
+              {
+                good = 0;
+              }
+
+              if(good == 1)
```
nyoungbq:
```suggestion
              neighborVoxelIdx = count + neighborVoxelIndexOffsets[l];
              if((l == 0 && k == 0) ||
                 (l == 1 && j == 0) ||
                 (l == 2 && i == 0) ||
                 (l == 3 && i == (dims[0] - 1)) ||
                 (l == 4 && j == (dims[1] - 1)) ||
                 (l == 5 && k == (dims[2] - 1)))
              {
                continue;
              }
              else 
```

This leverages lazy evaluation to speed up edge case checks, while increasing readability

imikejackson: I have a fix for this incoming.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
imikejackson: The first line inside of the `sendThrottledMessage` is a call to `auto now = std::chrono::steady_clock::now();` which is not cheap. I was just trying to cut down on that. Probably going to have to do some benchmarking again for this use case.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
imikejackson: Fixed

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/RequireMinimumSizeFeatures.cpp
imikejackson: # Just using ThrottledMessenger by itself
Screenshot 2026-02-04 at 18 48 53

## Only allowing at most 100 trys using ThrottledMessenger

Screenshot 2026-02-04 at 18 48 29

imikejackson: Adding an additional layer of throttling helps to avoid the call to get the current time. For this particular use we are looping over all voxels in an image geometry so this is about as "tight" a loop as we can get. There is nothing eles in that loop (like extra calculations or function calls) that would dominate the loop. This is why calling ThrottledMessenger can have such an impact and why I defended against it a bit.

nyoungbq (approved):

PR #1531, opened 2026-02-16
TEST: Test ParameterVersion() and clone() functions of each filter

imikejackson:
These functions were flagged by GCOV as never having been executed. This unit test will ensure that they get tested.


nyoungbq (changes requested):
The code itself looks fine, although other places in the code call `Application::DeleteInstance();` towards the end of test so be sure to validate its getting cleaned up by just exiting scope here. That said, the file structuring is not great. In each of the plugin level clone tests you are loading all the plugins each time, which removes the benefit storing them at the plugin level would normally have. This is an application level test so it should be moved to the `simplnx/test` folder and consolidated into one file, since each instance already runs every filter from every plugin. Lastly, this is optional as it is more personal preference to prefer existing over new, but there are already application level test files these cases could be moved into. `FilterValidationTest.cpp` seems like a good candidate to me because it already tests every filters name function in the case at the bottom, and these have a similar structure and intention.

imikejackson:
@nyoungbq Thank you for the recommendations. I rewrote the test cases inside of the FilterValidationTest.cpp file.

nyoungbq (approved):

PR #1536, opened 2026-02-18
ENH: RegularGridSampleSurfaceMesh algorithm updates for speed.

imikejackson:
The speedup is about 10x to 50x.

- [x] Unit Test file updated
- [x] Documentation updated


joeykleingers:
## PR Review: RegularGridSampleSurfaceMesh Algorithm Updates for Speed

### Summary

This PR replaces the previous point-in-polyhedron ray-casting approach (via `SampleSurfaceMesh` base class) with a scanline rasterization algorithm. It also adds support for 1-component face label arrays (part numbers) and non-int32 integer types. The class no longer inherits from `SampleSurfaceMesh`.

### What's Good

- **Significant architecture improvement** — scanline rasterization is fundamentally better suited to this problem than per-voxel ray casting. The O(Z * (T + Y * E)) complexity should be far faster than the previous per-cell approach.
- **Well-structured algorithm** — clear separation into Z-slice, edge computation, and scanline fill phases with good comments.
- **Buffer reuse** — `edges` and `intersections` vectors are reused across iterations to avoid per-slice allocations.
- **Cancellation support** — properly checks `m_ShouldCancel` between slices.
- **Good test coverage** — tests for create-geometry and existing-geometry modes, with multiple data types (int32, uint8) and both FaceLabels (2-comp) and PartNumber (1-comp) inputs. Float32 rejection test is also included.
- **Documentation** updated to reflect the new algorithm.

---

### Action Items

#### Must Fix

- [x] **Clean up face-label reading logic** (`RegularGridSampleSurfaceMesh.cpp:245-251`) — The code reads `label0 = faceLabelsRef[faceIdx]` unconditionally, then overwrites it inside the `numFaceLabelComps == 2` branch. The logic works correctly but is confusing and does a wasted read for 2-component arrays. Restructure to:
  ```cpp
  T label0, label1;
  if(numFaceLabelComps == 2)
  {
    label0 = faceLabelsRef[faceIdx * 2];
    label1 = faceLabelsRef[faceIdx * 2 + 1];
  }
  else
  {
    label0 = faceLabelsRef[faceIdx];
    label1 = T{0};
  }
  ```

- [x] **Remove unrelated `.gitignore` changes** — The diff removes `.claude/` from the ignore list and adds `/CLAUDE.md`, `TODO.md`, `/.claude/settings.local.json`. These are developer-environment changes unrelated to the algorithm and should be a separate commit/PR. Removing `.claude/` from `.gitignore` risks accidentally committing local Claude config in future changes.

- [x] **Remove `resources/claude_code/CLAUDE.md`** — A 279-line CLAUDE.md template is added. This is tooling configuration and should be a separate PR.

- [x] **Fix missing newline at end of `.gitignore`** — The diff shows `\ No newline at end of file`.

#### Should Fix

- [x] **Consider using an integer-only dispatch function** — `ExecuteDataFunctionNoBool` dispatches to float32/float64 in addition to integer types. The preflight correctly restricts to `GetIntegerDataTypes()`, but if someone bypasses preflight, the toggle logic using `==` on floats would be unreliable. Using an integer-only dispatch (if available) would make the algorithm itself type-safe.

#### Nice to Have

- [x] **Parallelize Z-slices** — The old code had per-slice parallelism. The new code processes Z-slices sequentially. Since each Z-slice writes to a distinct region of the output array (`z * cellsPerSlice` offset), the Z-loop could be parallelized with `ParallelTaskAlgorithm` or similar. The `edges` and `intersections` buffers would need to become thread-local.

- [ ] **Scale progress reporting interval** — `if(z % 10 == 0)` reports every 10 slices. For small Z dimensions (e.g., 5 slices) the user only gets one progress message. Consider scaling the interval based on `zDim`, or just reporting every slice since the message handler call is cheap.

joeykleingers (approved):
All checkmarked items have been addressed (verified by Claude Opus 4.6).

PR #1537, opened 2026-02-19
API: Major refactoring of all codes to use a strict Interface/AbstractBaseClass, Concrete style of naming

imikejackson:
## Major Refactoring of the SIMPLNX code base

The point of this PR is to create pure virtual Interfaces, Abstract classes and concreate classes where needed. Extract pure virtual where needed and create the intermediate abstract clases where needed.


imikejackson:
## Phase 15 Summary: Fix I-Prefix Naming Violations

Phase 15 completes the COM-style naming convention cleanup. All `I`-prefixed classes are now **true pure interfaces** — classes that had data members or concrete methods have been renamed to `Abstract*`.

---

### Part A: Renamed 8 I-prefixed non-interfaces → Abstract*

These classes used the `I` prefix but were **not** pure interfaces (they had data members, concrete method bodies, or static helpers):

| Old Name | New Name | Scope |
|----------|----------|-------|
| `IDataCreationAction` | `AbstractDataCreationAction` | 31 files, ~64 occurrences |
| `IDataIO` | `AbstractDataIO` | File rename + class rename |
| `IGeometryIO` | `AbstractGeometryIO` | File rename + class rename |
| `IGridGeometryIO` | `AbstractGridGeometryIO` | File rename + class rename |
| `INodeGeom0dIO` | `AbstractNodeGeom0dIO` | File rename + class rename |
| `INodeGeom1dIO` | `AbstractNodeGeom1dIO` | File rename + class rename |
| `INodeGeom2dIO` | `AbstractNodeGeom2dIO` | File rename + class rename |
| `INodeGeom3dIO` | `AbstractNodeGeom3dIO` | File rename + class rename |

### Part B: Promoted AbstractStringStore → IStringStore

`AbstractStringStore` had **no data members** and **12 pure virtual methods**. Its only concrete methods were trivial iterator/comparison helpers delegating to pure virtuals. By project convention, it qualifies as a pure interface and was renamed to `IStringStore`.

---

### Applied Across Repositories

- **simplnx**: 84 files changed (HDF5 IO classes)
- **FileStore**: 44 files changed (Zarr IO equivalents)
- **Python bindings**: Backwards-compatible alias added for `IDataCreationAction`

### Post-Rebase Fixes

After rebasing onto the latest develop branches, several conflicts were resolved:

- **`StatusCode` → `Result<>`**: The develop branch changed geometry method return types from `StatusCode` to `Result<>`. Updated 4 Abstract geometry headers to match the new interface signatures.
- **Constructor initializer lists**: 7 concrete geometry `.cpp` files had constructor calls reverted to old `I*` names during rebase — fixed back to `Abstract*`.
- **Constant qualifications**: Fixed `INodeGeometry2D::k_SharedEdgeListName` → `INodeGeometry1D::k_SharedEdgeListName` (3 files) and `INodeGeometry3D::k_SharedFacesListName` → `INodeGeometry2D::k_SharedFacesListName` (2 files) — the standalone interfaces don't inherit from each other.
- **`getDataRefAs<IGeometry>`**: Fixed to `getDataRefAs<AbstractGeometry>` in ComputeFeatureSizes (2 files) — `IGeometry` is now a standalone interface, not part of the `AbstractDataObject` hierarchy.
- **`ComputeFeatureSizesFilter`**: Restored `RectGrid` to the allowed geometry types (lost during rebase conflict resolution).
- **`RectGridGeom::checkUpdatedIdsImpl`**: Fixed call from `IGridGeometry::` to `AbstractGridGeometry::`.

imikejackson:
## Complete Summary of Name Changes in This Branch

This branch enforces a strict **COM-style naming convention** across the entire codebase:
- **`I` prefix** = pure virtual interface (no data members, no concrete methods)
- **`Abstract` prefix** = abstract base class (has data members and/or concrete implementations)
- **No prefix** = concrete class

### Classes Renamed (Old → New)

#### Phase 2
| Old Name | New Name |
|----------|----------|
| `SegmentFeatures` | `AbstractSegmentFeatures` |
| `IParallelAlgorithm` | `ParallelAlgorithm` |

#### Phase 3
| Old Name | New Name |
|----------|----------|
| `IFilter` | `AbstractFilter` |

A new pure interface `IFilter` was extracted from the old `IFilter` (which had concrete methods and was therefore an ABC, not an interface).

#### Phase 6
| Old Name | New Name |
|----------|----------|
| `IGeometry` | `AbstractGeometry` |
| `IGridGeometry` | `AbstractGridGeometry` |
| `INodeGeometry0D` | `AbstractNodeGeometry0D` |
| `INodeGeometry1D` | `AbstractNodeGeometry1D` |
| `INodeGeometry2D` | `AbstractNodeGeometry2D` |
| `INodeGeometry3D` | `AbstractNodeGeometry3D` |

New pure interfaces `IGeometry`, `IGridGeometry`, `INodeGeometry0D`–`3D` were extracted from each.

#### Phase 10
| Old Name | New Name |
|----------|----------|
| `IParameter` (near-pure) | `AbstractParameter` (base class portion) |

`IParameter` was promoted to a true pure interface; concrete methods moved to `AbstractParameter`.

#### Phase 11
| Old Name | New Name |
|----------|----------|
| `IArray` | `AbstractArray` |
| `IDataArray` | `AbstractDataArray` |
| `INeighborList` | `AbstractNeighborList` |
| `IArrayThreshold` | `AbstractArrayThreshold` |
| `IJsonPipelineParser` | `AbstractJsonPipelineParser` |

New pure interfaces `IArrayThreshold` and `IJsonPipelineParser` were extracted from the old classes of the same name.

#### Phase 12
| Old Name | New Name |
|----------|----------|
| `IDataIOManager` | `AbstractDataIOManager` |

#### Phase 14
| Old Name | New Name |
|----------|----------|
| `DataObject` | `AbstractDataObject` |

A new pure interface `IDataObject` was extracted.

#### Phase 15
| Old Name | New Name |
|----------|----------|
| `IDataCreationAction` | `AbstractDataCreationAction` |
| `IDataIO` | `AbstractDataIO` |
| `IGeometryIO` | `AbstractGeometryIO` |
| `IGridGeometryIO` | `AbstractGridGeometryIO` |
| `INodeGeom0dIO` | `AbstractNodeGeom0dIO` |
| `INodeGeom1dIO` | `AbstractNodeGeom1dIO` |
| `INodeGeom2dIO` | `AbstractNodeGeom2dIO` |
| `INodeGeom3dIO` | `AbstractNodeGeom3dIO` |
| `AbstractStringStore` | `IStringStore` |

---

### New Pure Interfaces Created

These are brand-new classes extracted during the refactoring:

| Interface | Extracted From | Phase |
|-----------|---------------|-------|
| `IFilter` | old `IFilter` (now `AbstractFilter`) | 3 |
| `ISegmentFeatures` | `SegmentFeatures` (now `AbstractSegmentFeatures`) | 2 |
| `IPipelineNode` | `AbstractPipelineNode` | 4 |
| `IPlugin` | `AbstractPlugin` | 5 |
| `IGeometry` | old `IGeometry` (now `AbstractGeometry`) | 6 |
| `IGridGeometry` | old `IGridGeometry` (now `AbstractGridGeometry`) | 6 |
| `INodeGeometry0D` | old `INodeGeometry0D` (now `AbstractNodeGeometry0D`) | 6 |
| `INodeGeometry1D` | old `INodeGeometry1D` (now `AbstractNodeGeometry1D`) | 6 |
| `INodeGeometry2D` | old `INodeGeometry2D` (now `AbstractNodeGeometry2D`) | 6 |
| `INodeGeometry3D` | old `INodeGeometry3D` (now `AbstractNodeGeometry3D`) | 6 |
| `IDataStructure` | `DataStructure` | 7 |
| `IArrayThreshold` | old `IArrayThreshold` (now `AbstractArrayThreshold`) | 11 |
| `IJsonPipelineParser` | old `IJsonPipelineParser` (now `AbstractJsonPipelineParser`) | 11 |
| `IDataObject` | `DataObject` (now `AbstractDataObject`) | 14 |

`IParameter`, `IStringStore`, `IDataStore`, `IListStore`, and `IDataFactory` were existing classes promoted/confirmed as pure interfaces (Phases 9, 10, 15).

---

### Python Backwards-Compatible Aliases

The following aliases were added to the Python bindings so existing scripts continue to work:

```python
simplnx.DataObject         → simplnx.AbstractDataObject
simplnx.IArray             → simplnx.AbstractArray
simplnx.IDataArray         → simplnx.AbstractDataArray
simplnx.INeighborList      → simplnx.AbstractNeighborList
simplnx.IDataCreationAction → simplnx.AbstractDataCreationAction
```

---

### Dead Code Removed
- `src/simplnx/Utilities/Parsing/JSON.deprecated/` — 15 files, not compiled or referenced anywhere (Phase 13)

### Serialization Compatibility
All renamed geometry classes preserve their original `k_TypeName` string values (e.g., `AbstractGeometry::k_TypeName = "IGeometry"`) to maintain backward compatibility with existing `.dream3d` files.

imikejackson:
Analysis of the requestor's code


  What it describes                                                                                                               
                                                                                                                                  
  The section lays out a COM-inspired plugin architecture where:
                                                                                                                                  
  1. Pure virtual interfaces (structs with = 0 methods) are shared as headers between host apps and plugins — no exported symbols,
   no proprietary types.                                                                                                          
  2. Concrete implementations derive from those interfaces but are not exported — they're internal to the plugin.
  3. A single class factory function (e.g., GetPrgniFace()) is the only exported symbol from each plugin. It constructs the
  concrete class and returns it upcast to the interface via std::dynamic_pointer_cast.
  4. The host app dynamically loads the plugin (via QLibrary/dlopen), resolves the factory symbol by name, calls it, and gets back
   a shared_ptr<Interface> to work through.

  Strengths of the description

  - The motivation is clearly stated: decouple plugins from the host so they can be built and distributed independently, even
  after the host ships.
  - The full roundtrip is shown — interface definition, implementation, factory, and host-side loading/usage — which makes the
  intent unambiguous.
  - The explicit mention of avoiding decorated/mangled names and using def files for explicit linkage shows awareness of the key
  cross-compiler ABI pitfall.

  Things worth noting / potential concerns

  1. extern "C" is missing from the factory. The example declares extern PRGNPI0_EXPORT std::shared_ptr<PrgniFace0>
  GetPrgniFace(); but doesn't use extern "C". Without it, the symbol name is still mangled, which undermines the stated goal of
  compiler-independent symbol resolution. The mention of "def files to alias the exports" is the workaround, but extern "C" is
  simpler and more portable.
  2. Returning std::shared_ptr across DLL boundaries is fragile. If the host and plugin are built with different compilers or even
   different STL versions, the std::shared_ptr layout may differ. COM solves this by returning raw interface pointers and using
  AddRef/Release for reference counting. If true cross-compiler compatibility is a goal, returning a raw pointer (or a COM-style
  IUnknown-like base) and managing lifetime through the interface itself would be safer.
  3. The singleton pattern in the factory (static-like pFilter global) means every call to GetPrgniFace() returns the same
  instance. This may be intentional (single plugin instance) but it prevents multiple independent instances of the same filter,
  which could limit pipeline parallelism.
  4. Virtual interface + no virtual ~Destructor: The example PrgniFace0 struct has no virtual destructor. Destroying through a
  base pointer would be undefined behavior. This is likely just an oversight in the example, but worth calling out since it's a
  classic C++ pitfall.
  5. Qt dependency in the host-side loading code (QLibrary, qDebug): The host example uses Qt for dynamic loading. If the goal is
  maximum portability, abstracting over dlopen/LoadLibrary might be preferable — though Qt may already be a given dependency in
  DREAM3DNX.

  Relationship to the refactoring work already done

  Looking at the memory notes (Phases 1–15), the project has already done extensive work extracting pure interfaces (IFilter,
  IPlugin, IGeometry, IDataStructure, IParameter, IDataStore, etc.) from their concrete implementations. This aligns directly with
   step 1 of the architecture described in the CLAUDE.md — having clean interface headers that can be shared without pulling in
  proprietary types. The ground has been well prepared for this plugin model.

  Summary

  The architecture is sound for the stated goal — it's a proven pattern (essentially COM-lite). The main gaps in the description
  are around ABI safety (extern "C", shared_ptr across boundaries, virtual destructors) which are implementation details that
  would need to be addressed when actually building it out. The existing refactoring work (extracting I* interfaces) has already
  laid most of the foundation for this.

joeykleingers:
I think this is a really important PR and will be worth it in the long run, because it will be great to be able to tell at a glance which classes are interfaces and which are abstract, and it also just makes the whole codebase look more professional, just IMO.

joeykleingers (approved):

mmarineBlueQuartz:
I strongly agree with the renaming scheme, but a couple of the items listed under "potential concerns" sound like glaring red flags. Predominantly among them is the idea of forcing Qt into base simplnx. Things like nxrunner do not and should not require Qt to function. Forcing all of Qt in as a dependency just for easier plugin support seems a bit ham-fisted. The point on undefined behavior on destructor calls is also alarming, and I am not aware of any unit tests we might have that would check for that. The only classes that might trigger that off the top of my head would be in the DataStructure, so it could be important to check that it has detailed enough unit tests.

Although, this comment is based purely on the detailed description of the PR and will give a more complete review after looking over it more.

imikejackson:
@mmarineBlueQuartz This PR is changing scope and is only about the renaming part.

To address your concerns about the original request, Qt's plugin loading was just being used as an example. In no-way-shape-or-form is Qt coming into simplnx. Been down that path once. Not going there again. :-)

src/simplnx/DataStructure/Geometry/IGeometry.cpp
nyoungbq, line 1: since implementation has all been moved to `AbstractGeometry` this file should be removed as it is functionally empty

src/simplnx/DataStructure/Geometry/IGridGeometry.cpp
nyoungbq, line 1: remove file, implementation moved to abstract class

src/simplnx/DataStructure/Geometry/IGridGeometry.hpp
```diff
@@ -2,244 +2,173 @@
 
 #include "simplnx/Common/Array.hpp"
 #include "simplnx/Common/StringLiteral.hpp"
-#include "simplnx/DataStructure/AttributeMatrix.hpp"
-#include "simplnx/DataStructure/Geometry/IGeometry.hpp"
+#include "simplnx/Common/Types.hpp"
+#include "simplnx/simplnx_export.hpp"
 
-#include "simplnx/Common/StringLiteral.hpp"
+#include <optional>
```
nyoungbq, line 8:
```suggestion
```
unused include

JDuffeyBQ:
I'm not necessarily opposed to the renaming though I am skeptical that it will actually aid us in readability/discoverability.

Some new classes were created that alter the existing object hierarchy. Particularly with the geometries having multiple inheritance. We may want to make larger changes to the hierarchy after, but I think it would require more discussion. Therefore I am against those changes. I think this PR should be just renaming.

For python there are some backwards compatibility aliases created but not for all. Additionally how long would we maintain those since this is a breaking change.

Additionally some new classes were created that don't seem to do anything specifically under json parsing?

src/simplnx/DataStructure/Geometry/INodeGeometry0D.cpp
nyoungbq, line 1: remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/DataStructure/Geometry/INodeGeometry1D.cpp
nyoungbq, line 1: remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/DataStructure/Geometry/INodeGeometry1D.hpp
```diff
@@ -13,291 +23,39 @@ class SIMPLNX_EXPORT INodeGeometry1D : public INodeGeometry0D
...
-   */
-  void getEdgeCoordinates(usize edgeId, nonstd::span<Point3Df> coords) const;
-
-  /**
-   * @brief Pure-Virtual
-   * @param recalculate This will allow for skipping execution when an
-   * `ElementDynamicList` exists and recalculate is `false`
-   * @return Result<>
+   * @brief
+   * @return IGeometry::StatusCode
```
nyoungbq, line 38:
```suggestion
   * @return Result<>
```

src/simplnx/DataStructure/Geometry/INodeGeometry2D.cpp
nyoungbq, line 1: remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/DataStructure/Geometry/INodeGeometry3D.cpp
nyoungbq, line 1: remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/DataStructure/IDataStructure.cpp
nyoungbq, line 1: remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/DataStructure/IStringStore.cpp
nyoungbq, line 1: Slightly Confused on this one, Implementations should be in `Abstract` classes rather than `I` classes

src/simplnx/DataStructure/IStringStore.hpp
nyoungbq, line 1: I think that the `Abstract` and `I` concept may be confused on this one

src/simplnx/Filter/IFilter.cpp
nyoungbq, line 1: destructor definition is default so move to header. remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/Filter/IParameter.cpp
nyoungbq, line 1: remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/Pipeline/IPipelineNode.cpp
nyoungbq, line 1: move default destructor to header. remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

src/simplnx/Plugin/IPlugin.cpp
nyoungbq, line 1: move default destructor to header. remove file, no implementation in `I` files. I files are pure virtual so no `.cpp` file needed

nyoungbq (commented):
I poured over everything in the `src/simplnx` folder (aka the main infrastructure) and the renaming scheme looks good for the most part. A few things I noticed though:

- `StringStore` seems a little off compared to the other store counterparts. I think it needs to be split into an `Abstract` and `I` class rather than renaming it to `I` since it breaks the paradigm. I would also be okay leaving it as just an `Abstract` class too
- Many of the `I` implementation files (`.cpp`) need to be removed and destructor defaults moved to header to bring it inline with typical paradigm used around the project
- The comments in the `Geometry` folder need their comments updated to reflect the `StatusCode` -> `Result<>` changes

Edit: After seeing Jareds feedback I also want to offer that I absolutely agree with the statement about the geometry heirarchy. I think it is definitely worth discussing re-architecting the polymorphism. Multiple inheritance incurs a definitive cost with the runtime vtable look up, so we need to make sure that frequently called methods do not incur this cost. Ideally we remove the diamond problem entirely to ensure there is no runtime cost, but I understand this may not be possible and as such a wider discussion is needed.

PR #1540, opened 2026-02-19
REV/ENH: Compute Feature Sizes and Geometry Error Handling

nyoungbq:
- Enabled RectGrid in the Filter
- Updated Geometries to use Result<> instead of status code
- Fixed findElementSizes() bugs in `ImageGeom` and `RectGridGeom`
- Move to Kahan Summation for less variance in RectGeom volume calculation
- Documentation of `ImageGeom` behavior 2D/3D
- Cleaned up redefinition functions for INodeGeom1D
- New test cases
- Extensive Refactoring 
- Logic Cleanup
- Space and Branch Cutting Optimizations
- Restricted Input Typing in Algorithm
- Help text cleanup
- New Preflight Checks
- Extensive Code Documentation
- progress feedback
- cancel checks

### DIFF
--- a/src/Plugins/SimplnxCore/docs/ComputeFeatureSizesFilter.md
+++ b/src/Plugins/SimplnxCore/docs/ComputeFeatureSizesFilter.md
@@ -6,9 +6,24 @@ Statistics (Morphological)
 
 ## Description
 
-This **Filter** calculates the sizes of all **Features**.  The **Filter** simply iterates through all **Elements** querying for the **Feature** that owns them and keeping a tally for each **Feature**.  The tally is then stored as *NumElements* and the *Volume* and *EquivalentDiameter* are also calculated (and stored) by knowing the volume of each **Element**.
+**Note:** Regarding using an *Image Geometry* as input, if you have a `1` in any of the dimensions it will be turned into a 2D calculation instead. See *Image Geometry Additional Considerations* section for more details.
+
+This **Filter** calculates the sizes, volumes/areas, and equivalent diameters of all **Features** in an **Image Geometry** or **Rectilinear Grid Geometry**.
+
+To do so, the **Filter** simply iterates through all **Elements** querying for the **Feature** that owns them and keeping a tally for each **Feature**. The tally is then stored as *NumElements* and the *Volume* and *EquivalentDiameter* are also calculated (and stored) by knowing the volume of each **Element**.
+
+Note here that **Image Geometry** will always be faster than its equivalent **Rectilinear Grid Geometry** because we leverage the uniformity of voxel sizes to preform volume/area calculations at the feature level rather than the cell level.
 
 During the computation of the **Feature** sizes, the size of each individual **Element** is computed and stored in the corresponding **Geometry**. By default, these sizes are deleted after executing the **Filter** to save memory. If you wish to store the **Element** sizes, select the *Generate Missing Element Sizes* option. The sizes will be stored within the **Geometry** definition itself, not as a separate **Attribute Array**.
+
+## Image Geometry Additional Considerations
+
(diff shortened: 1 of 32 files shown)


nyoungbq:
hand verified:
[rect_grid_test.md](https://github.com/user-attachments/files/25425843/rect_grid_test.md)

imikejackson (changes requested):
See conversation section

imikejackson:
## Issues Found

### HIGH Severity

- [x] 1. One-sided floating-point comparisons in tests** -- `ComputeFeatureSizesTest.cpp`

All 18 floating-point assertions in the `Validate*` functions use:
```cpp
REQUIRE((areas.getValue(1) - 22.22f) < epsilon);
```
This is missing `std::abs()`. If the actual value is *less* than expected, the difference is negative and the assertion **always passes**. For example, if `areas.getValue(1)` were `0.0f`, the test would still pass. Should be:
```cpp
REQUIRE(std::abs(areas.getValue(1) - 22.22f) < epsilon);
```

### MEDIUM Severity

- [x] 2. Error message is inverted** -- `ImageGeom.cpp:~171`

The error message says *"Image have more than 2 dimensions greater than 1"* when the actual condition is the opposite: the image has *fewer* than 2 dimensions greater than 1 (i.e., too many empty dimensions). Also has grammar error ("have" -> "has").

- [x] 3. Duplicate error code `-2137`** -- `HexahedralGeom.cpp:450`

`HexahedralGeom::findUnsharedFaces` uses error code `-2137`, which belongs to `TetrahedralGeom::findUnsharedFaces`. Should be `-2537` to follow the HexahedralGeom numbering scheme (-2530 through -2536).

- [x] 4. Copy-paste error in error message** -- `EdgeGeom.cpp:262`

`EdgeGeom::findElementCentroids` error message says `"TetrahedralGeom::{}"` instead of `"EdgeGeom::{}"`.

- [x] 6. No validation of `SaveElementSizes` side effect in tests**

The `SaveElementSizes=true` test variants don't verify the element sizes array was created, and the `false` variants don't verify it was cleaned up.

### LOW Severity

- [x] 7. Redundant `featureIds.getValue()` call** -- `ComputeFeatureSizes.cpp` (`ProcessRectGridGeom`, ~line 218-219)

```cpp
const int32 voxelFeatureId = featureIds.getValue(voxelIdx);
featureVoxelCounts[featureIds.getValue(voxelIdx)]++;  // Should use voxelFeatureId
```

- [x] 8. Duplicate progress message** -- `ComputeFeatureSizes.cpp` (`ProcessImageGeom`)

`"Finding Max Feature Id..."` is sent twice -- the second occurrence (after max is already found) should say something like `"Counting Voxels Per Feature..."`.

- [x] 9. Doxygen copy-paste errors** -- `INodeGeometry1D.hpp`

`findElementsContainingVert` and `findElementNeighbors` doxygen both say *"calculate the sizes of each element"* instead of describing their actual purpose.

- [x] 10. `static_cast<int32>` with `uint32` target** -- `ImageGeom.cpp:~165`

`emptyDimsCount` is declared `uint32` but incremented via `static_cast<int32>`. Should be `static_cast<uint32>`.

- [x] 11. Stale comment in test** -- `ComputeFeatureSizesTest.cpp:~236`

Comment says feature 1 volume is `2358.834` but validation code correctly checks `2362.434`.

- [x] 12. Documentation typos** -- `ComputeFeatureSizesFilter.md`

"preform" -> "perform", "ambigous" -> "ambiguous"

- [x] 13. Comment typos in algorithm** -- `ComputeFeatureSizes.cpp`

"spcaing", "calcualtion", "dentoted" in block comments.

- [x] 14. `parametersVersion()` not incremented** -- `ComputeFeatureSizesFilter.cpp`

Still returns `1` despite parameter label/description/type changes. May need to be `2` for pipeline migration.

- [x] 15. Leftover `uint16_t`** -- `HexahedralGeom.cpp:268`

One `uint16_t` remains where the rest of the file uses the project alias `uint16`.

imikejackson (approved):

PR #1542, opened 2026-02-21
ENH: Optimize various filters for out-of-core data access patterns.

imikejackson:
Replace the stack-based DFS flood-fill algorithm in SegmentFeatures with a two-phase chunk-sequential Connected Component Labeling (CCL) algorithm. This eliminates random-access patterns that cause severe OOC chunk thrashing (22-30x speedup for EBSD/CAxis, 29x for ScalarSegmentFeatures).

Apply the same vector-based data structure optimizations to FillBadData: replace unordered_map provisional labels with dense vector buffer, eliminate backward neighbor reads from OOC featureIdsStore, and replace hash-based region classification with direct vector lookups.

Shared UnionFind utility uses vector storage with path-halving compression and union-by-rank for near-O(1) amortized operations.


imikejackson:
This is now in #1545

PR #1545, opened 2026-02-24
ENH: Algorithm dispatch for OoOC filter optimizations

joeykleingers:
## Summary
- Add reusable `AlgorithmDispatch.hpp` utility (`IsOutOfCore`, `AnyOutOfCore`, `DispatchAlgorithm`) for dispatching between in-core and OOC algorithm implementations at runtime
- Add `ForceOocAlgorithm()` static flag and `ForceOocAlgorithmGuard` RAII class so unit tests can exercise the OOC algorithm path even in in-core builds
- Add documentation (`docs/AlgorithmDispatch.md`) explaining when and how to use algorithm dispatch

## Files Changed
- `src/simplnx/Utilities/AlgorithmDispatch.hpp` — Runtime dispatch utilities
- `docs/AlgorithmDispatch.md` — Usage documentation
- `CMakeLists.txt` — Add header to build

## Test Plan
- [x] Infrastructure is header-only; no runtime behavior changes until filters adopt it
- [x] Builds cleanly on both in-core and OOC configurations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -540,6 +540,7 @@ set(SIMPLNX_HDRS
   ${SIMPLNX_SOURCE_DIR}/Utilities/FilePathGenerator.hpp
   ${SIMPLNX_SOURCE_DIR}/Utilities/ColorTableUtilities.hpp
   ${SIMPLNX_SOURCE_DIR}/Utilities/FileUtilities.hpp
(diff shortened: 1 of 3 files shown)


imikejackson (changes requested):
As we discussed, there needs to be a clear path to exercise both code paths even with out-of-core is not being built. This PR should not be merged until this is figured out.

imikejackson (approved):
Somewhere in there is probably a way to use constexpr to compile out some of the logic but lets get this in.

PR #1546, opened 2026-02-24
ENH: SurfaceNets memory optimization via SoA layout

imikejackson:
## Summary
- Replace dense `Cell` struct array (16 bytes/cell) with Structure-of-Arrays layout (12 bytes/cell), saving ~25% cell map memory
- Replace `Vertex::cellIndex[3]` (12 bytes) with flat index (8 bytes), saving ~33% vertex array memory
- Merge init and setCellVertices into a single O(N³) pass, eliminating a redundant traversal
- Optimize interior cell label lookups with precomputed corner offsets and direct array indexing
- Cache edge crossing count during cell setup, eliminating the counting pass in SurfaceNets.cpp
- Use local vertex buffer during relaxation iterations for better cache locality
- Remove dead code: MMGeometryOBJ, MMGeometryGL, and unused methods (net -566 lines)

## Test plan
- [x] All 4 SurfaceNets tests pass (Default, Smoothing, Winding, Winding Smoothing)
- [x] Output is identical to existing exemplar data (319,447 vertices, 668,786 triangles)
- [ ] Performance comparison on larger dataset (in progress in separate session)

## Performance Comparison NO SMOOTHING

561 x 381 x 528 3D EBSD Stack
Before: 9.46 GB used and 1:14 (MM:SS) filter time
After: 9.20 GB used and 1:06 (MM:SS) filter time

## Smoothing

Aftger takes 12.80 GB (as reported by DREAM3D-NX itself)

This may not really be worth it. The best we could do is have a templated version based on the total number of voxels. If the total is < UINT32_MAX then we could shave down the memory buy a large margin. The issue is the 64 bit clean implementation needs Size_t indices stored which just takes a bunch of memory.

Anyone have any thoughts?


imikejackson:
Not enough memory savings to be worth it. Closing. Needs a complete rewrite

PR #1550, opened 2026-02-25
BUG: Fix memory leak in ITKImageReader

imikejackson:


imikejackson:
PR 1550 Assessment                                                                                                                                          
                                                                                                                                                              
  Title: BUG: Fix memory leak in ITKImageReader                                                                                                               
  State: Open                                                                                                                                                 
  Changes: +5/-322 across 2 files              

  What PR 1550 Does

  1. Adds DisconnectPipeline() in CropImageByVoxelBounds (ReadImageUtils.hpp:200-203) — disconnects the ROI filter's output from the pipeline before returning
   it
  2. Adds DisconnectPipeline() on the reader's output in both operator() overloads (lines 300, 332) — disconnects the reader's output immediately after
  GetOutput()
  3. Deletes DOCUMENTATION_PROGRESS.md — unrelated cleanup (321 lines of a documentation checklist)

  Evaluation Against My Review Findings

  ┌─────────────────────────────────────────────────────┬──────────┬───────────────────┐
  │                        Issue                        │ Severity │ Fixed by PR 1550? │
  ├─────────────────────────────────────────────────────┼──────────┼───────────────────┤
  │ Memory leak in ConvertImageToDataStore 2nd overload │ Critical │ NO                │
  ├─────────────────────────────────────────────────────┼──────────┼───────────────────┤
  │ Missing DisconnectPipeline() calls                  │ Medium   │ YES               │
  ├─────────────────────────────────────────────────────┼──────────┼───────────────────┤
  │ Preflight/execute cropping inconsistency            │ Low      │ No                │
  ├─────────────────────────────────────────────────────┼──────────┼───────────────────┤
  │ Duplicated UUID constants                           │ Low      │ No                │
  └─────────────────────────────────────────────────────┴──────────┴───────────────────┘

  The Critical Gap: ConvertImageToDataStore Leak is NOT Fixed

  The actual ASAN-detected leak is in ITKArrayHelper.hpp:453-454, the type-conversion overload of ConvertImageToDataStore:

  auto* rawBufferPtr = reinterpret_cast<T*>(pixelContainer->GetBufferPointer());
  pixelContainer->ContainerManageMemoryOff();  // Tells ITK: DON'T free this buffer
  // ... copies data via std::copy/std::transform ...
  // rawBufferPtr is NEVER freed → LEAK
  return {};

  PR 1525's new tests exercise this path — there are at least 3 test cases with ChangeDataType = true (lines 416, 472, 529 in ITKImageReaderTest.cpp),
  including one that combines cropping with type conversion. Each test leaks the ITK pixel buffer.

  Compare with the first overload (ITKArrayHelper.hpp:412-415) which correctly handles ownership:

  pixelContainer->ContainerManageMemoryOff();
  std::unique_ptr<T[]> newData(bufferPtr);  // ← Takes ownership, no leak

  Why DisconnectPipeline() Alone Doesn't Fix the Leak

  This project uses ITK 5.4, where DataObject::m_Source is a WeakPointer<ProcessObject> (confirmed at itkDataObject.h:559). Because it's a weak pointer:

  - There is no reference cycle between the filter and its output — the ROI filter is properly destroyed when its SmartPointer goes out of scope regardless of
   DisconnectPipeline()
  - DisconnectPipeline() only clears the weak back-reference and calls Modified() — it has no effect on reference counting or memory deallocation
  - The leak is caused by ContainerManageMemoryOff() telling ITK not to free the buffer, while no one else takes ownership — this is completely independent of
   pipeline management

  What's Needed to Actually Fix the Leak

  The fix for ITKArrayHelper.hpp:438-473 is simple — remove the unnecessary ContainerManageMemoryOff() call on line 454. Since the data is being copied (not
  transferred), ITK should retain ownership and free the buffer normally when the image is destroyed:

     auto* rawBufferPtr = reinterpret_cast<T*>(pixelContainer->GetBufferPointer());
  -  pixelContainer->ContainerManageMemoryOff();
     if constexpr(std::is_same_v<NewStoreT, T>)

  Summary

  PR 1550 adds good-practice DisconnectPipeline() calls that are consistent with the established pattern in ITKArrayHelper.hpp, but it does not fix the
  critical memory leak that is causing the ASAN failures. The actual leak is in the type-conversion overload of ConvertImageToDataStore in ITKArrayHelper.hpp,
   which was not modified by either PR. The deletion of DOCUMENTATION_PROGRESS.md is unrelated cleanup that should ideally be a separate commit.

nyoungbq (approved):

nyoungbq (approved):

PR #1555, opened 2026-03-03
ENH: Again require in-memory data for ITK filters

JDuffeyBQ:
ITK filters can only operate on in-memory data so copying data from out of core storage that may exceed the memory of the computer will generally fail anyways.

### DIFF
--- a/src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Common/ITKArrayHelper.hpp
+++ b/src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Common/ITKArrayHelper.hpp
@@ -312,7 +312,7 @@ std::vector<usize> GetComponentDimensions()
 }
 
 template <class PixelT, uint32 Dimensions>
-typename itk::Image<PixelT, Dimensions>::Pointer WrapDataStoreInImage(AbstractDataStore<UnderlyingType_t<PixelT>>& dataStore, const ImageGeomData& imageGeom)
+typename itk::Image<PixelT, Dimensions>::Pointer WrapDataStoreInImage(DataStore<UnderlyingType_t<PixelT>>& dataStore, const ImageGeomData& imageGeom)
 {
   using T = ITK::UnderlyingType_t<PixelT>;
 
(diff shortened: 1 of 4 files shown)


nyoungbq (approved):

imikejackson:
---
  PR #1555 Review: "ENH: Again require in-memory data for ITK filters"

  Overall: The simplification is sound and correct. Removing the out-of-core code path eliminates the duplicate-copy
  problem and previous source of leaks. A few items need attention before merging.

  ---
  1. Magic number — needs a named constant (blocker)

  ITKArrayHelper.hpp:909
  return MakeErrorResult(-9999, fmt::format("Input Array '{}' utilizes out-of-core data...", ...));
  Every other error code in this file follows the named-constant pattern in nx::core::ITK::Constants
  (k_ImageGeometryDimensionMismatch = -2000, etc.). -9999 is a raw magic number and should become:
  // in namespace Constants
  inline constexpr int32 k_OutOfCoreDataNotSupported = -2002;

  ---
  2. Non-idiomatic empty string check

  ITKArrayHelper.hpp:907
  if(inputArray.getDataFormat() != "")
  Should be:
  if(!inputArray.getDataFormat().empty())

  ---
  3. Naming convention violation — newData should be newDataPtr

  ITKArrayHelper.hpp:383 and 411
  std::unique_ptr<T[]> newData(bufferPtr);
  Per project conventions, local pointer and smart-pointer variables use the Ptr suffix. bufferPtr and rawBufferPtr
  elsewhere follow this correctly. Should be newDataPtr.

  ---
  4. Discarded Result<> in ReadImageUtils.hpp

  ReadImageUtils.hpp:315 (after the PR):
  ITK::ConvertImageToDataStore(dataStore, *outputImage);
  The old call used void ConvertImageToDataStore(image, dataStore). The new call uses the Result<>-returning overload
  ConvertImageToDataStore(dataStore, image) but discards the return value. If Result<> carries [[nodiscard]], this will
  generate a compiler warning. The result should be checked and propagated:
  auto result = ITK::ConvertImageToDataStore(dataStore, *outputImage);
  if(result.invalid())
  {
    return ConvertResult(std::move(result));
  }

  ---
  5. DataCheck() does not validate out-of-core — preflight will always pass

  ITKArrayHelper.hpp:884
  template <class ArrayOptionsT, template <class> class OutputT = detail::DefaultOutput_t>
  Result<OutputActions> DataCheck(...)
  The out-of-core guard is only in Execute(). A user with an out-of-core array will pass preflight and only hit the error
  at execute time. Adding the same check to DataCheck() gives an early, user-friendly preflight error. Consider adding:
  const auto& inputArray = dataStructure.getDataRefAs<IDataArray>(inputArrayPath);
  if(!inputArray.getDataFormat().empty())
  {
    return MakeErrorResult<OutputActions>(Constants::k_OutOfCoreDataNotSupported,
      fmt::format("Input Array '{}' utilizes out-of-core data. This is not supported within ITK filters.",
  inputArrayPath.toString()));
  }

  ---
  6. Documentation — WrapDataStoreInImage and ConvertImageToDataStore are undocumented

  The PR narrows all signatures from AbstractDataStore<T> to DataStore<T>, which is a meaningful precondition tightening.
  Neither WrapDataStoreInImage nor ConvertImageToDataStore have meaningful @brief text, and the change should be
  reflected. Minimal acceptable docs:

  /**
   * @brief Wraps an in-memory DataStore in an ITK ImportImageFilter image,
   *        giving ITK a zero-copy view of the store's data buffer.
   *
   * @pre @p dataStore must be an in-memory DataStore (not out-of-core).
   * @tparam PixelT  ITK pixel type.
   * @tparam Dimensions  Image dimensionality (2 or 3).
   * @param dataStore  In-memory data store whose buffer is imported.
   * @param imageGeom  Geometry providing origin, spacing, and dimensions.
   * @return Smart pointer to the constructed ITK image.
   */

  Similarly for both ConvertImageToDataStore overloads and the new early-out behavior in Execute().

  ---
  No memory leaks found

  The new ConvertImageToDataStore ownership transfer path (ContainerManageMemoryOff() + unique_ptr + move into DataStore)
  is correct and cleaner than the previous copy path. The WrapDataStoreInImage simplification is also correct — the
  ImportImageFilter with false for the last argument does not take ownership of the buffer, so the DataStore remains the
  owner throughout.

imikejackson (approved):

PR #1560, opened 2026-03-05
PERF: AlignSections Filters OoC Optimization

joeykleingers:
## Summary

Adds slice-buffered OOC algorithm paths for the 4 AlignSections filters, using dual-dispatch (Strategy C) to preserve the original in-core code untouched while adding OOC-optimized variants.

### Changes

**Base class (`AlignSections.cpp`)**
- New `AlignSectionsTransferDataOocImpl<T>` dispatched when any cell array is OOC
- Reads each Z-slice into a local buffer, applies the 2D X/Y shift in memory, writes back — eliminates per-tuple chunk thrashing

**AlignSectionsMisorientation**
- New `findShiftsOoc()` buffers 2 adjacent Z-slices of quats, cellPhases, and mask before the convergence loop
- Pre-loads first reference slice, then `std::swap` cur→ref each iteration to avoid re-reading the reference from DataStore
- Eliminates repeated ZarrStore reads during the 7×7 candidate shift grid convergence

**AlignSectionsMutualInformation**
- New `formFeaturesSectionsOoc()` buffers one Z-slice of quats, cellPhases, and mask for the per-slice 2D flood-fill segmentation

**AlignSectionsFeatureCentroid & AlignSectionsListFilter**
- Benefit from transfer phase optimization only (findShifts access patterns are already sequential/trivial)

**Tests**
- All 9 correctness tests now exercise both in-core and OOC algorithm paths via `GENERATE(false, true)` + `ForceOocAlgorithmGuard`
- Benchmark tests (200³ programmatic datasets) left without GENERATE for clean timing

### Benchmark Results (200×200×200)

| Filter | In-Core Before → After | OOC Before → After | OOC Speedup |
|--------|----------------------|-------------------|-------------|
| AlignSectionsMisorientation | 0.74s → 0.79s (~1.0x) | 32.89s → 16.14s | **2.0x** |
| AlignSectionsMutualInformation | 0.49s → 0.52s (~1.0x) | 15.61s → 15.14s | ~1.0x |
| AlignSectionsFeatureCentroid | 0.24s → 0.28s (~1.0x) | 8.41s → 8.82s | ~1.0x |
| AlignSectionsListFilter | 0.22s → 0.25s (~1.0x) | 7.50s → 7.95s | ~1.0x |

### Optimization Ceiling Analysis

The OOC speedups are more modest than Groups B/C/D because the transfer phase dominates OOC runtime and is bottlenecked by ZarrStore's per-element overhead (~55–75ns per `operator[]`: mutex lock/unlock + chunk lookup vs ~1ns for in-core DataStore). The Misorientation filter shows the most benefit (2.0x) because its findShifts convergence loop re-reads the same 2 slices many times — slice buffering plus reference-slice swap (reusing the previous iteration's current-slice buffer as the next iteration's reference) eliminates that repeated I/O.

Further improvement requires deeper OOC infrastructure changes:
1. **Bulk read/write API** on `AbstractDataStore` — eliminates ~47.8M mutex lock/unlock cycles per filter (~1–2s savings)
2. **Chunk-level bulk transfer** in FileCore — bypasses per-element chunk lookup entirely (estimated 3–5x transfer phase improvement)
3. **Larger FIFO cache** or per-array cache isolation — enables parallel OOC array processing

## Test Plan

- [x] All 9 correctness tests pass on in-core build (`simplnx-Rel`)
- [x] All 9 correctness tests pass on OOC build (`simplnx-ooc-Rel`)
- [x] `GENERATE(false, true)` exercises both algorithm paths in both builds
- [x] Benchmark tests confirm zero in-core regression

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMisorientation.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/AlignSectionsMisorientation.cpp
@@ -9,8 +9,6 @@
 
 #include <EbsdLib/LaueOps/LaueOps.h>
 
-#include <iostream>
-
 using namespace nx::core;
 
 // -----------------------------------------------------------------------------
@@ -98,10 +96,6 @@ Result<> AlignSectionsMisorientation::findShifts(std::vector<int64_t>& xShifts, 
     // Loop over the Z Direction
     for(int64_t iter = 1; iter < dims[2]; iter++)
     {
-      if(m_ShouldCancel)
-      {
-        return {};
-      }
       throttledMessenger.sendThrottledMessage([&]() { return fmt::format("Determining Shifts || {:.2f}% Complete", CalculatePercentComplete(iter, dims[2])); });
       if(getCancel())
       {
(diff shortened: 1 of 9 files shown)


joeykleingers:

PR #1564, opened 2026-03-17
ENH: Add bulk getValues/setValues API to AbstractDataStore

joeykleingers:
## Summary

- Adds `getValues(startIndex, span)` and `setValues(startIndex, span)` virtual methods to `AbstractDataStore<T>` for bulk read/write of contiguous element ranges
- `DataStore<T>` overrides use `std::memcpy` for maximum in-memory throughput
- Rewrites `fill()`, `copy()`, `copyFrom()`, `setTuple()`, and `fillTuple()` to use the bulk API internally so all existing callers benefit automatically

**Companion PR:** https://github.com/BlueQuartzSoftware/FileStore/pull/4

## Motivation

Per-element OOC access through `ZarrStore::operator[]` costs ~50-100ns each even when the chunk is cached (virtual dispatch -> mutex -> N-D position calc -> 6-slot FIFO cache search -> element access -> mutex unlock). Filter-level optimizations have minimized the number of accesses, but the per-access cost is fixed by the storage layer.

The bulk API allows `ZarrStore` (in the FileStore plugin) to override with chunk-aware implementations that use `memcpy` per chunk-row instead of per element, with three optimization levels:
1. **Per-row memcpy** — one memcpy per contiguous chunk-row (~50x reduction for 50^3 chunks)
2. **Chunk-sticky** — holds chunk `shared_ptr` across rows, avoids redundant cache lookups (~2500x fewer lookups per chunk)
3. **Multi-row extension** — merges rows when chunks cover full fast dimensions (up to entire chunk in one memcpy)

## Implementation Details

- Default `getValues`/`setValues` on `AbstractDataStore` loop over `getValue`/`setValue` for backward compatibility with any subclass that doesn't override
- Uses `std::make_unique<T[]>` for intermediate buffers to avoid the `std::vector<bool>` bit-packing issue

## Test Plan

- [x] `DataStore Bulk getValues/setValues` — full array, partial range, boundary, roundtrip, empty span, out-of-range, single element (7 sections)
- [x] `DataStore Bulk fill` / `copy` / `copyFrom` / `setTuple` / `fillTuple`
- [x] `DataStore Cross-API Roundtrip` — setValue->getValues, setValues->getValue, partial writes
- [x] `DataStore Bulk getValues with multi-component` — multi-dim tuples + float32
- [x] All pre-existing DataStore/DataArray tests pass
- [x] Both in-core and OOC builds pass (19/19 DataStore tests on each)

### DIFF
--- a/src/simplnx/DataStructure/AbstractDataStore.hpp
+++ b/src/simplnx/DataStructure/AbstractDataStore.hpp
@@ -10,6 +10,7 @@
 #include <algorithm>
 #include <compare>
 #include <iterator>
+#include <memory>
 #include <vector>
 
 namespace nx::core
@@ -411,6 +412,44 @@ public:
   virtual void setValue(usize index, value_type value) = 0;
 
   /**
+   * @brief Reads a contiguous range of values from the DataStore into a buffer.
+   * Reads buffer.size() elements starting from startIndex.
+   * @param startIndex The first flat element index to read from
+   * @param buffer Span to write values into; buffer.size() determines count
+   */
+  virtual void getValues(usize startIndex, nonstd::span<T> buffer) const
+  {
+    const usize count = buffer.size();
+    if(startIndex + count > getSize())
+    {
+      throw std::out_of_range(fmt::format("AbstractDataStore::getValues: range [{}, {}) exceeds size {}", startIndex, startIndex + count, getSize()));
+    }
+    for(usize i = 0; i < count; i++)
+    {
+      buffer[i] = getValue(startIndex + i);
+    }
+  }
+
+  /**
+   * @brief Writes a contiguous range of values from a buffer into the DataStore.
+   * Writes buffer.size() elements starting at startIndex.
+   * @param startIndex The first flat element index to write to
+   * @param buffer Span of values to write; buffer.size() determines count
+   */
+  virtual void setValues(usize startIndex, nonstd::span<const T> buffer)
+  {
+    const usize count = buffer.size();
+    if(startIndex + count > getSize())
+    {
+      throw std::out_of_range(fmt::format("AbstractDataStore::setValues: range [{}, {}) exceeds size {}", startIndex, startIndex + count, getSize()));
+    }
+    for(usize i = 0; i < count; i++)
+    {
+      setValue(startIndex + i, buffer[i]);
+    }
+  }
+
+  /**
    * @brief Returns the value found at the specified index of the DataStore.
    * This cannot be used to edit the value found at the specified index.
    * @param index
@@ -584,7 +623,16 @@ public:
    */
   virtual void fill(value_type value)
   {
-    std::fill(begin(), end(), value);
+    const usize totalSize = getSize();
+    constexpr usize k_BulkBufferSize = 8192;
+    const usize bufSize = (std::min)(totalSize, k_BulkBufferSize);
+    auto buffer = std::make_unique<T[]>(bufSize);
+    std::fill_n(buffer.get(), bufSize, value);
+    for(usize offset = 0; offset < totalSize; offset += bufSize)
+    {
+      const usize batchSize = (std::min)(bufSize, totalSize - offset);
+      setValues(offset, nonstd::span<const T>(buffer.get(), batchSize));
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/simplnx/DataStructure/AbstractDataStore.hpp
```diff
@@ -410,6 +411,44 @@ class AbstractDataStore : public IDataStore
    */
   virtual void setValue(usize index, value_type value) = 0;
 
+  /**
+   * @brief Reads a contiguous range of values from the DataStore into a buffer.
+   * Reads buffer.size() elements starting from startIndex.
+   * @param startIndex The first flat element index to read from
+   * @param buffer Span to write values into; buffer.size() determines count
+   */
+  virtual void getValues(usize startIndex, nonstd::span<T> buffer) const
```
JDuffeyBQ: Since the whole point of these functions is to avoid calling `getValue`/`setValue` I think they should be pure virtual. The store class itself knows the optimal way to implement these.

src/simplnx/DataStructure/DataStore.hpp
```diff
@@ -505,6 +505,45 @@ class DataStore : public AbstractDataStore<T>
...
+   * @param buffer Span to write values into
+   */
+  void getValues(usize startIndex, nonstd::span<T> buffer) const override
+  {
+    const usize count = buffer.size();
+    if(startIndex + count > this->getSize())
+    {
+      throw std::out_of_range(fmt::format("DataStore::getValues: range [{}, {}) exceeds size {}", startIndex, startIndex + count, this->getSize()));
+    }
+    std::memcpy(buffer.data(), m_Data.get() + startIndex, count * sizeof(T));
```
JDuffeyBQ: Prefer `std::copy` here.

src/simplnx/DataStructure/DataStore.hpp
```diff
@@ -505,6 +505,45 @@ class DataStore : public AbstractDataStore<T>
...
+   * @param buffer Span of values to write
+   */
+  void setValues(usize startIndex, nonstd::span<const T> buffer) override
+  {
+    const usize count = buffer.size();
+    if(startIndex + count > this->getSize())
+    {
+      throw std::out_of_range(fmt::format("DataStore::setValues: range [{}, {}) exceeds size {}", startIndex, startIndex + count, this->getSize()));
+    }
+    std::memcpy(m_Data.get() + startIndex, buffer.data(), count * sizeof(T));
```
JDuffeyBQ: Prefer `std::copy` here.

src/simplnx/DataStructure/AbstractDataStore.hpp
```diff
@@ -693,10 +766,9 @@ class AbstractDataStore : public IDataStore
   {
     usize numComponents = getNumberOfComponents();
     index_type offset = tupleIndex * numComponents;
-    for(usize i = 0; i < numComponents; i++)
-    {
-      setValue(offset + i, value);
-    }
+    auto tupleValues = std::make_unique<T[]>(numComponents);
```
JDuffeyBQ: The number of components is small. This should be a `std::array`.

src/simplnx/DataStructure/AbstractDataStore.hpp
```diff
@@ -584,7 +623,16 @@ class AbstractDataStore : public IDataStore
    */
   virtual void fill(value_type value)
   {
-    std::fill(begin(), end(), value);
+    const usize totalSize = getSize();
```
JDuffeyBQ: I feel like the this specific kind of logic in `copy` and `fill` belongs inside the store. These functions are already virtual so it's easy to override them.

src/simplnx/DataStructure/AbstractDataStore.hpp
```diff
@@ -677,10 +735,25 @@ class AbstractDataStore : public IDataStore
                                          totalSrcTuples * sourceNumComponents, destTupleOffset * numComponents, getSize()));
     }
 
-    auto srcBegin = source.begin() + (srcTupleOffset * sourceNumComponents);
-    auto srcEnd = srcBegin + (totalSrcTuples * sourceNumComponents);
-    auto dstBegin = begin() + (destTupleOffset * numComponents);
-    std::copy(srcBegin, srcEnd, dstBegin);
+    const usize totalElements = totalSrcTuples * sourceNumComponents;
```
JDuffeyBQ: Similar to the other functions I think that the store should be able to control the details of copying. Maybe have a `virtual copyFromUncheckedImpl()` or something equivalent. That way stores don't have repeat the bounds checks but can control the actual data transfer.

JDuffeyBQ (changes requested):

JDuffeyBQ (commented):
Changes look good. One thing is that it looks a bit odd at first to see a `getValues()` call with no return value. Maybe we could change the function names to make that a bit more obvious. Something like `getValuesIntoBuffer()`. Not a major issue but I think most of our other functions named with "get" return something so it would help differentiate them.

imikejackson:
> Changes look good. One thing is that it looks a bit odd at first to see a `getValues()` call with no return value. Maybe we could change the function names to make that a bit more obvious. Something like `getValuesIntoBuffer()`. Not a major issue but I think most of our other functions named with "get" return something so it would help differentiate them.

@joeykleingers @JDuffeyBQ 
How about 'copyInto()' or 'fillBuffer()'

imikejackson (approved):

imikejackson:
copyIntoBuffer
copyFromBuffer

imikejackson:
An entirely new rewrite of the OoC is coming that deprecates this entire PR.

PR #1569, opened 2026-03-25
BUG/REV: Compute Feature Neighbors

nyoungbq:
- Patched Major bug in calculation of Shared Surface Area List (Present in 6.5)
- Added 32 new test cases covering 3D/2D/1D/0D
- Hand validated output for each new test case
- Algorithm restructuring to reduce time complexity and branching in tight loops
- deepest nested loop reduction for non 3D cases
- extensive conversion of runtime branching to compile time constexpr
- assorted optimizations following profiling
- code documentation

3D Hand Validation attached
[compute_feature_neighbors_3D_ test.txt](https://github.com/user-attachments/files/26244915/compute_feature_neighbors_3D_.test.txt)

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureNeighbors.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureNeighbors.cpp
@@ -96,6 +96,10 @@ struct ComputeFeatureNeighborsFunctor
      * faces, but the segmentation done here would make it far less readable and in the greater context
      * the speed gain is minimal considering they are O(n-2) and O(1) respectively and the greater algorithm
      * is 0(6(n-2)^3).
+     *
+     * Note here that discussions were had of adding Kahan Summation for calculating the surface
+     * areas, but was decided against to conserve memory. At least until the issue presents itself
+     * in a real world dataset.
      */
     const std::array<FaceNeighborType, 6> faceNeighborInternalIdx = initializeFaceNeighborInternalIdx();
     const std::function<void(int64, int64, int64)> processFrameCell = [&](const int64 zIndex, const int64 yIndex, const int64 xIndex) -> void {
@@ -105,7 +109,7 @@ struct ComputeFeatureNeighborsFunctor
       const int32 feature = featureIds.getValue(voxelIndex);
       if(feature > 0)
       {
-        if constexpr(ProcessSurfaceFeaturesV)
+        if constexpr(ProcessSurfaceFeaturesV && !Is1DImageDimsState<ImageDimensionStateT>())
         {
           surfaceFeatures->setValue(feature, true);
         }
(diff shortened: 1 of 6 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureNeighbors.cpp
```diff
@@ -1,223 +1,617 @@
...
+     * faces, but the segmentation done here would make it far less readable and in the greater context
+     * the speed gain is minimal considering they are O(n-2) and O(1) respectively and the greater algorithm
+     * is 0(6(n-2)^3).
+     *
+     * Note here that discussions were had of adding Kahan Summation for calculating the surface
+     * areas, but was decided against to conserve memory. At least until the issue presents itself
+     * in a real world dataset.
+     */
+    const std::array<FaceNeighborType, 6> faceNeighborInternalIdx = initializeFaceNeighborInternalIdx();
+    const std::function<void(int64, int64, int64)> processFrameCell = [&](const int64 zIndex, const int64 yIndex, const int64 xIndex) -> void {
```
JDuffeyBQ:
```suggestion
    const auto processFrameCell = [&](const int64 zIndex, const int64 yIndex, const int64 xIndex) -> void {
```
It looks like this functor is never used in a way that needs the type erased behavior of `std::function` so we can just make it `auto` to use the exact lambda type instead. This may not have a significant effect partly because the optimizer may be able to devirtualize the calls. But since `std::function` is not needed anyways I think it's a good change.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureNeighbors.cpp
```diff
@@ -1,223 +1,617 @@
...
+{
+  return (ActualT::HasEmptyXDim == true && ActualT::HasEmptyYDim == true && ActualT::HasEmptyZDim == false) ||
+         (ActualT::HasEmptyXDim == true && ActualT::HasEmptyYDim == false && ActualT::HasEmptyZDim == true) ||
+         (ActualT::HasEmptyXDim == false && ActualT::HasEmptyYDim == true && ActualT::HasEmptyZDim == true);
+}
 
-// -----------------------------------------------------------------------------
-Result<> ComputeFeatureNeighbors::operator()()
+template <class ActualT>
+constexpr bool Is2DImageDimsState()
```
JDuffeyBQ: I think these could be moved into the `ImageDimensionsState` struct.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureNeighbors.cpp
```diff
@@ -1,223 +1,617 @@
...
-    onsurf = 0;
-    feature = featureIds[voxelIndex];
-    if(feature > 0 && feature < neighborlist.size())
+    // Process Planes for 2D and 3D (Stack) Images
+    if constexpr(!Is1DImageDimsState<ImageDimensionStateT>() && !IsExpectedImageDimsState<ImageDimensionStateT, SingleVoxelImage>())
     {
-      int64 xIdx = voxelIndex % dims[0];
-      int64 yIdx = (voxelIndex / dims[0]) % dims[1];
-      int64 zIdx = voxelIndex / (dims[0] * dims[1]);
+      const std::function<void(int64, int64, int64, std::vector<FaceNeighborType>&)> processFaceCell = [&](const int64 zIndex, const int64 yIndex, const int64 xIndex,
```
JDuffeyBQ:
```suggestion
      const auto processFaceCell = [&](const int64 zIndex, const int64 yIndex, const int64 xIndex,
```
Same as before.

JDuffeyBQ (commented):
Looks good. Just a couple of minor suggestions.

JDuffeyBQ (approved):

PR #1572, opened 2026-03-31
ENH: Convert pipelines to python scripts

imikejackson:
This lays down infrastructure that allows users to convert a .d3dpipeline file into a fully functioning python script.


wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+        lines: list[str] = []
+        lines.append(f"# Filter {index}: {human_name}")
+
+        # Separate setup lines from inline kwargs
+        setup_lines: list[str] = []
+        kwargs: list[tuple[str, str]] = []
+
+        for arg_name in sorted(args.keys()):
+            value = args[arg_name]
+            codec = self._registry.find(value)
```
JDuffeyBQ: We know the exact parameter since we have the filter so we should just use a mapping of filter parameter to python conversion.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+    "simplnx": "nx",
+    "orientationanalysis": "nxor",
+    "itkimageprocessing": "nxitk",
+}
+
+MODULE_IMPORTS: dict[str, str] = {
+    "simplnx": "import simplnx as nx",
+    "orientationanalysis": "import orientationanalysis as nxor",
+    "itkimageprocessing": "import itkimageprocessing as nxitk",
+}
```
JDuffeyBQ: This seems redundant given that an import statement can be constructed from the above `MODULE_ALIASES`.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+def _encode_datapath(value: Any) -> str:
+    """Produce e.g. 'nx.DataPath("DataContainer/Cell Data")'."""
+    path_str = value.to_string("/")
+    # Use repr() for the string to safely handle embedded quotes
+    return f"nx.DataPath({repr(path_str)})"
+
+
+def _encode_simple_value(value: Any) -> str:
+    """Encode a single simple value to a Python expression string."""
+    import pathlib
```
JDuffeyBQ: This should be with the other imports.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+Usage:
+    import simplnx as nx
+    from simplnx_utilities import create_default_generator
+
+    pipeline = nx.Pipeline.from_file("path/to/pipeline.d3dpipeline")
+    generator = create_default_generator()
+    print(generator.generate(pipeline))
+"""
+
+from __future__ import annotations
```
JDuffeyBQ: I believe this is only for deferred evaluation of annotations but those don't seem to be used? Also we should be importing the simplnx modules. And some type annotations can be updated from `Any`.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+    registry.register(CalculatorValueCodec())
+    registry.register(Dream3dImportCodec())
+    registry.register(ReadHDF5DatasetCodec())
+    registry.register(OEMEbsdScanCodec())
+    registry.register(CropGeometryCodec())
+    registry.register(DynamicTableCodec())
+    return registry
+
+
+def create_default_generator() -> PipelineCodeGenerator:
```
JDuffeyBQ: The `PipelineCodeGenerator` doesn't change so I think it would be easier if you could call this from a function something like `generate_python_pipeline()` using a constant version of the generator.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+# ---------------------------------------------------------------------------
+# PipelineCodeGenerator — the orchestrator
+# ---------------------------------------------------------------------------
+
+class PipelineCodeGenerator:
+    """Converts Pipeline/PipelineFilter objects to Python source code."""
+
+    def __init__(self, registry: CodecRegistry | None = None) -> None:
+        self._registry = registry or _create_default_registry()
+        self._resolver = FilterModuleResolver()
```
JDuffeyBQ: This implies that we could have a different registry but I don't think we have a reason for that. Similar thing for the resolver which doesn't even hold state anyways.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+        for mod in sorted(needed_modules):
+            if mod not in MODULE_IMPORTS:
+                import_lines.append(f"import {mod}")
+        sections.append("\n".join(import_lines))
+        
+        # Append the check_filter_result() function
+        sections.append("\n".join([
+            "import os",
+            "import shutil",
+            "from pathlib import Path",
```
JDuffeyBQ: What code requires these imports?

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+            "",
+            "def check_filter_result(filter: nx.IFilter, result: nx.IFilter.ExecuteResult) -> None:",
```
JDuffeyBQ: We talked about moving this function into the bindings module directly. Not difficult just wasn't a priority before.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
...
+        self._resolver = FilterModuleResolver()
+
+    # -- public API --
+
+    def generate(self, pipeline: Any) -> str:
+        """Full runnable script from a Pipeline object."""
+        filters = [pipeline[i] for i in range(len(pipeline))]
+        return self._generate_full(filters)
+
+    def generate_filters(self, filters: list[Any]) -> str:
```
JDuffeyBQ: These functions specifically want a `PipelineFilter` object but I think it should support instances of `IFilter` which the pipeline filter can just hand to that function. I think it would be to have a function to generate a single filter as well.

wrapping/python/CMakeLists.txt
```diff
@@ -199,13 +199,24 @@ if(SIMPLNX_PY_ENABLE_INSTALL)
...
   endif()
 endif()
 
+message(STATUS "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Bin/simplnx_utilities.py")
+#-------------------------------------------------------------------------------
+# Copy the Pipeline to Python Code utilities
+#-------------------------------------------------------------------------------
+file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/utils/simplnx_utilities.py 
+      DESTINATION "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/."
+    )
```
JDuffeyBQ: This won't copy to the correct place on the Visual Studio generator.

wrapping/python/CMakeLists.txt
```diff
@@ -199,13 +199,24 @@ if(SIMPLNX_PY_ENABLE_INSTALL)
     LIBRARY
       DESTINATION ${SIMPLNX_PY_INSTALL_DIR}
   )
+  install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/utils/simplnx_utilities.py                                                                      
+    DESTINATION ${SIMPLNX_PY_INSTALL_DIR}                                                                                                   
+  )
```
JDuffeyBQ: Should make sure this is good for conda builds.

wrapping/python/utils/simplnx_utilities.py
```diff
@@ -0,0 +1,632 @@
+"""Simplnx utilities for converting live Pipeline objects to runnable Python scripts.
+
+Usage:
+    import simplnx as nx
+    from simplnx_utilities import create_default_generator
+
+    pipeline = nx.Pipeline.from_file("path/to/pipeline.d3dpipeline")
+    generator = create_default_generator()
+    print(generator.generate(pipeline))
```
JDuffeyBQ: We should a test to ensure that we generate valid code ideally. Minimum is that the script runs.

JDuffeyBQ (changes requested):

PR #1582, opened 2026-04-07
ENH: Add missing cancel checks to lots of filters

imikejackson:


nyoungbq (approved):
All looks good, I was torn on the AlignGeometries changes on whether that cancel should be in the functions within the individual switch cases at the top level loop, but I do think that the loops would possibly suffer a minor slowdown and a single walk and read with no serious calculations shouldn't be too long for in-core applications so I am okay leaving it as is.

Side Note: Out of the scope of this PR, but align geometries could really use some user feedback around those functions. Not progress feedback in the loops in the function itself since that is basically a worse case senario for how much of the workload would be messaging, but notices for "Finding moving Centroid...", "Finding target Centroid...", "Beginning translation...", etc.

PR #1591, opened 2026-04-15
ENH: Add Multiple Output Types for Sample Surface Mesh Filters

nyoungbq:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


imikejackson (changes requested):
# Review — ENH: Add Multiple Output Types for Sample Surface Mesh Filters

## Overview
Adds user-selectable output integer type for the `Feature Ids` array in `RegularGridSampleSurfaceMeshFilter` and `UncertainRegularGridSampleSurfaceMeshFilter`. Templates the shared `SampleSurfaceMesh` utility on the output type, and loosens the Face Labels input type restriction from `int32` to any integer type. Also adds `GetNumericDataTypesAsHumanStrings()` / `GetIntegerDataTypesAsHumanStrings()` helpers.

## Critical Issues

### 1. Runtime crash when `k_UseCustomOutputType_Key = true` with mismatched types (RegularGrid)
The filter's preflight now creates the Feature Ids array with `outputType` (chosen by the user), which can differ from the Face Labels type. However, `Algorithms/RegularGridSampleSurfaceMesh.cpp:372` still dispatches on the **Face Labels** type:

```cpp
ExecuteDataFunctionIntType(ZSliceFunctor{}, ...getDataType(), ...);
```

and `RegularGridSampleSurfaceMesh.hpp:61` does:

```cpp
auto& featureIdsRef = m_DataStructure.getDataRefAs<DataArray<T>>(...FeatureIdsArrayPath).getDataStoreRef();
```

where `T` is the face labels type. Since `getDataRefAs` uses `dynamic_cast<T&>`, this will throw `std::bad_cast` whenever the user selects a custom output type different from the Face Labels type. The feature is fundamentally broken for the RegularGrid filter unless the two types happen to match.

Fix options: (a) dispatch on **both** types so the slice buffer writes convert correctly to the output type, or (b) restrict the user to a compatible output type in preflight.

### 2. No unit tests for the new parameters
- `RegularGridSampleSurfaceMeshTest.cpp` covers face-labels-typed paths but never sets `k_UseCustomOutputType_Key = true` nor exercises a `k_OutputType_Key` value different from the face labels type. The bug above is not caught by any existing test.
- `UncertainRegularGridSampleSurfaceMeshTest.cpp` does not set `k_OutputType_Key` at all.

Both tests need sections that: enable the custom/non-default output type; verify the resulting array's `DataType`; ensure the values are preserved correctly.

### 3. Silent narrowing on `static_cast<T>(iter)` (Uncertain path)
In `SampleSurfaceMesh.cpp:62, 123`, `iter` is a `usize` feature index. Casting into `T = int8/uint8` silently truncates once you have >127/255 features. Preflight should either reject narrow types or emit a warning when the feature count may exceed the type's range. Before this PR, `int32` made this essentially unreachable.

## Style / Correctness

- `RegularGridSampleSurfaceMeshFilter.cpp:4` — `#include \"SurfaceNetsFilter.hpp\"` is unused. Remove.
- `RegularGridSampleSurfaceMeshFilter.cpp:204` — `auto& faceLabelsArray` should be `const auto&` (read-only, and the surrounding method is `const`-qualified via `preflightImpl`).
- `UncertainRegularGridSampleSurfaceMeshFilter.cpp:107` — comment says `k_OutputType_Key = \"output_type\"`; actual key is `\"output_type_index\"`. Minor doc inconsistency.
- `TypesUtility.hpp` — the new `GetNumericDataTypesAsHumanStrings()` is added and never used in this PR; consider deferring until a caller lands.

## Performance

- `SampleSurfaceMesh.cpp:293, 301` — `ExecuteParallelFunction<::SampleSurfaceMeshImpl>` uses the default `ArrayUseAllTypes`, instantiating the impl for `bool`, `float32`, `float64` that the preflight now forbids. Pass `ArrayUseIntegerTypes` (defined in `Types.hpp:28`) to cut 3 dead template instantiations per impl (≈6 total), reducing compile time and binary size.
- No measurable runtime regression expected: `AbstractDataStore<T>&` has the same virtual `operator[]` cost as the prior `Int32AbstractDataStore&` (they're the same type).

## Memory / Thread Safety

- No new leaks: the `IDataArray&` → `AbstractDataStore<T>&` conversion is by-reference with no allocation.
- Parallel writes to `m_PolyIds[i]` from multiple threads were already present before this PR, so this PR doesn't worsen the known \"DataArray is not thread-safe for concurrent writes\" concern — but since you're already in this code, it may be worth converting the parallel workers to write into a thread-local buffer and then do a single-threaded copy-out, matching the pattern used in the new `RegularGridSampleSurfaceMesh::sendThreadSafeSliceUpdate`.

## Backwards Compatibility

- `BackwardsCompatibilityTest.cpp` additions correctly add the two new keys to the ignore list; version bumps in both filters look right (3→4 and 1→2). Default values preserve prior behavior. ✅

## Recommendation

The critical issue (#1) is a latent runtime crash whenever the new \"Use Custom Output Type\" feature is actually used on the RegularGrid filter with a non-matching type. The lack of tests exercising the new parameter is what allowed #1 to slip through — adding those tests will block the regression from recurring. Items #2 and #3 should be addressed before merge; the style/perf items are lower priority.

imikejackson (approved):

PR #1592, opened 2026-04-17
ENH: Parallelize Summation Loops in ComputeFeatureSizes

### LINKED ISSUE #490: Find Feature Sizes: Can this be parallelized

nyoungbq:
Fixes: #490 
~5% speed up on test cases
I left the feature array updates unparallelized because the overhead to do it feature by feature would cost more with how few features there are compared to voxels and most of the feature work is I/O to stores

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureSizes.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureSizes.cpp
@@ -35,10 +35,11 @@ using FeatureVolumesT = tbb::combinable<std::vector<float64>>;
 class ImageSummationImpl
 {
 public:
-  ImageSummationImpl(FeatureVoxelCountsT& voxelCounts, const SizeVec3& dims, const Int32AbstractDataStore& featureIds)
+  ImageSummationImpl(FeatureVoxelCountsT& voxelCounts, const SizeVec3& dims, const Int32AbstractDataStore& featureIds, const std::atomic_bool& shouldCancel)
   : m_VoxelCounts(voxelCounts)
   , m_Dims(dims)
   , m_FeatureIds(featureIds)
+  , m_ShouldCancel(shouldCancel)
   {
   }
 
@@ -47,6 +48,10 @@ public:
     std::vector<uint64>& threadLocalVoxelCounts = m_VoxelCounts.local();
     for(usize zIndex = start; zIndex < end; zIndex++)
     {
+      if(m_ShouldCancel)
+      {
+        return;
+      }
       const int64 zStride = m_Dims[0] * m_Dims[1] * zIndex;
       for(usize yIndex = 0; yIndex < m_Dims[1]; yIndex++)
       {
@@ -69,6 +74,7 @@ private:
   FeatureVoxelCountsT& m_VoxelCounts;
   const SizeVec3& m_Dims;
   const Int32AbstractDataStore& m_FeatureIds;
+  const std::atomic_bool& m_ShouldCancel;
 };
 
 Result<> ProcessImageGeom(ImageGeom& imageGeom, Float32AbstractDataStore& volumes, Float32AbstractDataStore& equivalentDiameters, Int32AbstractDataStore& numElements,
@@ -86,7 +92,12 @@ Result<> ProcessImageGeom(ImageGeom& imageGeom, Float32AbstractDataStore& volume
   FeatureVoxelCountsT threadLocalVoxelCounts([numFeatures] { return std::vector<uint64>(numFeatures, 0); });
   ParallelDataAlgorithm dataAlg;
   dataAlg.setRange(0, dims[2]);
-  dataAlg.execute(ImageSummationImpl(threadLocalVoxelCounts, dims, featureIds));
+  dataAlg.execute(ImageSummationImpl(threadLocalVoxelCounts, dims, featureIds, shouldCancel));
+
+  if(shouldCancel)
+  {
+    return {};
+  }
 
   // Reduce thread local feature voxel counts
   threadLocalVoxelCounts.combine_each(
@@ -222,12 +233,14 @@ Result<> ProcessImageGeom(ImageGeom& imageGeom, Float32AbstractDataStore& volume
 class RectGridSummationImpl
 {
 public:
-  RectGridSummationImpl(FeatureVoxelCountsT& voxelCounts, FeatureVolumesT& volumes, const SizeVec3& dims, const Int32AbstractDataStore& featureIds, const Float32AbstractDataStore& elemSizes)
+  RectGridSummationImpl(FeatureVoxelCountsT& voxelCounts, FeatureVolumesT& volumes, const SizeVec3& dims, const Int32AbstractDataStore& featureIds, const Float32AbstractDataStore& elemSizes,
+                        const std::atomic_bool& shouldCancel)
   : m_VoxelCounts(voxelCounts)
   , m_Volumes(volumes)
   , m_Dims(dims)
   , m_FeatureIds(featureIds)
   , m_ElemSizes(elemSizes)
+  , m_ShouldCancel(shouldCancel)
   {
   }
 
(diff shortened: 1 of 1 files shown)


imikejackson:
## Code Review — `ENH: Parallelize Summation Loops in ComputeFeatureSizes`

Reviewed with two lenses: (1) nitpicky-dev-on-a-bad-day and (2) assuming the author is an LLM (CODEX-ish). Also covered perf, memory, docs, and test coverage.

The 5% speedup is nice, but the patch introduces ~130 lines of duplicated scaffolding, silently removes cancel/progress responsiveness in the hot loop, and bumps into a thread-safety rule that `CLAUDE.md` explicitly calls out. Please take a look at the items below before merge.

---

### Thread safety / correctness (blocking)

- [ ] **`AbstractDataStore::getValue()` is documented as NOT thread-safe, even for concurrent reads across different indices.** See project `CLAUDE.md` → *Thread Safety*: *"The subscript operator (operator[]) and other access methods may have internal state or go through virtual function calls that are not safe for concurrent access, even when accessing different indices."* Both `ImageSummationImpl::convert` and `RectGridSummationImpl::convert` call `m_FeatureIds.getValue(voxelIdx)` (and `m_ElemSizes.getValue(...)` for rectgrid) concurrently from worker threads. This is the exact pattern `CLAUDE.md` warns against. Either (a) get a raw/const pointer to the resident buffer up front and index that, (b) guarantee the `AbstractDataStore` in play is always in-memory (`DataStore<T>` not OOC) and add a `DCHECK`/comment justifying why concurrent reads are safe here, or (c) disable parallelization à la `ComputeFeatureCentroids.cpp:185`.
- [x] **Cancel responsiveness regression.** The old loop polled `shouldCancel` every voxel. The new version only checks *after* the whole parallel scan has finished. For a 1024³ dataset, the user can no longer cancel during the hot loop — which is the whole reason they'd want to cancel. Plumb `m_ShouldCancel` into the worker and early-exit via `convert()`.
- [x] **Progress messaging regression.** `throttledMessenger.sendThrottledMessage(...)` inside the voxel loop is gone. The UI will now sit on "Finding Voxel Counts..." with no percentage updates until the entire volume is processed. Either emit a tick from the worker (under mutex) or switch to a `ParallelTaskAlgorithm` pattern that reports per chunk.
- [x] **Parallel Kahan summation is not equivalent to sequential Kahan summation.** Each worker runs its own Kahan accumulator and then `updateStore` does a plain `std::transform(..., std::plus<>{})` across threads — the cross-thread merge is uncompensated. Numerically this is usually still fine, but it will produce bit-different exemplars from the previous serial implementation. Confirm the existing exemplar-based tests still pass bit-exact, or note the tolerance in the PR description.
- [x] **`manager.execute(imageGeom.getDimensions(), ...)` passes a temporary `SizeVec3` by reference and the inner `ImageSummationImpl` stores `const SizeVec3& m_Dims`.** It's safe today because `dataAlg.execute(...)` is synchronous, but it's a lifetime footgun waiting to bite. Store `m_Dims` by value (it's 3 `usize`) or pass `dims` by value into `execute()`.

### Performance

- [x] **Per-call allocations inside the worker.** `ImageSummationImpl::convert` and `RectGridSummationImpl::convert` allocate `std::vector<uint64>(numFeatures)` (and two more float64 vectors for the rectgrid case) *every time TBB hands the functor a range*. With small ranges / work-stealing this is real overhead. Prefer thread-local storage (`enumerable_thread_specific`) or pre-sized per-worker buffers.
- [x] **Single global mutex in `updateStore` serializes all merges.** For `numFeatures` in the 10s of thousands and >4 cores, the merge can dominate. Consider a per-thread collector that reduces in one pass after `execute()` returns, instead of a lock-per-completion.
- [ ] **Slice-based parallelism (`setRange(0, dims[2])`)** leaves you with awful load balance for volumes that are thin in Z (e.g. 2048×2048×8). Consider ranging over the flat voxel count with a grainsize, or fall back to serial when `dims[2]` is small.
- [ ] **`ComputeFeatureCentroids.cpp` sets `setParallelizationEnabled(false)` with a comment that threading overhead eats the gain** for feature-count loops. That same risk exists here for tiny volumes — a ~5% gain on your test case could easily become a regression on small inputs. Worth adding a threshold check.

### Memory

- [ ] **RectGrid worker allocates `3 × numFeatures × 8 bytes` per call**: counts (uint64) + volumes (float64) + compensators (float64). At 1M features that's 24 MB × (number of active chunks). Worth calling out in the PR description or gating.

### Code style / nitpicks

- [x] **Near-identical duplication of `ParallelImageManager` and `ParallelRectGridManager`.** They share the mutex, the store-vector pattern, the `updateStore(std::transform+std::plus)` merge, and the nested functor shape. A small CRTP/templated base or a single `ParallelSummationManager<Impl>` would halve the surface area.
- [x] **`~ParallelImageManager() noexcept = default;`, `~ImageSummationImpl() noexcept = default;`, `~ParallelRectGridManager() noexcept = default;`, `~RectGridSummationImpl() noexcept = default;`** — all four are dead code. The compiler generates these for you.
- [x] **`mutable std::mutex m_Mutex` + `const` `updateStore` that mutates the externally-held vector via a non-const reference member** is cargo-cult const-correctness. Either make `updateStore` non-const (it *does* mutate state) or drop the `mutable`.
- [x] **`convert()` is the codebase convention** (`ConvertData.cpp`, `CreateColorMap.cpp`, etc.) — good, keep.
- [x] **Mixing `int64` and `usize` for indices.** `zStride`/`yStride` are `int64`, but `AbstractDataStore::getValue()` takes `usize`. Use `usize` throughout to avoid implicit sign conversions.
- [ ] **`const Int32AbstractDataStore& m_FeatureIds;` stored as reference** inside the worker. Fine (same lifetime story as `m_Dims`), but again prefer pointer-or-value to match the codebase's usual worker layout.
- [x] **"Manager" suffix** on `ParallelImageManager` / `ParallelRectGridManager` is uninformative. `ParallelImageSummationReducer` (or similar) would describe what they actually do.
- [x] **`explicit` on the single-arg `ParallelImageManager` ctor** is fine; the `ParallelRectGridManager` ctor takes two args so `explicit` is a no-op there. Drop it on the rectgrid one for consistency.

### LLM-author smells (CODEX lens)

- [ ] **Copy-pasted scaffolding instead of a small abstraction.** Two near-identical manager/impl pairs is a classic "LLM extended pattern rather than refactored" signal.
- [ ] **Boilerplate-y `= default` destructors** everywhere (see above).
- [ ] **Cargo-culted const + `mutable`** on `updateStore`.
- [ ] **Storing references to potentially-temporary arguments** (the `SizeVec3` lifetime issue) — an experienced human usually copies small POD-ish types by value.
- [ ] **Silent removal of cancel check and progress messages** without noting the tradeoff in the PR body — looks like "fit the code into the ParallelDataAlgorithm shape" rather than "preserve existing UX semantics while parallelizing." Worth calling out explicitly in the commit message either way.

### Documentation

- [ ] **`docs/ComputeFeatureSizesFilter.md` is not updated.** Not strictly required for a perf change, but a one-line "implementation is now multi-threaded" note under Notes/Caveats would help future triage.

### Unit tests

- [ ] **No test changes.** The existing `ComputeFeatureSizesTest.cpp` (708 lines) *should* still cover both `ProcessImageGeom` and `ProcessRectGridGeom`, but please confirm:
  - [ ] Tests pass in both in-core *and* OOC build configurations (this is the primary way to catch the thread-safety concern above).
  - [ ] Exemplar arrays still compare bit-exact after the parallel Kahan change; if not, widen tolerance explicitly rather than silently.
  - [ ] Consider adding (or re-running) a large-volume stress case that exercises the parallel path with many features — the existing tests are likely small enough that TBB won't actually parallelize.

---

Happy to drop the thread-safety blocker if you can confirm the `AbstractDataStore` concrete types used here are guaranteed in-memory and that concurrent `getValue` is actually safe despite what `CLAUDE.md` says. Otherwise, the safest minimal fix is probably to grab a `const int32*` / `const float32*` to the resident buffer once in `execute()` and index raw pointers from the workers.

imikejackson (changes requested):
See PR comments

imikejackson (approved):

PR #1594, opened 2026-04-21
BUG: Allow uint8 in Mask in InterpolatePointCloudToRegularGrid

### LINKED ISSUE #1481: BUG: Interpolate Point Cloud to Regular Grid does not handle uint8 for the Mask array
Take a look at EBSDSegmentFeatures and the use of MaskCompareUtilities. Also make sure to update the parameter to take UINT8 in addition to BOOL.

nyoungbq:
Fixes: #1481 

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/SimplnxCore/docs/InterpolatePointCloudToRegularGridFilter.md
+++ b/src/Plugins/SimplnxCore/docs/InterpolatePointCloudToRegularGridFilter.md
@@ -71,7 +71,7 @@ Median is not supported because it requires storing every vertex's contribution,
 
 ### Mask
 
-An optional boolean mask array may be provided. Vertices where the mask value is *false* are skipped entirely during interpolation – they do not contribute to any voxel's accumulated value or statistics.
+An optional boolean or uint8 mask array may be provided. Vertices where the mask value is *false* or *0* are skipped entirely during interpolation – they do not contribute to any voxel's accumulated value or statistics.
 
 ### Memory Usage
 
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/InterpolatePointCloudToRegularGrid.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/InterpolatePointCloudToRegularGrid.cpp
@@ -4,6 +4,7 @@
 #include "simplnx/DataStructure/Geometry/ImageGeom.hpp"
 #include "simplnx/DataStructure/Geometry/VertexGeom.hpp"
 #include "simplnx/Utilities/FilterUtilities.hpp"
+#include "simplnx/Utilities/MaskCompareUtilities.hpp"
 
 #include <cmath>
 #include <limits>
@@ -57,8 +58,8 @@ struct WriteWeightedAverageFunctor
 
 void computeKernel(uint64 interpolationTechnique, const std::vector<float32>& sigmas, std::vector<float32>& kernel, const int64 kernelNumVoxels[3])
 {
-  usize kDimX = static_cast<usize>(2 * kernelNumVoxels[0] + 1);
-  usize kDimY = static_cast<usize>(2 * kernelNumVoxels[1] + 1);
+  const auto kDimX = static_cast<usize>(2 * kernelNumVoxels[0] + 1);
+  const auto kDimY = static_cast<usize>(2 * kernelNumVoxels[1] + 1);
 
   for(int64 z = -kernelNumVoxels[2]; z <= kernelNumVoxels[2]; z++)
   {
@@ -66,10 +67,10 @@ void computeKernel(uint64 interpolationTechnique, const std::vector<float32>& si
     {
       for(int64 x = -kernelNumVoxels[0]; x <= kernelNumVoxels[0]; x++)
       {
-        usize kx = static_cast<usize>(x + kernelNumVoxels[0]);
-        usize ky = static_cast<usize>(y + kernelNumVoxels[1]);
-        usize kz = static_cast<usize>(z + kernelNumVoxels[2]);
-        usize idx = kz * kDimY * kDimX + ky * kDimX + kx;
+        const auto kx = static_cast<usize>(x + kernelNumVoxels[0]);
+        const auto ky = static_cast<usize>(y + kernelNumVoxels[1]);
+        const auto kz = static_cast<usize>(z + kernelNumVoxels[2]);
+        const usize idx = kz * kDimY * kDimX + ky * kDimX + kx;
 
         if(interpolationTechnique == InterpolatePointCloudToRegularGrid::k_Uniform)
         {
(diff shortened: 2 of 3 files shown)


imikejackson (changes requested):
# Code Review

## Overview

Closes #1481. The PR does two semantic things plus unrelated cosmetics:

1. **Filter:** broadens `k_InputMaskPath_Key`'s `AllowedTypes` from `{boolean}` to `{boolean, uint8}`.
2. **Algorithm:** replaces the raw `BoolArray::store_type*` mask pointer with a `MaskCompareUtilities::MaskCompare` polymorphic wrapper so both bool and uint8 arrays are handled.
3. **Cosmetics (unrelated):** sprinkles `auto`/`const auto` in place of explicit types across `computeKernel()` and `operator()()`.

---

## Correctness / Bugs

### :red_circle: Catch clause misses the wrong-type exception
`InstantiateMaskCompare` can throw **two** different exceptions:
- `std::out_of_range` — if the path is missing (from `getDataRefAs<IDataArray>`).
- `std::runtime_error` — if the array exists but is neither `bool` nor `uint8` (the `default:` branch in `MaskCompareUtilities.cpp:27`).
- It can also throw `std::bad_cast` from the `dynamic_cast<BoolArray&>` / `UInt8Array&` if the type enum and actual class disagree.

The PR only catches `std::out_of_range` (`InterpolatePointCloudToRegularGrid.cpp:159`). In the same "called outside the normal Preflight/Execute API" scenario the comment justifies, a `runtime_error` will escape and crash the caller. Either widen to `std::exception` or add a second `catch`. (`BadDataNeighborOrientationCheck.cpp:49` has the same latent bug — consistency with existing code, not correctness.)

### :yellow_circle: Redundant `maskCompare != nullptr` check
`InterpolatePointCloudToRegularGrid.cpp:219`:
```cpp
if(m_InputValues->useMask && maskCompare != nullptr && !maskCompare->isTrue(i))
```
If `useMask` is true and we got past the try/catch without returning, `maskCompare` is guaranteed non-null. The old code had the same vestigial null check, so this is a carry-over, but now the logic is clearer and the check really is dead. Either drop the null check, or drop `useMask` (they're equivalent at this point).

---

## Performance

### :yellow_circle: Virtual call in the hot path
Before: `mask->getValue(i)` on `AbstractDataStore<bool>*` — non-virtual, direct.
After: `maskCompare->isTrue(i)` — **virtual** call through `MaskCompare*`.

The `MaskCompareUtilities` header itself warns: *"There is virtual function call overhead with using these structs and functions."*

This call is in the outer vertex loop (once per vertex, not per kernel voxel), so on a typical point cloud it's O(numVerts) virtuals — almost certainly unmeasurable relative to the 3D kernel accumulation. But for a purely mask-excluded run (most vertices rejected early), virtual dispatch is a meaningful fraction of work.

Both `BoolMaskCompare::isTrue` and `UInt8MaskCompare::isTrue` also use `.at()` (bounds-checked) versus the old path's `getValue()`. Worth verifying this doesn't matter; if it does, a templated specialization path would be faster.

### :yellow_circle: `std::unique_ptr` heap allocation
One extra heap allocation per execution — trivially insignificant, just noting it.

---

## Threading / Memory

- No threading change; the outer loop is still serial. No new thread-safety concerns.
- No new persistent allocations besides the one `unique_ptr`. The big `interpSourceData` / `copySourceData` footprint is untouched.

---

## Test Coverage — this is the real gap

`InterpolatePointCloudToRegularGridTest.cpp` is **not updated at all**. The only mask-using test (`Masked Vertices`, line 268) exercises the `bool` path, which already worked. The entire *point* of this PR — the `uint8` path — has zero coverage.

Needed at minimum:
- A `DYNAMIC_SECTION` or new `TEST_CASE` that mirrors "Masked Vertices" but constructs a `UInt8Array` for the mask (with 0/1 values, and also a few non-{0,1} values like 255 to assert "non-zero == true" semantics).
- Optional: a negative test exercising the catch block (construct an array of type `int32`, call the algorithm directly bypassing preflight, expect `-54060`). This is the only way to prove the only new error-handling branch in the algorithm.

`createTestDataStructure` hard-codes `BoolArray::CreateWithStore<DataStore<bool>>` (test file line 68). It needs a parameter or overload to build a uint8 mask.

---

## Documentation

### :yellow_circle: Parameter description is stale
`InterpolatePointCloudToRegularGridFilter.cpp:76`:
```cpp
"DataPath to the boolean mask array. Values that are true will mark that vertex as usable."
```
Still says "boolean" and "true." After this PR the field accepts `uint8`, where semantics are "non-zero = true." Update to something like:
```
"DataPath to a boolean or uint8 mask array. Values that are non-zero will mark that vertex as usable."
```

### :yellow_circle: Markdown doc is stale
`InterpolatePointCloudToRegularGridFilter.md:74`:
> An optional boolean mask array may be provided. Vertices where the mask value is *false* are skipped entirely during interpolation …

Same fix — mention uint8, use "zero/non-zero" wording. Worth adding a one-line caveat that `uint8` values are interpreted as "0 = skip, anything else = include" so users migrating old legacy DREAM3D (SIMPL) masks know what to expect.

No new algorithm complexity is introduced, so **no infographic is warranted** — the existing doc still applies.

---

## Style (nitpicky)

### :yellow_circle: Scope creep: unrelated `auto` refactor
Roughly half the diff has nothing to do with the bug. The PR converts a dozen `usize`/`float64` local variables to `auto`/`const auto`:
- `InterpolatePointCloudToRegularGrid.cpp:60-61, 69-72, 143-145, 242-244, 253, 289`

This is fine in isolation, but:
- **Inconsistent** — some become `const auto`, some plain `auto`, with no obvious rule (line 143-145 get plain `auto`, while 60-61 get `const auto`, even though all five are loop-invariant).
- **Loses information at reading sites** — `usize kDimX` told the reader "this is the size type." With `auto kDimX = static_cast<usize>(...)` the type is still recoverable but the eye has to jump right. Project convention (per `.clang-tidy`) leans toward explicit types except where `auto` avoids a long templated name.
- **Pollutes commit history** — unrelated to the bug. Makes `git blame` on the kernel code point at this PR for reasons unrelated to "why uint8 support."

Recommend either reverting the style churn, or splitting into a separate `STY` commit.

### :yellow_circle: Commit-message typo
The single commit: `convert mak from bool to uint8 or bool in InterpolatePointCloud` — missing prefix (`BUG:`), misspells "mask", and `mak from bool to uint8 or bool` is confusing. The PR title is correct; the commit itself should be reworded on rebase or squashed.

### :green_circle: Error code `-54060`
Unique in the codebase, fits project convention. Good.

---

## Suggested Changes (priority order)

1. **Must-fix:** Add a uint8 mask test (mirror "Masked Vertices" with `UInt8Array`, include a 255-value to prove non-zero semantics).
2. **Must-fix:** Update the parameter help text (`InterpolatePointCloudToRegularGridFilter.cpp:76`) and the markdown doc (`…Filter.md:74`) to mention uint8 and "non-zero = include."
3. **Should-fix:** Broaden the catch to `std::exception` (or add a second catch for `std::runtime_error`/`std::bad_cast`) so the documented "called outside normal API" scenario actually behaves as documented.
4. **Should-fix:** Remove redundant `maskCompare != nullptr` check (or drop the `useMask &&` — either, not both).
5. **Nice-to-have:** Revert the `usize` → `auto` churn, or split into its own `STY` commit.
6. **Nice-to-have:** Reword the commit message to `BUG: Allow uint8 mask in InterpolatePointCloudToRegularGrid` before merge.

---

## Risk Assessment

Low merge risk overall — the filter is serial, the change is a narrow type-domain extension, and the algorithmic core is untouched. The main risk is **silent coverage gap**: shipping a "now supports uint8" PR with only the bool path exercised means the uint8 path could regress later without anyone noticing until a user file trips it.

imikejackson (approved):

PR #1595, opened 2026-04-22
FILT: Add ReadNIfTIFile filter for NIfTI-1 (.nii / .nii.gz) volumes

imikejackson:
* Read single-file NIfTI-1 (magic 'n+1') into an ImageGeom plus a typed DataArray; reject the .hdr/.img pair format (magic 'ni1') with a descriptive error
* Transparent .nii and .nii.gz support via zlib (gzopen/gzread)
* Detect and correct file endianness using sizeof_hdr == 348
* Datatype support: uint8/int8, uint16/int16, uint32/int32, uint64/int64, float32, float64, RGB24 (3 x uint8), RGBA32 (4 x uint8); complex and 128-bit float types rejected
* Orientation from sform/qform when present; fall back to pixdim and zero origin. Warn when the stored affine has a non-trivial rotation since the simplnx ImageGeom is axis-aligned
* Optional scl_slope/scl_inter scaling on read; promote output to float32 when applied, per NIfTI-1 spec skip for RGB24/RGBA32
* Reject 4D+ volumes in this version; leave a clear error for users with fMRI or diffusion time series
* New utility layer (utils/NiftiUtilities.{hpp,cpp}) wrapping the 348-byte header parse so future writers can reuse it
* Round-trip unit tests synthesize .nii and .nii.gz fixtures and cover uint8, int16 (scaled + unscaled), float32, RGB24, RGBA32, plus rejection of 4D and pair-format files

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadNIfTIFile.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadNIfTIFile.cpp
@@ -80,8 +80,8 @@ Result<CropBounds> ComputeCropBounds(const nx::core::nifti::NiftiMetadata& md, c
     auto startIndex = srcImageGeom->getIndex(xBoundPhysical[0], yBoundPhysical[0], zBoundPhysical[0]);
     if(!startIndex.has_value())
     {
-      return MakeErrorResult<CropBounds>(-34740, fmt::format("Could not map starting physical bounds ({}, {}, {}) to a valid voxel index within the NIfTI volume.", xBoundPhysical[0], yBoundPhysical[0],
-                                                             zBoundPhysical[0]));
+      return MakeErrorResult<CropBounds>(
+          -34740, fmt::format("Could not map starting physical bounds ({}, {}, {}) to a valid voxel index within the NIfTI volume.", xBoundPhysical[0], yBoundPhysical[0], zBoundPhysical[0]));
     }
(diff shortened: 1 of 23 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadNIfTIFile.cpp
```diff
@@ -0,0 +1,210 @@
...
+    for(usize i = 0; i < toRead; i++)
+    {
+      NativeT raw = buffer[i];
+      if(byteSwap)
+      {
+        raw = nx::core::byteswap(raw);
+      }
+
+      OutputT outVal;
+      if constexpr(std::is_same_v<OutputT, float32>)
```
nyoungbq: If float64 or double is an acceptable OutputT as well, this block should be adapted to handle both or a separate if case for float64 should be defined.

nyoungbq (commented):
Looks like a solid library wrapping filter. The constexpr variables defined at the top via a bit shift to allow for potential compiler optimization is a nice touch. I assume you thoroughly vetted the tests, so I will approve as soon as the floating point feedback is addressed one way or the other

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ReadNIfTIFile.cpp
imikejackson: Fixed. Good Catch.

nyoungbq (approved):

PR #1604, opened 2026-04-23
[WIP] ENH: Add example pipelines for the ITK filters

imikejackson:
Example Pipelines are being added for more filters.

@joeykleingers I think we need to brainstorm a bit more to figure out how to bit this off in manageable chunks.

- [ ] ITK Image Processing
- [ ] SimplnxCore
- [ ] OrientationAnalysis


imikejackson:
These will be covered by the port of ITK filters to SimplnxCore

PR #1620, opened 2026-05-18
Fix UniversalHermannMauguin dataset parsing in GrainMapper3D reader

nikprabhu:
## Summary

This PR fixes a likely typo/bug in `GrainMapperReader::readPhaseInfo()` (in `GrainMapper3DUtilities.cpp`).

Previously, `phase.UniversalHermannMauguin` was incorrectly assigned using `k_Name` instead of `k_UniversalHermannMauguin`.

As a result, the `UniversalHermannMauguin` field always duplicated the phase name instead of reading the intended HDF5 dataset.

This fix updates the reader to correctly load the `UniversalHermannMauguin` dataset.

---

## Additional Question / Discussion

While tracing the full metadata flow and looking through the test case example, I noticed that this field is currently never written into the SIMPLNX `DataStructure`.

At the moment, only:

* `MaterialName`
* `CrystalStructures`
* `LatticeConstants`

are written into the ensemble arrays.

I wanted to ask whether omitting `UniversalHermannMauguin` was:

* intentional for compatibility/minimal metadata reasons,
* or simply an unfinished implementation.

My understanding is that some information is already reduced when converting the full space group into a Laue group / crystal structure representation. Preserving `UniversalHermannMauguin` would help retain the original crystallographic metadata when translating GrainMapper3D data into DREAM3D/SIMPLNX format.

On the other hand, if the intention was to keep ensemble metadata minimal, then it may also make sense to avoid reading metadata that is never persisted, since that would make the workflow and intent of the code clearer.

For that reason, I have intentionally kept this PR limited to fixing the dataset read itself and have not modified the code to write this metadata into the SIMPLNX `DataStructure`.

If support for writing this metadata is desired, I would be happy to open a follow-up PR to store it as an ensemble `StringArray`, as I have already wrapped my head around the metadata flow through the reader and ensemble creation logic.

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/GrainMapper3DUtilities.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/GrainMapper3DUtilities.cpp
@@ -398,7 +398,7 @@ herr_t GrainMapperReader::readPhaseInfo(hid_t parentId)
       return error;
     }
 
-    error = H5Lite::readStringDataset(phaseGid, Constants::k_Name, phase.UniversalHermannMauguin);
+    error = H5Lite::readStringDataset(phaseGid, Constants::k_UniversalHermannMauguinName, phase.UniversalHermannMauguin);
     if(error < 0)
     {
       return error;
--- a/src/Plugins/OrientationAnalysis/test/ReadGrainMapper3DTest.cpp
+++ b/src/Plugins/OrientationAnalysis/test/ReadGrainMapper3DTest.cpp
@@ -49,7 +49,7 @@ TEST_CASE("OrientationAnalysis::ReadGrainMapper3D:Default_Parameters", "[Orienta
     args.insertOrAssign(ReadGrainMapper3DFilter::k_ReadLabDCT_Key, std::make_any<bool>(true));
     args.insertOrAssign(ReadGrainMapper3DFilter::k_CreatedDCTImageGeometryPath_Key, std::make_any<DataPath>(computedDCTGeometryPath));
     args.insertOrAssign(ReadGrainMapper3DFilter::k_CellAttributeMatrixName_Key, std::make_any<std::string>(k_Cell_Data));
-    args.insertOrAssign(ReadGrainMapper3DFilter::k_CellEnsembleAttributeMatrixName_Key, std::make_any<std::string>(k_EnsembleAttributeMatrix));
+    args.insertOrAssign(ReadGrainMapper3DFilter::k_CellEnsembleAttributeMatrixName_Key, std::make_any<std::string>(k_Cell_Ensemble_Data));
 
     args.insertOrAssign(ReadGrainMapper3DFilter::k_ConvertPhaseToInt32_Key, std::make_any<bool>(true));
     args.insertOrAssign(ReadGrainMapper3DFilter::k_ConvertOrientationData_Key, std::make_any<bool>(true));
@@ -78,6 +78,8 @@ TEST_CASE("OrientationAnalysis::ReadGrainMapper3D:Default_Parameters", "[Orienta
   UnitTest::CompareExemplarToGenerateAttributeMatrix(dataStructure, exemplarDCTGeometryPath.createChildPath(k_Cell_Data), dataStructure, computedDCTGeometryPath.createChildPath(k_Cell_Data));
   UnitTest::CompareExemplarToGenerateAttributeMatrix(dataStructure, exemplarAbsorptionCTGeometryPath.createChildPath(k_Cell_Data), dataStructure,
                                                      computedAbsorptionCTGeometryPath.createChildPath(k_Cell_Data));
+  UnitTest::CompareExemplarToGenerateAttributeMatrix(dataStructure, exemplarDCTGeometryPath.createChildPath(k_Cell_Ensemble_Data), dataStructure,
+                                                     computedDCTGeometryPath.createChildPath(k_Cell_Ensemble_Data));
 
   UnitTest::CheckArraysInheritTupleDims(dataStructure);
 }
@@ -110,7 +112,7 @@ TEST_CASE("OrientationAnalysis::ReadGrainMapper3D:NonCompatible_Parameters", "[O
     args.insertOrAssign(ReadGrainMapper3DFilter::k_ReadLabDCT_Key, std::make_any<bool>(true));
     args.insertOrAssign(ReadGrainMapper3DFilter::k_CreatedDCTImageGeometryPath_Key, std::make_any<DataPath>(computedDCTGeometryPath));
     args.insertOrAssign(ReadGrainMapper3DFilter::k_CellAttributeMatrixName_Key, std::make_any<std::string>(k_Cell_Data));
-    args.insertOrAssign(ReadGrainMapper3DFilter::k_CellEnsembleAttributeMatrixName_Key, std::make_any<std::string>(k_EnsembleAttributeMatrix));
+    args.insertOrAssign(ReadGrainMapper3DFilter::k_CellEnsembleAttributeMatrixName_Key, std::make_any<std::string>(k_Cell_Ensemble_Data));
 
     args.insertOrAssign(ReadGrainMapper3DFilter::k_ConvertPhaseToInt32_Key, std::make_any<bool>(false));
     args.insertOrAssign(ReadGrainMapper3DFilter::k_ConvertOrientationData_Key, std::make_any<bool>(false));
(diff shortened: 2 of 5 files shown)


imikejackson (approved):

imikejackson:
I'm not sure if I intentially left it blank or I just forgot about it. Clearly the unit test did not reveal the bug so that will need to be updated also.

I do think we should keep as much of the meta data as possible for these data sets. Especially if the user is writing the results out to a .dream3d file, as you stated, there would be some loss of information which really should be preserved.

nikprabhu:
Thanks for the clarification. As mentioned, I will leave this PR as is since it is already approved, unless you would prefer the metadata preservation changes to be included here as well.

Otherwise, in the follow-up PR, I will also address extending the test coverage.

imikejackson:
@all-contributors please add @nikprabhu for bug

imikejackson:
@nikprabhu Let's go ahead and do the full meta data and unit test update.

nikprabhu:
Thanks. I have now pushed two additional commits extending the original fix.

The first commit (https://github.com/BlueQuartzSoftware/simplnx/pull/1620/commits/97de07d73de87ac33731bf98d1d5ebc724ed94e9) addresses writing `UniversalHermannMauguin` metadata from the GrainMapper3D HDF5 file into the ensemble data of the simplNX `DataStructure`. For now, I retained the original `UniversalHermannMauguin` naming for the string array to stay consistent with GrainMapper3D terminology. I did consider whether a more general identifier such as `SpaceGroupSymbol` or `HermannMauguinSymbol` would fit better alongside the more generic ensemble array naming, but I assumed most users working with GrainMapper3D would already be familiar with this terminology.

The second commit (https://github.com/BlueQuartzSoftware/simplnx/pull/1620/commits/eae9d123cc2465f309385c5d24d8992a41c27bd3) extends the existing test coverage to also compare ensemble data, in addition to the already existing cell data comparison, against the exemplar DREAM3D file.

While implementing this, I realized that even this test would not have caught the original bug because `UniversalHermannMauguin` was previously neither written into the ensemble data nor present in the exemplar DREAM3D file itself. So for this metadata to become part of the current testing workflow, the exemplar DREAM3D file would also need to be updated. I was not able to locate the `DREAM3D_Data` directory directly in the repository, although I do have the generated test data available locally in the build directory.

I was also wondering what your preferred direction would be regarding the testing strategy here.

One possibility would be to make the test more robust by directly validating specific metadata values against the original GrainMapper3D HDF5 source file. That likely would have caught the original bug immediately, but it also starts moving the current test away from its existing “generated DREAM3D structure vs exemplar DREAM3D structure” approach.

Another possibility would be to continue relying on the exemplar-based approach, but update the exemplar file and explicitly validate that certain expected ensemble metadata values are identical between the generated and exemplar DREAM3D structures.

Before extending the test further, I wanted to first get your thoughts on this.

imikejackson:
@nikprabhu We can update the test file. We keep them in a "Data Archive" release in this repository. You will not be able to update the repository so we will just need to coordinate a bit on moving forward. 


Sorry for the delay in response, we were moving offices last week and are finally getting back to actual work this week.

imikejackson:
@nikprabhu You need to give me access to your repo so I can make changes to this branch. Or I can just take over the branch and work it from there. You will still be listed as a contributor.

nikprabhu:
Sorry @imikejackson for the delayed response. I got caught up with a few other things.

I have now formatted the test file according to clang-format, so that pipeline should hopefully pass as well.

I already have "Allow edits by maintainers" enabled for this PR. Does that give you the access you need, or are you running into some specific permission issue?

If necessary, I can also give you collaborator access to my fork.

nikprabhu:
@imikejackson just wanted to gently check if you're waiting on any action from my side for this PR.

imikejackson:
@nikprabhu Take a look at the changes that I made. If you are happy with them then we can look at merging this PR sooner rather than later.

nikprabhu:
@imikejackson  Sorry, may be I am missing some detail here but I do not immedietely see the changes that you are talking about.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/GrainMapper3DUtilities.cpp
```diff
@@ -398,7 +398,7 @@ herr_t GrainMapperReader::readPhaseInfo(hid_t parentId)
       return error;
     }
 
-    error = H5Lite::readStringDataset(phaseGid, Constants::k_Name, phase.UniversalHermannMauguin);
+    error = H5Lite::readStringDataset(phaseGid, Constants::k_UniversalHermannMauguinName, phase.UniversalHermannMauguin);
```
imikejackson, line 401: Behavioral concern: previously this read `k_Name`, which is always present, so it never failed here. Now reading the `UniversalHermannMauguin` dataset, an `error < 0` will abort `readPhaseInfo()` and bubble up as the `-39801` failure in `copyPhaseInformation()`. If any GrainMapper3D file (older export, or tooling that omits this optional field) lacks a `UniversalHermannMauguin` dataset in a PhaseNN group, files that read fine before this PR will now fail entirely. Consider tolerating a missing dataset (leave the string empty and continue) rather than treating it as fatal.

src/Plugins/OrientationAnalysis/test/ReadGrainMapper3DTest.cpp
```diff
@@ -78,6 +78,8 @@ TEST_CASE("OrientationAnalysis::ReadGrainMapper3D:Default_Parameters", "[Orienta
   UnitTest::CompareExemplarToGenerateAttributeMatrix(dataStructure, exemplarDCTGeometryPath.createChildPath(k_Cell_Data), dataStructure, computedDCTGeometryPath.createChildPath(k_Cell_Data));
   UnitTest::CompareExemplarToGenerateAttributeMatrix(dataStructure, exemplarAbsorptionCTGeometryPath.createChildPath(k_Cell_Data), dataStructure,
                                                      computedAbsorptionCTGeometryPath.createChildPath(k_Cell_Data));
+  UnitTest::CompareExemplarToGenerateAttributeMatrix(dataStructure, exemplarDCTGeometryPath.createChildPath(k_Cell_Ensemble_Data), dataStructure,
```
imikejackson, line 82: Two issues with this added comparison:

1. `CompareExemplarToGenerateAttributeMatrix` calls `getDataRefAs<AttributeMatrix>` on the exemplar path and **throws** if it doesn't exist. This now requires the exemplar `7_0_SimulatedMultiPhase.dream3d` to contain an AM named exactly `Cell Ensemble Data` (the value of `k_Cell_Ensemble_Data`) under `LabDCT (default)`. Please confirm CI is green / the exemplar AM is actually named that.

2. The helper iterates over the **exemplar's** arrays and skips anything not present there. The pre-existing exemplar was generated by code that never wrote a `UniversalHermannMauguin` array, so the newly-populated values are never actually compared - only `CrystalStructures`, `MaterialName`, and `LatticeConstants` are. The real bug being fixed is therefore not covered by this test. Regenerating and re-uploading the exemplar with a correctly-read `UniversalHermannMauguin` array would both resolve the name-match dependency above and give the fix genuine coverage.

imikejackson (commented):
A few notes on the fix and the test coverage. The core read correction (`k_Name` -> `k_UniversalHermannMauguinName`) and the preflight/execute wiring of the new `UniversalHermannMauguin` ensemble array look correct and tuple-dimension consistent. Three things worth addressing, detailed inline.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/utilities/GrainMapper3DUtilities.cpp
nikprabhu: I must admit that I am not familiar with the evolution of the GrainMapper3D data format, i.e. whether `UniversalHermannMauguin` has always been part of the phase information or was introduced later as optional metadata.
That said, I fully agree that the check should be non-fatal. If the dataset is missing, it would be better to leave the string empty and continue reading rather than aborting the entire import, especially if that helps preserve backward compatibility with older GrainMapper3D exports.
I agree to update the implementation accordingly.

src/Plugins/OrientationAnalysis/test/ReadGrainMapper3DTest.cpp
nikprabhu: Yes, this is exactly the concern I raised earlier in this thread. Since `CompareExemplarToGenerateAttributeMatrix()` iterates over the arrays present in the exemplar, the exemplar file need to be updated with the new `UniversalHermannMauguin` information before this bug can actually be covered by the test.

Since the exemplar data archive is maintained separately, could you please share the preferred workflow for updating it?

imikejackson:
[0001-TEST-Fix-testing-with-hard-coded-REQUIRES-instead-of.patch](https://github.com/user-attachments/files/29506601/0001-TEST-Fix-testing-with-hard-coded-REQUIRES-instead-of.patch)
@nikprabhu Please apply this patch and push.

src/Plugins/OrientationAnalysis/test/ReadGrainMapper3DTest.cpp
imikejackson: See my other reply.

nikprabhu:
@imikejackson made a new commit with the patch applied and tested locally.

imikejackson (approved):

PR #1631, opened 2026-06-04
VV: EbsdLib 3.0.0 + V&V of 6 Filters

imikejackson:


nyoungbq (approved):
Approved, I didn't find any issues. The V&V stuff does feel overly wordy, but I think the information being conveyed still comes across so I am approving. On the C-Axis V&V, since you asked specifically, the python script processing of the data to make it suitable for both version of SIMPL came across fine. I understood the logic flow, so I would leave it.

src/Plugins/OrientationAnalysis/test/ComputeIPFColorsTest.cpp
```diff
@@ -143,6 +144,78 @@ TEST_CASE("OrientationAnalysis::ComputeIPFColors", "[OrientationAnalysis][Comput
...
+    const size_t size = a.getSize();
+    for(size_t i = 0; i < size; ++i)
+    {
+      if(a[i] != b[i])
+      {
+        return true;
+      }
+    }
+    return false;
+  };
```
JDuffeyBQ: This is just `!std::equal(a, b)`

src/Plugins/OrientationAnalysis/test/ComputeFaceIPFColoringTest.cpp
```diff
@@ -132,6 +133,70 @@ TEST_CASE("OrientationAnalysis::ComputeFaceIPFColoringFilter: Invalid filter exe
...
+    const size_t size = a.getSize();
+    for(size_t i = 0; i < size; ++i)
+    {
+      if(a[i] != b[i])
+      {
+        return true;
+      }
+    }
+    return false;
+  };
```
JDuffeyBQ: This is just `!std::equal(a, b)`

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/ComputeIPFColorsFilter.cpp
```diff
@@ -147,6 +154,19 @@ Result<> ComputeIPFColorsFilter::executeImpl(DataStructure& dataStructure, const
...
+  {
+  case 1:
+    inputValues.colorKey = ebsdlib::ColorKeyKind::PUCM;
+    break;
+  case 2:
+    inputValues.colorKey = ebsdlib::ColorKeyKind::NolzeHielscher;
+    break;
+  default:
+    inputValues.colorKey = ebsdlib::ColorKeyKind::TSL;
+    break;
```
JDuffeyBQ: Rather than default I think it should be the exact value. A value out of range of an enum is an error. And this is also done in the filter as well so it could be a function.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/ComputeFaceIPFColoringFilter.cpp
```diff
@@ -147,6 +161,19 @@ Result<> ComputeFaceIPFColoringFilter::executeImpl(DataStructure& dataStructure,
...
+  case 1:
+    inputValues.ColorKey = ebsdlib::ColorKeyKind::PUCM;
+    break;
+  case 2:
+    inputValues.ColorKey = ebsdlib::ColorKeyKind::NolzeHielscher;
+    break;
+  default:
+    inputValues.ColorKey = ebsdlib::ColorKeyKind::TSL;
+    break;
+  }
```
JDuffeyBQ, line 176: See comment in algorithm class

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/WritePoleFigureFilter.cpp
```diff
@@ -310,6 +316,16 @@ Result<> WritePoleFigureFilter::executeImpl(DataStructure& dataStructure, const
...
   inputValues.IntensityPlot3Name = filterArgs.value<DataObjectNameParameter::ValueType>(k_IntensityPlot3Name);
 
+  switch(filterArgs.value<ChoicesParameter::ValueType>(k_HexConvention_Key))
+  {
+  case 1:
+    inputValues.HexConvention = ebsdlib::HexConvention::XParallelAStar;
+    break;
+  default:
+    inputValues.HexConvention = ebsdlib::HexConvention::XParallelA;
+    break;
```
JDuffeyBQ: Similar to other enum comment.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborMisorientations.cpp
```diff
@@ -65,14 +65,18 @@ Result<> ComputeFeatureNeighborMisorientations::operator()()
     const NeighborList<int32_t>::VectorType featureNeighborList = inNeighborList.at(static_cast<int32_t>(i));
 
     tempMisorientationLists[i].assign(featureNeighborList.size(), -1.0);
+    // Initialize the divisor once per outer-loop iteration (per feature). Previously this was
+    // assigned inside the inner j-loop, which clobbered the per-mismatch decrement below — the
+    // resulting divisor only reflected the LAST neighbor's match/mismatch state, producing wrong
+    // per-feature averages whenever neighbors had mixed phases. Fixed 2026-06-02 during V&V cycle.
+    tempMisoList = featureNeighborList.size();
```
JDuffeyBQ: This is only used inside the loop so the definition could be moved here.

src/Plugins/OrientationAnalysis/test/ComputeFeatureNeighborMisorientationsTest.cpp
```diff
@@ -18,75 +23,136 @@ using namespace nx::core;
...
+const std::string k_FeaturePhasesName = "FeaturePhases";
+const std::string k_AvgQuatsName = "AvgQuats";
+const std::string k_NeighborListName = "NeighborList";
+const std::string k_CrystalStructuresName = "CrystalStructures";
 
-TEST_CASE("OrientationAnalysis::ComputeFeatureNeighborMisorientationsFilter", "[OrientationAnalysis][ComputeFeatureNeighborMisorientationsFilter]")
+const std::string k_MisorientationListOutName = "MisorientationListOut";
+const std::string k_AvgMisorientationsOutName = "AvgMisorientationsOut";
+
+inline std::array<float32, 4> QuatFromPhi1Deg(float32 phi1Deg)
```
JDuffeyBQ: None of the functions here should have `inline`.

src/Plugins/OrientationAnalysis/test/ComputeKernelAvgMisorientationsTest.cpp
```diff
@@ -19,65 +24,119 @@ using namespace nx::core;
...
+
+const std::string k_FeatureIdsName = "FeatureIds";
+const std::string k_CellPhasesName = "Phases";
+const std::string k_QuatsName = "Quats";
+const std::string k_CrystalStructuresName = "CrystalStructures";
+
+const std::string k_KAMOutName = "KernelAverageMisorientationsOut";
 
-TEST_CASE("OrientationAnalysis::ComputeKernelAvgMisorientationsFilter", "[OrientationAnalysis][ComputeKernelAvgMisorientationsFilter]")
+inline std::array<float32, 4> QuatFromPhi1Deg(float32 phi1Deg)
```
JDuffeyBQ: Same issue with `inline`

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/BadDataNeighborOrientationCheck.cpp
```diff
@@ -59,32 +64,66 @@ Result<> BadDataNeighborOrientationCheck::operator()()
...
+  // Validate that every entry in the CrystalStructures ensemble array is a valid Laue-group index
+  // (< orientationOps.size()). Catches malformed inputs such as a legacy CreateEnsembleInfo sentinel
+  // value (999) at ensemble index 0 before they cause an out-of-bounds dereference in the per-voxel
+  // loop below. The UnknownCrystalStructure value is explicitly allowed as a sentinel; voxels whose
+  // phase resolves to it will be skipped by the cellPhases > 0 guard. CrystalStructures is typically
+  // tiny (2-4 entries), so the cost is negligible.
+  const usize numOrientationOps = orientationOps.size();
+  for(usize i = 0; i < crystalStructures.getSize(); ++i)
+  {
+    if(crystalStructures[i] >= numOrientationOps && crystalStructures[i] != static_cast<uint32>(ebsdlib::CrystalStructure::UnknownCrystalStructure))
```
JDuffeyBQ: Isn't `ebsdlib::CrystalStructure::UnknownCrystalStructure` already `uint32`?

src/Plugins/OrientationAnalysis/test/BadDataNeighborOrientationCheckTest.cpp
```diff
@@ -15,6 +15,11 @@
 namespace fs = std::filesystem;
 using namespace nx::core;
 
+// Wrapped in an anonymous namespace so every symbol below has internal linkage. Test TUs that share
+// the OrientationAnalysisUnitTest binary must not expose same-named external symbols, or the linker
+// silently merges duplicate definitions across files (see PR #1630 for that failure mode).
```
JDuffeyBQ: I don't think we need to keep re-explaining anonymous namespaces.

src/Plugins/OrientationAnalysis/test/BadDataNeighborOrientationCheckTest.cpp
```diff
@@ -35,12 +40,58 @@ const std::string k_CStuctsName = "Crystal Structures";
 const DataPath k_CStuctsArrayPath = k_CellEnsembleDataPath.createChildPath(k_CStuctsName);
 } // namespace VerificationConstants
 
+namespace ClassFourInvariants
+{
+// Capture the current mask array contents into a std::vector for before/after invariant checks.
+inline std::vector<uint8> CaptureMask(const DataStructure& dataStructure)
```
JDuffeyBQ: `inline` issue again

src/Plugins/OrientationAnalysis/test/BadDataNeighborOrientationCheckTest.cpp
```diff
@@ -1647,3 +1709,198 @@ TEST_CASE("OrientationAnalysis::BadDataNeighborOrientationCheckFilter: SIMPL Bac
...
+  // Compare: Run 2's output must equal Run 1's output (filter has reached a fixed point).
+  const auto& maskArray = dataStructure.getDataRefAs<UInt8Array>(VerificationConstants::k_MaskArrayPath);
+  const auto& store = maskArray.getDataStoreRef();
+  REQUIRE(maskAfterRun1.size() == store.getSize());
+  for(usize i = 0; i < store.getSize(); ++i)
+  {
+    if(store.getValue(i) != maskAfterRun1[i])
+    {
+      const std::string errorMsg = fmt::format("Idempotence violated at index {}. Run1: {} | Run2: {}", i, maskAfterRun1[i], store.getValue(i));
+      CAPTURE(errorMsg);
```
JDuffeyBQ: This could use `INFO()` instead.

src/Plugins/OrientationAnalysis/test/BadDataNeighborOrientationCheckTest.cpp
```diff
@@ -1647,3 +1709,198 @@ TEST_CASE("OrientationAnalysis::BadDataNeighborOrientationCheckFilter: SIMPL Bac
...
+
+  // Expected: center (index 4) flips; corners stay bad (only 2 good neighbors each).
+  const std::array<uint8, 9> expectedMask = {0, 1, 0, 1, 1, 1, 0, 1, 0};
+  const auto& maskStore = dataStructure.getDataAs<UInt8Array>(VerificationConstants::k_MaskArrayPath)->getDataStoreRef();
+  for(usize i = 0; i < 9; ++i)
+  {
+    if(maskStore.getValue(i) != expectedMask[i])
+    {
+      const std::string errorMsg = fmt::format("2D fixture: values diverged at index {}. Expected: {} | Actual: {}", i, expectedMask[i], maskStore.getValue(i));
+      CAPTURE(errorMsg);
```
JDuffeyBQ: This could use `INFO()` instead.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborCAxisMisalignments.cpp
```diff
@@ -104,11 +104,16 @@ Result<> ComputeFeatureNeighborCAxisMisalignments::operator()()
     const NeighborList<int>::VectorType& currentNeighborList = neighborList[featureIdx];
     auto& currentMisalignmentList = misalignmentLists[featureIdx];
     currentMisalignmentList.resize(currentNeighborList.size(), -1.0);
+    // Initialize the divisor once per outer-loop iteration (per feature). Previously this was
+    // assigned inside the inner j-loop, which clobbered the per-mismatch decrement below — the
+    // resulting divisor only reflected the LAST neighbor's match/mismatch state, producing wrong
+    // per-feature averages whenever neighbors had mixed phases. Fixed 2026-06-04 during V&V cycle
+    // (sibling of the same divisor bug fixed in ComputeFeatureNeighborMisorientations on 2026-06-02).
+    hexNeighborListSize = currentNeighborList.size();
```
JDuffeyBQ: Definition here could also be moved into loop.

src/Plugins/OrientationAnalysis/test/ComputeFeatureNeighborCAxisMisalignmentsTest.cpp
```diff
@@ -1,92 +1,245 @@
...
+  usize totalFeatures = 0;
+};
 
-TEST_CASE("OrientationAnalysis::ComputeFeatureNeighborCAxisMisalignmentsFilter: Valid Filter Execution", "[OrientationAnalysis][ComputeFeatureNeighborCAxisMisalignmentsFilter]")
+// Build an ImageGeom-backed scaffold. Cell-level arrays (FeatureIds, Phases) are sized
+// {nZ, nY, nX}; feature-level arrays (FeaturePhases, AvgQuats, NeighborList) are sized
+// {numFeatures}; ensemble-level arrays (CrystalStructures) are sized {numCrystalStructures}.
+// Defaults: every cell assigned to feature 1 / phase 1; every feature phase 0 (caller to set);
+// identity quats; empty neighbor lists; CrystalStructures[0] = sentinel 999.
+inline ToyData CreateScaffold(usize nX, usize nY, usize nZ, usize numFeatures, usize numCrystalStructures)
```
JDuffeyBQ: `inline` issue

src/simplnx/DataStructure/IListStore.hpp
```diff
@@ -49,19 +49,16 @@ class IListStore
 
   /**
    * @brief Returns the total number of lists in the list store.
-   * @return uint64 The number of lists
+   * @return usize The number of lists
    */
-  virtual uint64 getNumberOfLists() const = 0;
+  virtual usize getNumberOfLists() const = 0;
```
JDuffeyBQ, line 54: Is there a reason for changing this type?

JDuffeyBQ (changes requested):
There's a commit titled "BUG: WriteImageFilter preflight did not validate tuple dimensions" that has no changes to `WriteImageFilter`?

src/simplnx/DataStructure/IListStore.hpp
imikejackson: Because it is a USIZE, and not a uint64.

JDuffeyBQ: I was mostly just wondering if there was a reason it had been uint64 previously.

src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeFeatureNeighborMisorientations.cpp
imikejackson: Refactored — the declaration now lives in the outer per-feature loop, with `featureNeighborList.size()` (resp. `currentNeighborList.size()`) as its initializer. That removes both the function-scope declaration and the redundant `= 0;` reset at the end of the outer-loop body. Same change applied to the sibling `ComputeFeatureNeighborCAxisMisalignments.cpp:112`.

One clarifying question on intent: the variable can't go any tighter than the outer loop because the divisor is read *after* the inner j-loop completes (it's the per-feature average's denominator at line 103 / line 167). Was the outer loop what you meant, or were you thinking of something else?

imikejackson: Refactored — the declaration now lives in the outer per-feature loop, with `featureNeighborList.size()` (resp. `currentNeighborList.size()`) as its initializer. That removes both the function-scope declaration and the redundant `= 0;` reset at the end of the outer-loop body. Same change applied to the sibling `ComputeFeatureNeighborCAxisMisalignments.cpp:112`.

One clarifying question on intent: the variable can't go any tighter than the outer loop because the divisor is read *after* the inner j-loop completes (it's the per-feature average's denominator at line 103 / line 167). Was the outer loop what you meant, or were you thinking of something else?

imikejackson:
## Review feedback addressed — 3 follow-up commits

| Commit | Scope |
|---|---|
| `afd06e2dd` | All 13 of @JDuffeyBQ's inline review categories + 1 follow-up enum-default site found via an adversarial sweep (`WritePoleFigure.cpp:622` silent `default: break;` on a CrystalStructure switch). |
| `c111a40e0` | V&V report wordiness pass across all 7 V&V'ed filters (per @nyoungbq's feedback). At-a-glance cells cut from 1000-2200 chars down to 300-800 chars; no evidence removed. |
| `f148c5a92` | Round 2: revert + cleanup pass — see below. |

### Round-2 cleanup commit (`f148c5a92`)

Dead-code removal: all 5 algorithm classes in PR scope (`BadDataNeighborOrientationCheck`, `ComputeFaceIPFColoring`, `ComputeFeatureFaceMisorientation`, `ComputeFeatureNeighborMisorientations`, `WritePoleFigure`) had a `getCancel()` method that was never called — algorithm bodies use `m_ShouldCancel` directly. Method removed entirely from headers and `.cpp` files.

LLM-style boilerplate scrub: dropped a 6-line hedge-comment block in `ComputeAvgCAxes.cpp` finalize loop + 2 boilerplate Doxygen blocks above parallel-impl classes.

Readability polish: `temp` scratch variable replaced with inline `static_cast<float32>` (`ComputeAvgCAxes.cpp`); mask-load error message now includes the underlying `std::out_of_range::what()` (`BadDataNeighborOrientationCheck.cpp`); 3 of 4 `if(!maskCompare->isTrue(...))` sites resolved via named bool local; inline comment on the `w` variable's first use in `ComputeFeatureNeighborCAxisMisalignments.cpp`.

@JDuffeyBQ — all 16 of your inline threads resolved with the commits above. The commit-title clarification on `c43ec2f85` is pending @imikejackson's follow-up.

cc @nyoungbq @jmarquisbq

nyoungbq (approved):
Looks good, changed vv files are much easier to read

PR #1632, opened 2026-06-04
VV: Compute Feature Neighbor C-Axis Misalignments fully V&V'ed

imikejackson:
Summary:
- Found and fixed 1 bug (divisor clobbered inside inner j-loop of algorithm.cpp, sibling of F#2's D1 — production-relevant via shipping EBSD_Hexagonal_Data_Analysis.d3dpipeline which runs with find_avg_misals=true);
- documented 5 deviations from DREAM3D 6.5.171 (D1 divisor bug, D2 avg-array fillValue uncertainty (latent, confirmed dormant on current backend), D4 EbsdLib quat-matrix swap precision drift, D5 PR #1438 preflight-banner UX downgrade, D6 Hexagonal_Low support gap);
- retired 1 test (hex-only exemplar consumer — could not trigger the divisor bug; removed its download_test_data line from test/CMakeLists.txt);
- unit tests replaced with 4 inlined Class 1 (Analytical) + Class 4 (Invariant) test fixtures (incl. a 10x10x1 6-feature realistic microstructure with 3 bug-exposing per-feature configurations);
- added 3 V&V source-tree deliverables (report, deviations, provenance);
- performed empirical A/B against 3 binaries (DREAM3D 6.5.171 official, 6.5.172 with D1+D4+D6 backports, SIMPLNX) and confirmed byte-for-byte match between 6.5.172 (post-backport) and SIMPLNX — 24/24 values bit-identical. Artifacts at /Users/mjackson/Desktop/F6_AB_Test/;
- fixed Hexagonal_Low note in user-facing doc.


imikejackson:
Taken over by https://github.com/BlueQuartzSoftware/simplnx/pull/1631

PR #1634, opened 2026-06-05
PERF: Enlarge zlib read buffer in NIfTI reader for NAS performance

imikejackson:
The NIfTI reader opened files with a bare gzopen(), leaving zlib at its 8 KB default internal buffer. zlib refills from the underlying file in buffer-sized chunks, so a multi-GB .nii.gz triggered hundreds of thousands of small reads. On network attached storage each refill is a latency-bound round-trip, making reads disproportionately slow compared to a local disk.

Call gzbuffer() with a 4 MiB buffer (new shared constant nifti::k_GzReadBufferSize) immediately after every gzopen(), before the first read/seek. This cuts the number of NAS round-trips by ~512x on a 2 GB file while costing only a few MiB of working memory. Larger buffers yield diminishing returns once they exceed the link's bandwidth-delay product and client read-ahead.

### DIFF
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/utils/NiftiUtilities.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/utils/NiftiUtilities.cpp
@@ -119,8 +119,10 @@ Result<NiftiMetadata> ReadNiftiHeader(const std::filesystem::path& filePath, boo
     return MakeErrorResult<NiftiMetadata>(-34700, fmt::format("Could not open NIfTI file for reading: '{}'", pathStr));
   }
 
+  gzbuffer(gz, k_GzReadBufferSize);
+
   nifti_1_header hdr{};
-  int bytesRead = gzread(gz, &hdr, static_cast<unsigned int>(k_HeaderSize));
+  int bytesRead = gzread(gz, &hdr, k_HeaderSize);
   gzclose(gz);
 
   if(bytesRead != static_cast<int>(k_HeaderSize))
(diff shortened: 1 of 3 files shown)


src/Plugins/SimplnxCore/src/SimplnxCore/utils/NiftiUtilities.hpp
```diff
@@ -46,6 +46,23 @@ inline constexpr usize k_HeaderSize = 348;
...
+ * syscall each. On a local disk that is negligible, but on network
+ * attached storage (NAS) every refill is a separate network round-trip,
+ * so a multi-GB `.nii.gz` triggers hundreds of thousands of small,
+ * latency-bound reads. Raising the buffer to 4 MiB cuts the number of
+ * round-trips by ~512x while costing only a few × this size in working
+ * memory (negligible next to the voxel volume). Larger values yield
+ * diminishing returns once the buffer exceeds the link's
+ * bandwidth-delay product and the client's own read-ahead.
+ */
+inline constexpr usize k_GzReadBufferSize = 1u << 22; // 4 MiB
```
JDuffeyBQ: This should be defined to be the type zlib expects

src/Plugins/SimplnxCore/src/SimplnxCore/utils/NiftiUtilities.cpp
```diff
@@ -119,6 +119,11 @@ Result<NiftiMetadata> ReadNiftiHeader(const std::filesystem::path& filePath, boo
     return MakeErrorResult<NiftiMetadata>(-34700, fmt::format("Could not open NIfTI file for reading: '{}'", pathStr));
   }
 
+  // Match the voxel reader's enlarged zlib buffer so the (single) header read
+  // is served by one large refill rather than zlib's 8 KB default. Harmless
+  // here since only 348 bytes are read, but keeps the two open sites consistent.
```
JDuffeyBQ: I think we should just direct to the variable's comment rather than repeating the rationale everywhere.

PR #1640, opened 2026-06-17
VV/BUG: Identify Sample Full V&V and Minor patches

nyoungbq:
- Full V&V Deviation, Provenance, and final report
- Bring tie breaking behavior of slice by slice in line with main alg and legacy
- Defined proper handling of SingleVoxelImage via compile time change that doesn't effect the existing behavior of other types


imikejackson:
## A/B verification of `IdentifySample-D1` — it was mis-attributed (now corrected)

Ran the degenerate-geometry A/B that `IdentifySample-D1` flagged as unverified: the exact 2D/1D/single-voxel fixtures authored as legacy `.dream3d`, run through **stock DREAM3D 6.5.171**, **SIMPLNX** (rebuilt from this branch), and a **6.5.172 proof-patch** build. The result corrects D1's root-cause hypothesis.

### Results
| Case | Geometry | FillHoles | stock-171 | patched-172 | NX (this PR) |
|---|---|---|---|---|---|
| 2D EmptyZ {3,4,1} | no | ✓ | ✓ | ✓ |
| 2D EmptyX {1,3,4} | no | ✓ | ✓ | ✓ |
| 1D {5,1,1} (tie-break) | no | ✓ | ✓ | ✓ |
| 3D {4,4,4} | yes | ✓ | ✓ | ✓ |
| 2D EmptyZ {5,5,1} | **yes** | **✗ hole NOT filled** | ✓ filled | ✓ filled |
| single voxel {1,1,1} | **yes** | ✓ not filled | ✓ | ✓ not filled |

### Three takeaways

1. **Connectivity is NOT a deviation.** 2D, 1D, and the `>=` tie-break produce byte-identical largest-component output across all three (first four rows). D1's stated root cause — "legacy's fixed 3D-stride neighbor formula… suspected incorrect" — is **refuted**: legacy's coordinate-based neighbor guards degenerate correctly, so the dimensionality dispatch is a faithful re-implementation. The `>` → `>=` slice tie-break change is consistent with both the main functor and legacy (`IdentifySample.cpp:234`).

2. **The real D1 deviation is hole-fill on 2D/1D — legacy is WRONG, NX correct (now proven).** Legacy's Phase-2 boundary test (`Processing/ProcessingFilters/IdentifySample.cpp:287`) includes `|| plane==0 || plane==(zp-1)`, which is *always true* when `zp==1` — so legacy flags every voxel in a 2D geometry as boundary-touching and **never fills any 2D/1D hole**. The 6.5.172 proof patch (a dimensionality-aware boundary check) makes legacy fill the 2D hole identically to NX, while leaving 3D and connectivity byte-for-byte unchanged — pinning the cause to the boundary test, not neighbor dispatch.

3. **The single-voxel case is a SIMPLNX bug fixed to match legacy — the reverse of what D1 suspected.** Legacy correctly does *not* fill a lone bad voxel. Pre-PR NX wrongly filled it (the neighbor loop never runs, so `touchesBoundary` stayed false). I rebuilt nxrunner from this branch and confirmed **change-1** (`touchesBoundary = k_NeighborCount == 0`) now leaves it unfilled = stock-171. So change-1 brings NX *into line with* legacy.

### Docs updated (committed to this branch, `a2a8cbe82`)
- `vv/deviations/IdentifySampleFilter.md`: D1 re-rooted to the hole-fill boundary test, status changed from "suspected" to **A/B-proven**, scoped to 2D/1D `FillHoles`; single-voxel reframed as a SIMPLNX fix (not a legacy deviation).
- `vv/IdentifySampleFilter.md`: at-a-glance, summary, algorithm-relationship, and deviations sections updated to match.

The 6.5.172 proof patch lives on the internal DREAM3D `v6_5_172` branch (not for release).

imikejackson (changes requested):
See comments. If you agree with changes then we can merge.

PR #1643, opened 2026-06-26
BUG: Make H5SUPPORT_MUTEX_LOCK a real process-wide HDF5 lock

joeykleingers:
## Summary

`H5SUPPORT_MUTEX_LOCK()` declared a fresh function-local `std::mutex` on every
expansion and immediately locked it, so it provided no mutual exclusion across
calls or threads (and expanded to nothing when `H5Support_USE_MUTEX` was
undefined). The four HDF5 helpers that use it (`IsGroup`, `GetObjectPath`,
`GetDatasetType`, `StringForHDFType`) were therefore effectively unsynchronized
against the non-thread-safe HDF5 C library.

* Add a single process-wide `ApiLock()` (a leaky Meyers singleton) and point the
  macro at it.
* Define `H5Support_USE_MUTEX` for the `simplnx` target
  (`target_compile_definitions(simplnx PUBLIC ...)`) so the macro is actually
  engaged rather than compiled out. `PUBLIC` also propagates the define to
  `simplnx_test`, so the regression guard runs live instead of taking the
  `SUCCEED()` no-op path.
* Add a regression test that hammers the shared lock from many threads and
  asserts no lost updates (guards against reintroducing the per-call-local-mutex
  bug).

## Scope / caveats

* **Coarse, process-wide lock.** This serializes the four macro-guarded HDF5 leaf
  calls on one mutex. It trades some HDF5-call concurrency for correctness against
  the non-thread-safe vcpkg HDF5 build.
* **Limited coverage.** The lock currently guards only those four helpers. The
  many other direct `H5*` calls throughout `FileIO`, `DatasetIO`, and
  `Dream3dIO` are **not** yet serialized by it, so this does not by itself close
  the broader HDF5 thread-safety story — it gives the rest of the code a single
  lock to adopt as it is hardened.

## Test Plan
- [x] `simplnx_test` HDF5 suite passes with the lock enabled (H5 Utilities,
  DatasetIO, HDF5ImplicitCopy, the ApiLock regression guard, Read Legacy
  DREAM3D-NX Data, Contains HDF5 IO Support)

### DIFF
--- a/src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
+++ b/src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
@@ -308,12 +308,18 @@ std::string SIMPLNX_EXPORT GetNameFromFilterType(H5Z_filter_t id);
  * macro; any other code needing the HDF5 lock aliases it rather than defining a
  * second one (two locks guarding one non-thread-safe library would still race).
  *
+ * SCOPE — the lock currently guards only the four macro-guarded helpers listed
+ * below. The many other direct H5* calls scattered through FileIO, DatasetIO,
+ * and Dream3dIO are NOT serialized by it, so this does not by itself make all
+ * HDF5 access thread-safe; it makes these four helpers safe and gives the rest a
+ * single lock to adopt as they are hardened.
+ *
  * CONTRACT — this is a NON-recursive std::mutex. Lock it ONLY around leaf HDF5
  * C-API calls, and NEVER while calling a function that may re-acquire it — in
  * particular the macro-guarded helpers (IsGroup, GetObjectPath, GetDatasetType,
- * StringForHDFType) and probeSingleDeflateEligibility, which lock it themselves.
- * Re-entering on the same thread would self-deadlock. Never hold it around heavy
- * CPU work (e.g. (de)compression) either — that must run off the lock.
+ * StringForHDFType), which lock it themselves. Re-entering on the same thread
+ * would self-deadlock. Never hold it around heavy CPU work (e.g. (de)compression)
+ * either — that must run off the lock.
  */
 SIMPLNX_EXPORT std::mutex& ApiLock();
 
--- a/src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
+++ b/src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
@@ -12,16 +12,14 @@
 #include <hdf5.h>
 
 #include <iostream>
+#include <mutex>
 #include <numeric>
 #include <stdexcept>
 #include <string>
 #include <vector>
 
 #ifdef H5Support_USE_MUTEX
-#include <mutex>
-#define H5SUPPORT_MUTEX_LOCK()                                                                                                                                                                         \
-  std::mutex mutex;                                                                                                                                                                                    \
-  std::lock_guard<std::mutex> lock(mutex);
+#define H5SUPPORT_MUTEX_LOCK() std::lock_guard<std::mutex> h5ApiLock(nx::core::HDF5::Support::ApiLock());
 #else
 #define H5SUPPORT_MUTEX_LOCK()
 #endif
@@ -299,6 +297,25 @@ std::string SIMPLNX_EXPORT StringForHDFType(hid_t dataTypeIdentifier);
  * @return std::string
  */
 std::string SIMPLNX_EXPORT GetNameFromFilterType(H5Z_filter_t id);
+
+/**
+ * @brief Returns the process-wide HDF5 API mutex.
+ *
+ * The vcpkg HDF5 build is not thread-safe, so every thread that calls into the
+ * HDF5 C library must serialize on this single lock. It is a Meyers singleton —
+ * one instance for the whole process — so every translation unit linked into
+ * libsimplnx shares the exact same mutex. It backs the H5SUPPORT_MUTEX_LOCK()
+ * macro; any other code needing the HDF5 lock aliases it rather than defining a
+ * second one (two locks guarding one non-thread-safe library would still race).
+ *
+ * CONTRACT — this is a NON-recursive std::mutex. Lock it ONLY around leaf HDF5
+ * C-API calls, and NEVER while calling a function that may re-acquire it — in
+ * particular the macro-guarded helpers (IsGroup, GetObjectPath, GetDatasetType,
+ * StringForHDFType) and probeSingleDeflateEligibility, which lock it themselves.
+ * Re-entering on the same thread would self-deadlock. Never hold it around heavy
+ * CPU work (e.g. (de)compression) either — that must run off the lock.
+ */
+SIMPLNX_EXPORT std::mutex& ApiLock();
 
 } // namespace Support
 } // namespace nx::core::HDF5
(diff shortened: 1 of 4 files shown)


src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
```diff
@@ -12,16 +12,14 @@
 #include <hdf5.h>
 
 #include <iostream>
+#include <mutex>
 #include <numeric>
 #include <stdexcept>
 #include <string>
 #include <vector>
 
 #ifdef H5Support_USE_MUTEX
```
imikejackson, line 21: **The lock is inert in this build.** `H5Support_USE_MUTEX` is never defined anywhere in the simplnx tree — a full-repo grep (all sources, `CMakeLists.txt`, `cmake/`, presets, JSON) finds only this `#ifdef`. So `H5SUPPORT_MUTEX_LOCK()` still expands to nothing, and the four helpers remain unsynchronized against the non-thread-safe HDF5 C library exactly as before. `ApiLock()` compiles and links but is called by nothing.

If the intent is to actually engage the lock, this PR should also define `H5Support_USE_MUTEX` for the simplnx target (with a perf justification). If the macro is intentionally left off (HDF5 serialized elsewhere, or vcpkg HDF5 built thread-safe), please say so in the PR description — otherwise the change reads as a behavioral bug fix when runtime behavior is actually unchanged.

src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
```diff
@@ -300,5 +298,24 @@ std::string SIMPLNX_EXPORT StringForHDFType(hid_t dataTypeIdentifier);
...
+ * HDF5 C library must serialize on this single lock. It is a Meyers singleton —
+ * one instance for the whole process — so every translation unit linked into
+ * libsimplnx shares the exact same mutex. It backs the H5SUPPORT_MUTEX_LOCK()
+ * macro; any other code needing the HDF5 lock aliases it rather than defining a
+ * second one (two locks guarding one non-thread-safe library would still race).
+ *
+ * CONTRACT — this is a NON-recursive std::mutex. Lock it ONLY around leaf HDF5
+ * C-API calls, and NEVER while calling a function that may re-acquire it — in
+ * particular the macro-guarded helpers (IsGroup, GetObjectPath, GetDatasetType,
+ * StringForHDFType) and probeSingleDeflateEligibility, which lock it themselves.
```
imikejackson: **`probeSingleDeflateEligibility` does not exist** anywhere in the codebase — a whole-tree grep for `probe` finds only the GBPD/GBCD orientation algorithms, and nothing other than the four macro-guarded helpers acquires `ApiLock()`. Please remove this reference, or correct it to the actual function name if one was intended (and if such a function exists, it should be acquiring the lock, which it currently isn't).

test/H5Test.cpp
```diff
@@ -1176,3 +1180,36 @@ TEST_CASE("DatasetIO: writeSpan bypasses chunking for small arrays even with com
   REQUIRE(info->layout == nx::core::UnitTest::DatasetLayout::Contiguous);
   REQUIRE(info->hasDeflate == false);
 }
+
+TEST_CASE("HDF5 ApiLock serializes access via H5SUPPORT_MUTEX_LOCK", "[simplnx][HDF5]")
+{
+#ifdef H5Support_USE_MUTEX
```
imikejackson, line 1186: **This regression guard is gated off in the default build.** Because `H5Support_USE_MUTEX` is never defined (see the macro comment in `H5Support.hpp`), this test always compiles the `#else` branch and just `SUCCEED()`s — it can't catch a reintroduction of the original per-call-local-mutex bug in CI.

Even when enabled, it increments an `int` under the macro rather than exercising any HDF5 call, so it validates only that the macro resolves to one shared `std::mutex`. That's a fair *macro* regression test, but it needs at least one build configuration where `H5Support_USE_MUTEX` is on for the guard to mean anything.

imikejackson (commented):
Inline notes below. The singleton design (out-of-line `SIMPLNX_EXPORT ApiLock()`, leaky Meyers init) is the right mechanism and well-justified. My main concern is that, as it lands on `develop`, runtime behavior is unchanged because the macro is gated behind a define that is never set in this build — details inline.

imikejackson:
## Review checklist

Tracking items from the inline review above. The singleton design itself is sound — these are the gaps before merge.

**Blocking**

- [x] **Resolve the inert-macro question.** `H5Support_USE_MUTEX` is never defined anywhere in the simplnx tree, so `H5SUPPORT_MUTEX_LOCK()` still expands to nothing and runtime behavior is unchanged. Either:
  - [x] define `H5Support_USE_MUTEX` for the simplnx target (e.g. `target_compile_definitions`), with a note on the perf impact of a process-wide HDF5 lock; **or**
  - [ ] document explicitly in the PR description that the macro is intentionally left disabled (and why), so it isn't read as a behavioral fix.
- [x] **Fix the `probeSingleDeflateEligibility` doc reference** in `H5Support.hpp` — no such function exists; remove it or replace with the correct name.

**Non-blocking**

- [x] Ensure the new `H5Test.cpp` regression guard runs in at least one CI build configuration (i.e. with `H5Support_USE_MUTEX` defined); otherwise it permanently takes the `SUCCEED()` no-op path and guards nothing.
- [x] Add a one-line note (PR description or comment) that this lock covers only the four macro-guarded helpers, not the many other direct `H5*` calls across `FileIO`/`DatasetIO`/`Dream3dIO`, so the broader HDF5 thread-safety story isn't implied to be closed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
joeykleingers: Enabled it: added `target_compile_definitions(simplnx PUBLIC "H5Support_USE_MUTEX")` to the top-level CMakeLists.txt with a comment noting the perf trade-off (a coarse, process-wide lock around the four HDF5 leaf calls). Chose `PUBLIC` so the define also reaches `simplnx_test`, otherwise the regression guard below stays a compile-time no-op. The four helpers now actually serialize on the shared `ApiLock()`.

src/simplnx/Utilities/Parsing/HDF5/H5Support.hpp
joeykleingers: Removed `probeSingleDeflateEligibility` — no such function exists; only the four macro-guarded helpers acquire the lock. Also added a SCOPE note to the `ApiLock()` doc comment that those four helpers are the only current coverage (the other direct `H5*` calls in FileIO/DatasetIO/Dream3dIO are not yet serialized).

test/H5Test.cpp
joeykleingers: With `H5Support_USE_MUTEX` now defined `PUBLIC` on the `simplnx` target it propagates to `simplnx_test`, so this test compiles the live branch in every build configuration rather than the `SUCCEED()` no-op. Confirmed it runs and passes locally: ctest "HDF5 ApiLock serializes access via H5SUPPORT_MUTEX_LOCK".

imikejackson (approved):

PR #1648, opened 2026-06-30
VV: ConvertOrientations fully V&V'ed

imikejackson:
## Summary
Full V&V of `ConvertOrientationsFilter` (Convert Orientation Representation) plus the algorithm cleanup the V&V surfaced. The filter is thin plumbing over EbsdLib's `Orientation::toX()` conversions (Rowenhorst 2015); the transform math is already verified by EbsdLib's own test suite, so this work verifies the filter's **value-add** — dispatch routing, component striding, and preflight — against an independent oracle, then explains the differences from DREAM3D 6.5.171.

* **Tests:** Retire the unknown-provenance `k_InitValues` test (circular-oracle risk) and replace it with a full 8x8 dispatch + striding test (every cross-type pair incl. Stereographic, 3 distinct multi-tuple orientations) using EbsdLib-3.0.0-derived landmarks, plus a Class 1 stereographic closed-form check. 1032 assertions pass.
* **Algorithm cleanup:** remove dead validation functors and unused error constants (`-67001`/`-67002`); add cancel checking, thread-safe throttled progress (`sendThreadSafeProgressMessage`), and `requireArraysInMemory` for out-of-core safety; bind input/output arrays by reference instead of copying.
* **V&V deliverables:** report (`vv/ConvertOrientationsFilter.md`), deviations D1–D4 (`vv/deviations/`), and a live legacy 6.5.171 A/B comparison (`vv/comparisons/`). Toy fixture: 4 of 6 shared conversions bit-identical, max |Δ| = 1.78e-6 (cubochoric) — float32 round-off from the differing intermediate-conversion paths, no bug.
* **Docs:** correct the inaccurate in-place Euler normalization note; add the Rowenhorst 2015 reference and a float32 precision note.

Oracle: Class 3 (Rowenhorst 2015) + Class 1 (Stereographic closed form) + Class 4 (round-trip).

## Test Plan
- [x] `ctest -R "ConvertOrientations"` passes on the in-core build (`NX-Com-Qt69-Vtk96-Rel`), 1032 assertions
- [x] Live legacy 6.5.171 A/B comparison run (PipelineRunner vs nxrunner, byte-identical input)
- [x] Out-of-core build run (intentionally skipped — single-algorithm filter, `requireArraysInMemory` applied)
- [ ] Second-engineer oracle review (V&V report is `READY FOR REVIEW`)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ConvertOrientations.cpp
```diff
@@ -207,27 +92,47 @@ ConvertOrientations::ConvertOrientations(DataStructure& dataStructure, const IFi
...
+                                                             k_TypeNames[static_cast<size_t>(m_InputValues->OutputType)]));
 
-  // Allow data-based parallelization
+  // Allow data-based parallelization; require both arrays resident in memory for out-of-core stores.
   ParallelDataAlgorithm parallelAlgorithm;
   parallelAlgorithm.setRange(0, totalPoints);
+  IParallelAlgorithm::AlgorithmArrays algArrays;
+  algArrays.push_back(&inputArray);
+  algArrays.push_back(&outputArray);
+  parallelAlgorithm.requireArraysInMemory(algArrays);
```
JDuffeyBQ, line 135: Did something change so that these need to be in memory? Looking at their usage they don't seem to have anything preventing them from being out of core.

imikejackson: Nothing changed that requires these arrays to be memory-resident — you're right that the conversion only touches them through the `AbstractDataStore` subscript operator, so OOC stores work fine.

`requireArraysInMemory()` isn't blocking OOC or forcing residency; it's just `setParallelizationEnabled(CheckArraysInMemory(arrays))`. In-memory stores keep the parallel (TBB) execution; if either store is OOC it drops to serial, since the ChunkCache-backed stores aren't thread-safe for concurrent subscript access. OOC inputs still convert correctly, just single-threaded.

I added it as the standard thread-safety guard so this filter matches the other OrientationAnalysis algorithms that parallelize over cell data (`ComputeIPFColors`, `ComputeFZQuaternions`, `ComputeKernelAvgMisorientations`, `ComputeGBCDPoleFigure`, `ComputeFaceIPFColoring` all call it). I'd keep it for consistency and OOC safety.

JDuffeyBQ (approved):

PR #1653, opened 2026-07-03
VV: Fill Bad Data Complete VV and minor patches

nyoungbq:
- Complete V&V reports
- Promoted neighbors to int64 (was int32) to better support OOC data
- Patched edge case overflow bug
- Patched dead preflight code and added a test to validate it
- Cleaned up, and trimmed dead code branches

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


imikejackson:
Pushed a review-fix commit to this branch (`vv/fill_bad_data`) addressing the adversarial review:

- **Infinite-loop fix (the main one):** `phaseFourIterativeFill` looped forever when bad-data voxels remain but none has a positive neighbor to fill from (all-bad volume, or an enclosed bad pocket) — `count` could never reach 0. Added a no-progress guard that stops with a warning and leaves the unfillable voxels unchanged. New test `AllBadData_TerminatesWithoutHang` pins it.
- **Signed/unsigned bug:** the feature-count scan compared a negative `featureName` against unsigned `numFeatures`, sign-extending to a huge value that would blow up the `featureNumber` allocation. Now only positive ids count.
- **Preflight test** now asserts the exact code `-16500` (was invalid-only).
- **V&V paperwork:** reverted the `stamp complete` commit's COMPLETE back to READY FOR REVIEW to match the report body/provenance (second-engineer review still pending), and flagged the **Test 08 A/B evidence** as an open item — `test_08_input.dream3d` and `fill_bad_data_vv/comparison_report.md` are not in the committed `6_5_fill_bad_data.tar.gz` (it holds Tests 01–07, 11, 13 only), so that legacy comparison isn't reproducible yet. It needs to be added to the archive (or the claim dropped) before COMPLETE.

All 13 `FillBadData` TEST_CASEs pass (in-core). Two things for you to close out: the Test 08 evidence archival, and the archived `TEST_PLAN.md`'s Phase-1 connectivity note (it says 26-connectivity, but `phaseOneCCL` uses 6-connectivity / face neighbors).

imikejackson (approved):

imikejackson (approved):

imikejackson (approved):

Issue #1650, opened 2026-07-02
V&V of MTR Pipeline Filters


imikejackson:
# V&V Work Plan

> 🔴 **BEHIND SCHEDULE.** Original deadline was **2026-06-30** — now past. As of **2026-07-01**, all active filters below are **OVERDUE** and not yet started. This is a recovery schedule showing the *earliest realistic* completion, not the committed date.

- **Recovery assumptions:** start **Wed 2026-07-01** 
    - weekends off
    - Fri 2026-07-03 excluded (Independence Day observed)
- engineers work queues sequentially, all in parallel
- **4 working days per filter, except imikejackson at 2 days/filter (~2× throughput)** 
- 5 filters already have PRs in review and are excluded from the recompute.
- **Earliest projected completion:** **2026-07-23**, driven by mmarinebluequartz / jduffeybq / nyoungbq (4 filters × 4 days = 16 working days each). - imikejackson clears 7 filters by **2026-07-21** 
- This is ~3.5 weeks past the 2026-06-30 deadline.

Each item: **C++ class** — display name · plugin · start→due · PR number. Check the box and fill the PR number as each lands.

## PRs

- **WritePoleFigureFilter** — Generate and Write Pole Figure Images · OrientationAnalysis · PR #1647 
    - [x] In Review
    - [x] Merged
- **ComputeFaceIPFColoringFilter** — Compute IPF Colors (Face) · OrientationAnalysis · PR #1646 
    - [x] In Review
    - [x] Merged
- **ComputeFeatureSizesFilter** — Compute Feature Sizes · SimplnxCore · PR #1638 
    - [x] In Review
    - [ ] Merged
- **ConvertOrientationsFilter** — Convert Orientation Representation · OrientationAnalysis · PR #1648 
    - [x] In Review
    - [x] Merged
- **ComputeAvgOrientationsFilter** — Compute Feature Average Orientations · OrientationAnalysis · PR #1645
    - [x] In Review
    - [x] Merged

### mmarinebluequartz — 4 filters @ 4 days · finishes 2026-07-23

- [ ] **CAxisSegmentFeaturesFilter** — Segment Features (C-Axis Misalignment) · OrientationAnalysis · 2026-07-01→07-07 · PR #___
- [ ] **MultiThresholdObjectsFilter** — Multi-Threshold Objects · SimplnxCore · 2026-07-08→07-13 · PR #___
[...]

PR #1655, opened 2026-07-06
VV: Neighbor Orientation Correlation fully V&V'ed

imikejackson:
## Summary

Full V&V of `NeighborOrientationCorrelationFilter` per the v2 MTR V&V policy (part of #1650).

* Oracle: **Class 2** NumPy reference implementation + **Class 1** hand derivations (convention-free co-axial rotation fixtures) + **Class 4** invariants, encoded as 11 inline test fixtures.
* **Fixed 2 bugs** found by oracle reconciliation, both inherited from legacy:
  * **D2** — the level loop double-decremented `currentLevel` (legacy regression from 2014, commit `0bbeb1d49`), running `ceil((6-Level)/2)` passes instead of the documented `6 - Level`.
  * **D3** — `best = 0` was reset inside the neighbor-selection loop (present since the filter's 2013 origin), so the *last* neighbor with any similar pair won instead of the *most similar* neighbor.
* **Documented 4 deviations** from DREAM3D 6.5.171 (`vv/deviations/NeighborOrientationCorrelationFilter.md`): D1 stale-`w` (legacy can copy *different-phase* data into a cell), D2, D3, D4 float/double precision (latent).
* SIMPLNX vs DREAM3D 6.5.171 comparison on 11 legacy-native fixtures: fixed SIMPLNX matches the oracle on all 11; 6.5.171 matches the reference implementation's defect-model prediction on all 11. Each root cause was proven by applying the corresponding surgical fix to a local build of the legacy source, after which the legacy output became **bit-identical** to SIMPLNX.
* **Test hardening / exemplar retirement**: the Small IN100 exemplar comparison had been a hollow pass since 2022 — the v1 archive's container name never matched the test's `Exemplar Data` lookup and the comparison silently skipped every array. Because a regenerated exemplar would be a circular oracle (forbidden by the V&V policy), the archive dependency was removed entirely: the test now snapshots the cell data in memory before the filter runs and asserts the Class 4 invariants at 4.4M cells. **This filter has no exemplar archive**; exact expected outputs live in the 11 inline fixtures. The provenance sidecar records the retirement of both the v1 (hollow) and v2 (circular, never consumed) archives.
* Quality pass: removed dead `neighborDiffCount` work (4 bytes/cell + one wasted misorientation calc per neighbor), dead `RUN_TASK` macro, and the per-task copy of the `bestNeighbor` vector (8 bytes/cell per array) in the transfer stage.
* Docs corrected: *Cleanup Level* is a pass count, not a neighbor-count threshold; the default (6) performs zero passes.

Deliverables: `src/Plugins/OrientationAnalysis/vv/{,deviations/,provenance/}` per the v2 policy. Second-engineer oracle review requested as part of this PR.

## Test Plan
- [x] 14/14 `NeighborOrientationCorrelationFilter` tests pass on in-core build (`simplnx-Rel`)
- [x] 14/14 pass on out-of-core build (`simplnx-ooc-Rel`)
- [x] 11/11 oracle fixtures match the reference implementation via nxrunner A/B

## Update (adversarial + senior-engineer review pass)

* **Tie-break revision (user-approved):** the corrected argmax now resolves ties to the **last** neighbor in scan order (`>=` + count > 0), so fully-tied neighborhoods pick the identical neighbor as 6.5.171. Fixture deviation count vs 6.5.171 dropped from 10/11 to **5/12** — exactly the fixtures where D1–D3 genuinely change the outcome.
* **Production-scale A/B (Small IN100, 4,444,713 cells, Level 2):** 14.29% of cells differ vs 6.5.171 (115,380 filled only by SIMPLNX = D2's restored passes; 496,900 of 999,660 co-modified cells differ = D1/D3 + cascade). The patched local legacy build reproduces SIMPLNX **bit-for-bit on every cell**, bounding D4 (float/double precision) at zero observed effect.
* New deviation **D5**: legacy transfers NeighborList/String cell arrays, SIMPLNX transfers only DataArray types.
* New fixture **F12** (4×5×3 anisotropic dims) closes the stride/axis-swap coverage gap; F04/F10 split; new preflight Level validation (negative Level errors, Level ≥ 6 warns it is a no-op) with tests; cancel checking in the transfer stage; Cleanup Level tooltip and doc corrections.
* The hollow-exemplar claim is now bisect-proven: the comparison was hollow from its 2022-07-24 introduction (`d199bc749`); the archive SHA never changed.
* 17/17 tests pass in in-core and OOC builds.


src/Plugins/OrientationAnalysis/vv/NeighborOrientationCorrelationFilter.md
nyoungbq, line 1: This feels a little overly wordy, but this may be a personal preference so up to you if you want to reword

nyoungbq (approved):

src/Plugins/OrientationAnalysis/vv/NeighborOrientationCorrelationFilter.md
imikejackson: I made some adjustments to make the report clearer and a bit more concise.

PR #1662, opened 2026-07-09
FILT: M3C SurfaceMesh Algorithm added.

imikejackson:
## Summary

Adds a new **Create Surface Mesh (M3C Multi-Material Marching Cubes)** filter — an in-memory port of the legacy DREAM.3D `M3CEntireVolume` algorithm (the functional replacement for the `M3CSliceBySlice` filter, which was removed from DREAM.3D ~10 years ago). It generates a watertight, conformal **Triangle Geometry** with **Face Labels** (`int32`×2) and **Node Types** (`int8`) — the same output data model as the QuickMesh and Surface Nets filters — so it is a drop-in input for the existing downstream mesh filters. Optional per-face transfer of selected Cell/Feature arrays and an optional winding-consistency repair are supported.

Highlights beyond the straight port:

* **Sliding-window sweep** — the volume is meshed one Z slice at a time, so per-site working memory is `O(sliceArea)` instead of `O(volume)`. Peak memory is now dominated by the generated mesh rather than the input dimensions (a 512³ volume no longer needs a ~18 GB scratch floor).
* **Multithreaded by default** — the per-cube work runs across all cores (each cube flips a private edge copy, so no locking). Output is deterministic and independent of thread count. Serial reference paths are kept behind `M3C_SERIAL=1` / `M3C_WHOLE_VOLUME=1` for validation.
* **64-bit site indexing / unsigned 32-bit mesh ids** so large volumes are addressed without overflow.
* **Shared win:** `MeshingUtilities::RepairTriangleWinding` was `O(features × triangles)` (a full triangle rescan per feature); it is now `O(triangles)`. This also speeds up Surface Nets and QuickMesh — on a 100M-triangle mesh the winding pass drops from hours to seconds.
* Legacy **SIMPL pipeline conversion** (`M3CSliceBySlice` → this filter), filter documentation, and a surface-meshing comparison table added to all three mesh-filter docs.

## Test Plan

- [x] Exemplar regression test (Small IN100) — vertices, faces, Face Labels, and Node Types
- [x] 5 toy datasets covering every case-table / saddle path (triple lines, quad points, checkerboard saddle + `treat_anomaly`, single-voxel inclusion, interleaved diagonal)
- [x] Independent mesh-integrity checks (no degenerate triangles, indices in range, ordered Face Labels, valid Node Types)
- [x] Preflight error path (`-90200` tuple mismatch) and RectGrid warning path (`-90210`)
- [x] SIMPL backward-compatibility (UUID + argument mapping)
- [x] Exercised on a real 512³ / 82,725-grain / 109M-triangle volume
- [ ] Verified on an out-of-core build
- [ ] Exemplar published to the DREAM3D Data_Archive (currently a local reference)


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+// The multi-material marching-cubes core operates on flat, 1-based index arrays;
+// the helpers below are transcribed closely from
+//   DREAM3D/Source/Plugins/SurfaceMeshing/SurfaceMeshingFilters/Unsupported/M3CEntireVolume.cpp
+// Only the thin operator() layer touches simplnx (read ImageGeom+FeatureIds,
+// write TriangleGeom+FaceLabels+NodeTypes).
+// =============================================================================
+namespace
+{
+using int64 = std::int64_t;
+using uint64 = std::uint64_t;
```
nyoungbq:
```suggestion
```
Do not redefine fundamental library level scalar types just include `Types.hpp`

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+
+// Index of a padded site / voxel (i.e. an index into the FeatureId grid). MUST be 64-bit: a large
+// Image Geometry can have well over 2^31 voxels, and node ids derived as 7*site must not overflow.
+using SiteId = std::int64_t;
+// Sentinel stored in the compacted-node map for candidate slots that are not real mesh nodes.
+constexpr uint32_t k_UnusedNodeId = std::numeric_limits<uint32_t>::max();
```
nyoungbq: Use library type conventions unless you are doing file format specific I/O operations

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+                {
+                  SiteId tnode = nodeID[ii];
+                  sq[k].FCnode = tnode;
+                  nodeType[tnode] = M3CNodeType::k_QuadPoint;
+                }
+              }
+              else
+              {
+                SiteId tnode = nodeID[ii];
+                if(nodeType[tnode] != -1)
```
nyoungbq:
```suggestion
                if(nodeType[tnode] != M3CNodeType::k_Unused)
```
I can't find a place where the node is truly initialized to `-1`, I think this case should be switched to the initialization value as shown above (k_Unused is `0`)

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+            index++;
+            SiteId tnode = static_cast<SiteId>(e1[cedge].node_id[0]);
+            int tspin = e1[cedge].nSpin[0];
+            e1[cedge].node_id[0] = e1[cedge].node_id[1];
+            e1[cedge].node_id[1] = tnode;
+            e1[cedge].nSpin[0] = e1[cedge].nSpin[1];
+            e1[cedge].nSpin[1] = tspin;
+          }
+        }
+        chaser = burnt_loop[index - 1];
```
nyoungbq, line 1070: I think there might be potential for overflow here. Assume index = numN -1 at the start of new while loop (it passes check at the end), If index gets incremented  three or more times in the nested for loop it will be an out of bounds access to `burnt_loop` since it is only initialized to `numN + 2`. If there is no possible way this could happen the reasoning should be documented here.

Also there may be a risk of an infinite loop here if index cases are never hit at all since the code would essentially be static repeated checks. I do think that this would have to be caused by degenerate data, since the test case passes, but I would definitely double check this and loops similar to this (I saw several here).

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+            if(cspin == arraySpin[kk])
+            {
+              arraySpin[kk] = -1;
+              nburnt++;
+            }
+          }
+        }
+      }
+      (void)nburnt;
+      nodeType[BCnode] = static_cast<int8_t>(nds);
```
nyoungbq: I don't know if this is possible, but `nds` looks like it could exceed `4` here which is the max defined value in `M3CNodeType` in the form of `k_QuadPoint = 4`. In laplacian smoothing anything above 4 (5, 6, 7, and 8 cases) are not checked for. This also applies later where the 10 is added for "surface promotion", where 14 is the upper limit in `SurfaceQuadPoint`. I don't know if a case above 4 is degenerate or impossible but I would add handling or document how its an impossible case.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+  }
+  if(const char* serial = std::getenv("M3C_SERIAL"); serial != nullptr && std::string_view(serial) == "1")
+  {
+    return runWindowed(false);
+  }
+  return runWindowed(true);
+}
+
+// -----------------------------------------------------------------------------
+Result<> M3CSurfaceMeshing::runEntireVolume()
```
nyoungbq, line 2764: This function has a TODO for handling RectGridGeom. I would either add handling for it or disable it as an acceptable input type in params

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+        nTri0 = nTri0 + get_number_case0_triangles(arrayFE.data(), e, nFE);
+      }
+      else if(nFC == 2)
+      {
+        nTri2 = nTri2 + get_number_case2_triangles(arrayFE.data(), e, nFE, arrayFC, nFC);
+      }
+      else if(nFC > 2 && nFC <= 6)
+      {
+        nTriM = nTriM + get_number_caseM_triangles(arrayFE.data(), e, nFE, arrayFC, nFC);
+      }
```
nyoungbq, line 1512: LLM found this:

[BUG] Missing nFC == 1 dispatch — potential mesh holes
  
The cube dispatch in get_number_triangles:1444 and both sweep variants:

```cpp
  if(nFC == 0) → case0
  else if(nFC == 2) → case2
  else if(nFC > 2 && nFC <= 6) → caseM
  // nFC == 1 → SILENT SKIP — zero triangles emitted
```

  A cube with exactly one face containing a triple-line junction (face-center node, case 7/11/13/14) is geometrically valid at a grain-edge that grazes only one face of
  the cube. With nFC == 1 the cube produces no triangles, leaving a gap in the mesh at that location — a non-watertight result. No error is raised.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
```diff
@@ -0,0 +1,3320 @@
...
+
+  const SiteCoords siteCoords{fileDim[0], fileDim[1], fileDim[0] * fileDim[1], {res[0], res[1], res[2]}, {origin[0], origin[1], origin[2]}};
+  const NodeCoords nodeCoords{siteCoords};
+  const NeighborAccessor neighbors{numSites, numSitesPerPlane, xDim};
+
+  // Mesh-scale accumulators grow with the output; only the per-site squares are windowed.
+  std::vector<int8_t> nodeType(static_cast<size_t>(7) * numSites, 0);
+  std::vector<Segment> fedges;
+  std::vector<Triangle> triangles;
+  std::vector<SiteId> mCubeID;
```
nyoungbq, line 2880: Might be worth guessing at a `reserve()` size for these to avoid repeated reallocation on these larger vectors.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Removed. The file now includes `simplnx/Common/Types.hpp` and uses the `nx::core` scalar types directly.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Done — swept the whole algorithm file: `uint32_t`→`uint32`, `int8_t`→`int8`, `int32_t`→`int32`, `std::int64_t`/`std::uint64_t`→`int64`/`uint64`, `size_t`→`usize`. No file-format I/O in this file, so nothing kept the std names.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: You're right that `-1` never occurs — but the legacy code has the identical dead check (`M3CEntireVolume.cpp:305` zero-initializes the node kinds, `:1148` tests `!= -1`), so it was vestigial there too and the branch was always taken.

Switching to `!= k_Unused` would change behavior in a harmful way: a node marked `k_Unused` by a box-surface edge must be re-promoted to `k_Default` when a real edge touches it, otherwise node compaction in `finalizeMesh` (which keeps only `nodeType > 0`) drops a vertex that stored edges/triangles still reference — corrupting the remapped indices.

Fix applied: removed the dead comparison, made the assignment unconditional (identical behavior to legacy and to the code as posted), and documented the reasoning at both copies.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Good catch — the bound holds for well-formed data but only by invariant, and nothing enforced it. For valid label data each do-pass extends the chain by exactly one edge: `endNode` is fixed for the entire pass, a cube's face edges contain no duplicate node pairs (adjacent faces share only one edge-midpoint node and the 2D case tables never emit duplicate edges), and within one spin-pair loop every node has degree <= 2 (the 3-4 edges meeting at a face-center node all carry distinct spin pairs). So `index` steps 1 per pass and never exceeds `numN`. The same argument covers the infinite-loop case: the initial burn phase groups edges by connectivity, so a group can never strand a pass with no extension.

Since that isn't enforceable for degenerate/non-manifold input, all 8 chase loops (count + generate, case2 + caseM, open + closed branches) now have two guards: an in-loop `index >= numN` break that caps writes below `burnt_loop.size()`, and a per-pass no-progress break that bails instead of spinning. Both are unreachable for valid input (verified: exemplar output is unchanged). The invariant is documented at `get_number_case2_triangles`.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Confirmed — it is reachable, not degenerate. Brute-forcing all 4140 set-partitions of a cube's 8 corner labels shows that when `nFC >= 3` the distinct-label count `nds` ranges 3..8, so node types 5-8 (15-18 after the +10 surface promotion) were being written. Legacy wrote the raw count and had the same issue.

Fixed by clamping to `k_QuadPoint` ("4 or more grains") at all three assignment sites, matching what QuickSurfaceMesh does (`QuickSurfaceMesh.cpp:1545`). The clamp changed real output: Small IN100 has thirteen 5-junction body centers, so the exemplar was regenerated and published as `M3CSurfaceMeshingExemplar_v2.tar.gz` (v1 kept in the archive). An independent h5py diff confirms v2 differs from v1 in exactly those 13 of 38914 NodeTypes values (all `5->4` / `15->14`); vertices, faces, and FaceLabels are byte-identical.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Went with disabling it: the geometry parameter now allows `IGeometry::Type::Image` only, matching the SurfaceNets precedent (QuickSurfaceMesh supports RectGrid properly via `getPlaneCoords`, but M3C's on-demand node coordinates assume uniform spacing throughout). The `-90210` warning path and the unit-spacing fallback are removed — both `run*` methods read the `ImageGeom` directly — and the test section now asserts preflight rejection. Docs updated to state RectGrid is not supported and why.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Checked this one exhaustively and it's a false positive: a face contributes a face-center node only for square cases 7/11/13/14/19, which is purely a function of the 4 corner labels. Enumerating all 4140 set-partitions of a cube's 8 corner labels through the exact `get_square_index` logic gives `nFC` in {0, 2, 3, 4, 5, 6} — exactly one face-center face is geometrically impossible (a lone triple-line crossing can't terminate inside the cube without turning on other faces), so the `nFC == 1` gap silently skips nothing. `nFC > 6` can't occur either (6 faces). Legacy has the identical dispatch with the `nFC == 1` else branch commented out.

Rather than adding dead handling, the dispatch now carries a comment documenting the enumeration result so the gap is explicitly justified (it also covers the `else -> caseM` branch in the generate dispatches).

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp
imikejackson: Partially applies: `triangles` and `mCubeID` never grow incrementally — they stay empty through the pass-1 count and are then `resize()`d once to the exact triangle count, so `reserve()` buys nothing there (now documented at the declarations). `fedges` is the only vector that grows via `push_back`; it now gets a one-shot capacity reservation a few slices into pass 1, extrapolated from the per-slice edge rate (+5% margin), in both the serial and parallel sweeps. That avoids the repeated geometric reallocations — and their transient memory spikes — on multi-GB edge lists while never triggering for small volumes.

nyoungbq (approved):

PR #1645, opened 2026-06-30
VV: ComputeAvgOrientations fully V&V'ed

imikejackson:
## Summary
Full Verification & Validation of `ComputeAvgOrientationsFilter` (Compute Feature Average Orientations). The filter is verified correct against an independent oracle and then diff-explained against the official DREAM3D 6.5.171 `FindAvgOrientations`.

* Retire the circular `7_ComputeAvgOrientation_v2` exemplar (it was regenerated from the filter's own output). Replace it with an inline oracle: Class 1 analytical closed-form Rodrigues averages, Class 2 EbsdLib reference for the vMF/Watson methods (reusing `DirectionalStats`' already-tested values rather than re-testing the EM math), and Class 4 invariants.
* Five unit tests cover all three averaging methods plus both preflight error paths (`-54670`, `-651`), closing the prior GCOV vMF/Watson coverage gap.
* Algorithm-review fixes (output-neutral): disable unsafe parallelization in the vMF/Watson worker per the thread-safety policy, add cancel checking and Rodrigues progress messaging, guard the crystal-structure index against out-of-range values, remove dead code.
* Add the V&V report, deviations, and provenance under `src/Plugins/OrientationAnalysis/vv/`.

## Legacy comparison (vs official DREAM3D 6.5.171)
Numerically equivalent on all real features. Divergences are confined to feature-0 / unindexed-voxel handling:
* **D2** — SIMPLNX averages `FeatureId == 0` voxels (phase > 0); legacy skips them. Deliberate algorithmic choice.
* **D3** — empty feature gets a clean identity quaternion; legacy left it `(0,0,0,0)`. Legacy bug.
* **D4** — sub-epsilon precision (`AvgEulerAngles` max diff 4.77e-7) from the EbsdLib quaternion/Euler library swap.
* **D1** — downgraded: the northern-hemisphere canonicalization could not be made to diverge and has no observable effect.

The von Mises-Fisher and Watson methods are new and have no legacy equivalent.

## Test Plan
- [x] All 5 `ComputeAvgOrientations` tests pass on the in-core build (vcpkg EbsdLib 3.0.0)
- [x] Out-of-core build (deferred)

PR #1664, opened 2026-07-09
BUG: Fix ComputeAvgOrientations phase gating + V&V follow-ups

### LINKED ISSUE #1659: BUG: ComputeAvgOrientations vMF/Watson averaging ignores Phases (includes phase-0 voxels)
Found during the adversarial post-merge review of #1645 (ComputeAvgOrientations, already merged to `develop`).

## Problem

In `ComputeAvgOrientations.cpp`, the vMF/Watson averaging path is inconsistent about the `Phases` array between its two passes:

- The per-feature voxel **count** gates on phase (`computeVmfWatsonAverage`, ~lines 331–335): `if(currentPhase > 0) { featureNumVoxels[currentFeatureId]++; ... }`.
- The worker's **gather** loop does **not** gate on phase (~line 99): `if(featureIdsRef[voxelIdx] == featureId)` with no phase check. (The `phasesRef` line that would have enabled it was commented out pre-PR and was removed as dead code rather than wired up.)

Consequence: for a feature containing, e.g., one phase-1 voxel and one phase-0 (unindexed) voxel, `featureNumVoxels == 1` but `fzQuats.size() == 2`. The single-voxel shortcut is bypassed and the EM average runs including the phase-0 garbage quaternion, FZ-reduced with the wrong crystal ops. The Rodrigues path (line ~431) excludes phase-0 voxels correctly, so the two paths disagree.

No current test fixture mixes phase-0 and phase>0 voxels within one feature, so the suite cannot detect this even though the provenance claims to verify "correct per-feature voxel gathering + FZ reduction + phase→crystal-structure lookup."

## Fix

Gate the vMF/Watson gather loop on `phase > 0` to match the counting pass and the Rodrigues path, and add a fixture with a mixed phase-0/phase>0 feature.

Filter: `src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp`

### LINKED ISSUE #1660: TEST: ComputeAvgOrientations Rodrigues oracle never exercises non-trivial crystal symmetry
Follow-up from the adversarial post-merge review of #1645 (ComputeAvgOrientations, merged to `develop`).

## Gap

The Class-1 analytical test for the Rodrigues averaging path uses only **Triclinic** symmetry (`k_Triclinic = 4`), which by the report's own admission makes `getNearestQuat` a sign-pick no-op. The filter's actual value-add on the Rodrigues path is the running-average / symmetry-reduction interaction (reset-to-identity on the first voxel, `getNearestQuat(curAvgQuat, voxQuat)` against the evolving mean under the crystal's symmetry operators, count weighting). Under Triclinic none of that is exercised.

Cubic (24-operator) correctness of that interaction is currently validated **only** by the legacy 6.5.171 A/B comparison — which the project's own `docs/vv_templates/oracle_classes.md` (edited in the same PR) classifies as "useful for diff explanation, not for correctness."

## Suggested fix

Add a Class-4 invariant fixture: feed the same physical orientation as several cubic-equivalent representations (q, S·q, −q for symmetry operators S) into one feature; the average must equal that orientation (within tolerance). This exercises the cubic symmetry-reduction branch with an implementation-independent expectation.

Filter: `src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp`
Test: `src/Plugins/OrientationAnalysis/test/ComputeAvgOrientationsTest.cpp`

### LINKED ISSUE #1661: REV: ComputeAvgOrientations V&V follow-ups (guards, tolerances, provenance)
Minor follow-ups from the adversarial post-merge review of #1645 (ComputeAvgOrientations, merged to `develop`). Grouped here to avoid issue spam; split out if any warrants its own PR.

Filter: `src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp`, test `.../test/ComputeAvgOrientationsTest.cpp`, deliverables under `.../vv/`.

- **Untested new guards + silent drop.** The two out-of-range crystal-structure guards added by the PR (`ComputeAvgOrientations.cpp:85-88` and `:435-438`) are not exercised (both fixtures put `999` only at ensemble index 0, never dereferenced), and they silently drop voxels/features with an unknown crystal structure — no warning is emitted and the docs don't say so. Add coverage + a warning, or document the drop.
- **Phase index unguarded (OOB read).** `xtalRef[phaseValue]` (~:82) and `crystalStructures[currentPhase]` (~:433) are not range-checked; a cell `Phases` value ≥ the ensemble tuple count is an out-of-range read. Preflight's `-651` validates tuple counts, not phase values.
- **Tolerance provenance mismatch.** `vv/provenance/ComputeAvgOrientationsFilter.md` states quaternion ±1e-6 / kappa ±1e-4 relative, but the test uses `k_QuatMargin = 5.0e-3f` and `k_KappaRel = 0.02`. Reconcile the doc to the shipped tolerances (or tighten the test) and justify the 2% kappa margin.
- **Documented assertions that don't exist.** The provenance claims a "Euler invariant (Φ≈0, all finite)" check for z-rotation fixtures and a "voxel-ordering independence (verified)" reordered-fixture test; neither is encoded in the test. Either add them or remove the claims.
- **`featureIdToPhaseMap` last-writer-wins** for a multi-phase feature (vMF/Watson uses one phase per feature; Rodrigues is per-voxel) — undocumented; note it.
- **Report self-contradictions:** "4/4" vs 5 tests; Oracle section "OOC dual-build confirmed in Phase 8" vs V&V-phase "OOC build skipped"; a never-written `vv/provenance/7_ComputeAvgOrientation_v2.md` reference. Clean up.
- **Error tests don't assert codes** (`-54670` no-method-enabled, `-651`); and a duplicate `-651` test now coexists after #1644. Assert codes and de-duplicate.
- **QUESTION:** preflight passes with all three averaging methods disabled, deferring to a guaranteed runtime `-54670`. Consider making it a preflight error so the GUI surfaces it before execution.

imikejackson:
Addresses the three follow-up issues from the adversarial post-merge review of PR #1645.

Fixes #1659:
* Gate the vMF/Watson voxel-gather loop on phase > 0 to match the counting pass and the Rodrigues path. Previously a phase-0 voxel inside a feature contributed a garbage quaternion to the EM average and defeated the single-voxel shortcut. Pinned by a new regression test with mixed phase-0/phase-1 features.

Closes #1660:
* Add a Class-4 invariant test that exercises the Rodrigues symmetry-reduction branch under Cubic_High: five representations of the same orientation (Rz(30+90k), -Rz(30)) must average to Rz(30). Cubic correctness no longer rests solely on the legacy A/B diff.

Closes #1661:
* Range-check the Phases value against the CrystalStructures tuple count in both averaging paths (previously an out-of-range read).
* Report dropped voxels/features as warnings (-54671 unknown crystal structure, -54672 out-of-range phase) instead of silently skipping; document the behavior in the filter docs.
* Reject the all-methods-disabled configuration in preflight (-54673) instead of deferring to a guaranteed runtime failure; keep the runtime -54670 backstop and pin it via direct algorithm invocation.
* Assert error codes in the error tests and remove the duplicate -651 TEST_CASE introduced by PR #1644.
* Encode the Euler invariant (Phi ~= 0, finite) checks and the voxel-ordering-independence fixture the provenance claimed.
* Reconcile the provenance tolerances to the shipped values (quat 5e-3, kappa 2%) with justification; document the last-writer-wins featureIdToPhaseMap behavior for multi-phase features; fix the V&V report self-contradictions (test count, OOC status, dangling provenance reference).

### DIFF
--- a/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp
+++ b/src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp
@@ -34,6 +34,7 @@ public:
   {
     // Input FeatureIds + Input Orientations. All these should come from the same Attribute Matrix or have the same number of tuples
     Int32AbstractDataStore& featureIdsRef = m_DataStructure.getDataRefAs<Int32Array>(m_InputValues->cellFeatureIdsArrayPath).getDataStoreRef();
+    Int32AbstractDataStore& phasesRef = m_DataStructure.getDataRefAs<Int32Array>(m_InputValues->cellPhasesArrayPath).getDataStoreRef();
     Float32AbstractDataStore& quatsRef = m_DataStructure.getDataRefAs<Float32Array>(m_InputValues->cellQuatsArrayPath).getDataStoreRef();
     // Ensemble Level Data
     UInt32AbstractDataStore& xtalRef = m_DataStructure.getDataRefAs<UInt32Array>(m_InputValues->crystalStructuresArrayPath).getDataStoreRef();
@@ -61,6 +62,7 @@ public:
     }
 
     usize numVoxels = featureIdsRef.getNumberOfTuples();
+    const usize numEnsembles = xtalRef.getNumberOfTuples();
 
     std::vector<ebsdlib::LaueOps::Pointer> ops = ebsdlib::LaueOps::GetAllOrientationOps();
 
@@ -95,8 +97,12 @@ public:
       // they want to find the average orientation of
       for(usize voxelIdx = 0; voxelIdx < numVoxels; voxelIdx++)
       {
-        // If the feature Id of the voxel matches the current feature Id, then grab that orientation
-        if(featureIdsRef[voxelIdx] == featureId)
+        // If the feature Id of the voxel matches the current feature Id, then grab that orientation.
+        // The phase gate MUST match the counting pass in computeVmfWatsonAverage() (and the
+        // Rodrigues path): phase-0/unindexed voxels are excluded from the average (issue #1659),
+        // and an out-of-range phase index would read past the CrystalStructures array (issue #1661).
+        const int32 voxelPhase = phasesRef[voxelIdx];
+        if(featureIdsRef[voxelIdx] == featureId && voxelPhase > 0 && static_cast<usize>(voxelPhase) < numEnsembles)
         {
           const ebsdlib::QuatD q1(quatsRef[voxelIdx * 4], quatsRef[voxelIdx * 4 + 1], quatsRef[voxelIdx * 4 + 2], quatsRef[voxelIdx * 4 + 3]);
           fzQuats.push_back(op->getFZQuat(q1)); // Fundamental Zone Reduction
@@ -276,16 +282,18 @@ Result<> ComputeAvgOrientations::operator()()
 
   MessageHelper messageHelper(m_MessageHandler);
 
-  Result<> result;
+  // Warnings (e.g. dropped voxels/features) from each path are merged and returned together.
+  Result<> finalResult;
   if(m_InputValues->useRodriguesAverage)
   {
     messageHelper.sendMessage("Computing Rodrigues Average Orientations");
 
-    result = computeRodriguesAverage();
+    Result<> result = computeRodriguesAverage();
     if(result.invalid())
     {
       return result;
     }
+    finalResult = MergeResults(std::move(finalResult), std::move(result));
   }
   if(m_InputValues->useVonMisesAverage || m_InputValues->useWatsonAverage)
   {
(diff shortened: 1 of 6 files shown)


src/Plugins/OrientationAnalysis/src/OrientationAnalysis/Filters/Algorithms/ComputeAvgOrientations.cpp
```diff
@@ -318,23 +327,50 @@ Result<> ComputeAvgOrientations::computeVmfWatsonAverage()
...
+  usize unknownXtalFeatureCount = 0;
+  {
+    const std::vector<ebsdlib::LaueOps::Pointer> ops = ebsdlib::LaueOps::GetAllOrientationOps();
+    for(const auto& [featureId, phaseValue] : featureIdToPhaseMap)
+    {
+      if(crystalStructures[phaseValue] >= ops.size())
+      {
+        unknownXtalFeatureCount++;
+      }
+    }
```
nyoungbq:
```suggestion
    const usize numValidXtal = ebsdlib::LaueOps::GetAllOrientationOps().size();
    for(const auto& [featureId, phaseValue] : featureIdToPhaseMap)
    {
      if(crystalStructures[phaseValue] >= numValidXtal)
      {
        unknownXtalFeatureCount++;
      }
    }
```
No need to size check a static vector in a loop or copy the entire vector into stack, just store the size since it gets freed right after counting

nyoungbq (approved):

PR #1667, opened 2026-07-09
BUG: Make preflight geometry import metadata-only

joeykleingers:
## Summary

During a metadata-only preflight, `DataStructureReader` now leaves geometry connectivity — vertex/face/edge lists, derived connectivity, and rectilinear-grid bounds — as an `EmptyDataStore` carrying only shape and type (exactly like ordinary attribute arrays), instead of eagerly bulk-reading it from disk. This removes a wasted per-geometry disk read from every preflight, which is costly on high-latency storage such as network mounts. Execute is unchanged — it imports the real connectivity stores directly in `readGroup`.

* `readGroup` already imports every array as an `EmptyDataStore` when `useEmptyDataStores` is set, so `loadRequiredData` is now skipped during preflight.
* Geometry IO readers (`EdgeGeomIO`, `INodeGeom0d/1d/2d/3dIO`, `RectGridGeomIO`) no longer register connectivity as required data — those ids were only ever consumed by the eager preflight read.
* Forward `useEmptyDataStores` from the `std::filesystem::path` overload of `DataStructureReader::ReadFile` to the `FileIO` overload; it was previously dropped and always defaulted to `false`.

## Test Plan

- [ ] `DREAM3DFileTest` passes (adds preflight metadata-only coverage asserting geometry connectivity stores are empty after preflight and populated after execute)
- [ ] Existing DREAM3D read/write and geometry tests pass

### DIFF
--- a/src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
+++ b/src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
@@ -10,6 +10,9 @@
 #include "simplnx/DataStructure/DataGroup.hpp"
 #include "simplnx/DataStructure/DataStore.hpp"
 #include "simplnx/DataStructure/DataStructure.hpp"
+#include "simplnx/DataStructure/IDataArray.hpp"
+#include "simplnx/DataStructure/IDataStore.hpp"
+#include "simplnx/DataStructure/IO/HDF5/DataStructureReader.hpp"
 #include "simplnx/Filter/Arguments.hpp"
 #include "simplnx/Filter/FilterHandle.hpp"
 #include "simplnx/Parameters/Dream3dImportParameter.hpp"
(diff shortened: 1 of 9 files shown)


imikejackson:
I pushed a3fcfc7b2 to this branch after reviewing — two things, no change to the approach:

* **Removed the now-dead required-data machinery from `DataStructureReader`.** With the geometry IO readers no longer registering connectivity ids, nothing calls `addRequiredId`/`addRequiredPath` anymore, so `loadRequiredData` was a guaranteed no-op in every code path — including the `!useEmptyDataStores` call this PR kept, which implied the pass still mattered for full reads. Deleted `addRequiredPath`, both `addRequiredId` overloads, `loadRequiredData`, and the `m_RequiredPaths`/`m_RequiredIds` members. The static `FinishImportingObject` stays; `Dream3dIO` uses it for the per-path execute import.

* **Broadened the new test to cover every reader this PR touched.** It previously only exercised TriangleGeometry (`INodeGeom2dIO`); it's now table-driven over geoms.dream3d's edge, triangle, quad, tetrahedral, and hexahedral geometries plus the rect-grid X/Y/Z bounds, with a `DYNAMIC_SECTION` per connectivity array in all three blocks (preflight, execute, `ReadFile` path overload).

Also chased the two riskiest paths and found no regressions: no `preflightImpl` in simplnx or downstream plugins reads connectivity *values* (only shape/type, which `EmptyDataStore` preserves — value access throws loudly), and execute through `ImportH5ObjectPathsAction` is unaffected because connectivity arrays are ordinary paths in `m_Paths` and get their real stores from `DataArrayIO::finishImportingData`.

Verified on both builds: full SimplnxCore + core-library suites pass 100% on in-core (`simplnx-Rel`) and out-of-core (`simplnx-ooc-Rel`).

imikejackson (approved):

PR #1668, opened 2026-07-09
ENH: Cache preflight metadata for .dream3d imports

joeykleingers:
## Summary

Adds `Dream3dPreflightCache`, a process-wide cache of preflight (metadata-only) `DataStructure`s imported from `.dream3d` files, and routes the two preflight consumers (`ReadDREAM3DFilter` and `ImportH5ObjectPathsAction`) through it. A pipeline re-preflights on every parameter edit, and each pass previously re-traversed the input file's full HDF5 metadata tree — hundreds of small, latency-bound reads that freeze the UI for seconds on high-latency storage such as network mounts. After the first import, every subsequent fetch of an unchanged file becomes a single filesystem `stat`.

**The cache.** Keyed by canonical file path and validated on every fetch by comparing the file's `(size, mtime)` against the values captured when the entry was populated; any mismatch evicts and re-reads. Files modified more recently than a 2-second trust window are never served, guarding against coarse mtime granularity on SMB/NFS. `fetch()` never hands out the cached master — it returns a copy whose array stores are freshly allocated, so a mutation in one preflight pass can never bleed into another; metadata-only stores keep these copies free of bulk data. The table is bounded (capacity 8, LRU by fetch recency). Two mutexes with a fixed lock order keep hits lock-light while serializing the HDF5 metadata reads (the bundled HDF5 C library is built without its thread-safe option).

**Consumers.** `ReadDREAM3DFilter::preflightImpl` and `ImportH5ObjectPathsAction::apply` now fetch from the cache instead of opening the file every pass. `ImportH5ObjectPathsAction` opens the file only in execute mode, where the bulk-array reads actually need a handle.

**Fewer redundant opens.** `Dream3dImportParameter::validatePath` no longer re-opens the file as HDF5 to prove readability (a redundant round-trip on every edit for a check the consuming filter's preflight already makes) — it now only stats the path. `FinishImportingObject` is split so its metadata-only insert (`FinishImportingObjectPreflight`) needs no open file handle.

**Test instrumentation.** `FileIO` gains a process-wide `ReadFile` open counter and a test-only fapl-configurator hook (unset in production, so opens use `H5P_DEFAULT`), letting tests assert I/O behavior deterministically instead of against wall-clock time.

## Test Plan

- [ ] `Dream3dPreflightCache` unit tests pass: miss-then-hit, missing-file error passthrough, handout isolation, mtime/size invalidation, trust window, `invalidate()`, LRU eviction, concurrency hammer
- [ ] `DREAM3DFileTest` integration test passes: a real `ReadDREAM3DFilter` pipeline performs zero file opens on the second preflight, execute still reads bulk data, and a rewrite invalidates the entry
- [ ] `DelayVfd` `[.benchmark]` demonstrates the before/after under simulated storage latency

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -606,6 +606,7 @@ set(SIMPLNX_HDRS
   ${SIMPLNX_SOURCE_DIR}/Utilities/Meshing/TriangleUtilities.hpp
 
   ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/DREAM3D/Dream3dIO.hpp
+  ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/DREAM3D/Dream3dPreflightCache.hpp
 
   ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/HDF5/H5.hpp
   ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/HDF5/H5Support.hpp
@@ -803,6 +804,7 @@ set(SIMPLNX_SRCS
   ${SIMPLNX_SOURCE_DIR}/Utilities/Meshing/TriangleUtilities.cpp
 
   ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/DREAM3D/Dream3dIO.cpp
+  ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/DREAM3D/Dream3dPreflightCache.cpp
 
   ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/HDF5/H5.cpp
   ${SIMPLNX_SOURCE_DIR}/Utilities/Parsing/HDF5/H5Support.cpp
--- a/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadDREAM3DFilter.cpp
+++ b/src/Plugins/SimplnxCore/src/SimplnxCore/Filters/ReadDREAM3DFilter.cpp
@@ -6,7 +6,7 @@
 #include "simplnx/Parameters/StringParameter.hpp"
 #include "simplnx/Pipeline/Pipeline.hpp"
 #include "simplnx/Utilities/Parsing/DREAM3D/Dream3dIO.hpp"
-#include "simplnx/Utilities/Parsing/HDF5/IO/FileIO.hpp"
+#include "simplnx/Utilities/Parsing/DREAM3D/Dream3dPreflightCache.hpp"
 
 #include "simplnx/Utilities/SIMPLConversion.hpp"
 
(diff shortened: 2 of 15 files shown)


imikejackson:
Performed an adversarial review of this PR against a local Release build (macOS/AppleClang). **No bugs found — nothing was changed or pushed to the branch.**

Correctness claims verified against the codebase:

* The handout-isolation contract holds in **both** copy paths: `DataStructure`'s copy constructor *and* copy assignment (used in `fetch()`'s insert path) rebuild `m_DataObjects` via `DataObject::shallowCopy()`, so a handout shares only store pointers with the cached master, and `RefreshStores()` severs exactly that link. `DataArray`/`StringArray`/`NeighborList` is the complete set of store-bearing `IArray` types; geometry-owned arrays are reachable through `getAllDataPaths()`, and `ScalarData` holds its value by member so `shallowCopy` already isolates it.
* Both v8 (`useEmptyDataStore`) and legacy v7 (`createLegacyDataArray`, `readLegacyStringArray`) preflight imports produce metadata-only stores, so cached masters stay kilobyte-sized for either file version.
* Locking: consistent `m_ReadMutex` → `m_Mutex` ordering with no inverse path; every `fetch()` outcome increments exactly one of hits/misses; the just-inserted entry can never be its own LRU victim, and `std::map::erase` of other nodes leaves the held `Entry&` valid.
* Staleness: write-then-read within one pipeline run, same-size rewrites, coarse (2 s) mtime granularity, and cross-test file reuse all resolve to a re-read rather than a stale serve. Future-dated mtimes (clock skew on NFS) degrade to permanent re-reads — fail-safe in the correct direction.
* The `FinishImportingObject` split is behavior-preserving, and execute-mode `ImportH5ObjectPathsAction` still reads bulk data from a direct file open (verified via the integration test's post-execute value check).

Local test results: all 9 `Dream3dPreflightCache` unit tests pass, `DREAM3DFileTest: PreflightCache` passes, all 24 DREAM3D-related ctest cases pass, and the hidden `[.benchmark]` shows 928 ms cold vs 0.7 ms warm at 2 ms/op simulated latency (~1300x).

Two observations, no action required:

1. In `fetch()`, the `std::error_code` written by `fs::file_size` is overwritten by `fs::last_write_time` (which clears it on success), so a directory path slips past the stat guard with a garbage size token. Traced the outcome: the HDF5 open then fails cleanly with -25 and failures are never cached, so it is unreachable as a defect.
2. `ImportH5ObjectPathsAction::apply` now returns `{}` instead of propagating import warnings. Since warnings could only ever surface on a cache miss (a hit has none to replay), dropping them uniformly is the more consistent behavior — just noting the change.

imikejackson (approved):

PR #1670, opened 2026-07-13
BUG: Fix periodic feature-centroid model in Triangle and Feature centroid filters

### LINKED ISSUE #1665: ComputeTriangleGeomCentroids: periodic centroid uses a constant-offset approximation (same defect class fixed in ComputeFeatureCentroids)
## Summary

`GeometryHelpers::Topology::AdjustCentroidsForPeriodicFaces(const BoundingBox3Df&, const BoundingBoxFaces&, ...)` — used by **`ComputeTriangleGeomCentroids`** — computes a periodic (wrap-aware) centroid by adding a **constant offset** to the naive centroid:

```cpp
// src/simplnx/Utilities/GeometryHelpers.cpp
centroids[3 * featureId + 0] += std::abs(maxPoint[0] - minPoint[0]) / 2.0f;  // and y, z
```

This is the same class of defect that was found and fixed for **`ComputeFeatureCentroids`** during its V&V (see the `ImageGeom` overload, now removed). A constant offset cannot compute a correct wrapped centroid: the true minimum-image centroid depends on **where the feature's mass actually sits**, not merely that the feature spans/touches opposing periodic faces. Two features that trip the same face condition but have different point distributions require different corrections, so no single offset is correct for both.

## Why it's wrong

For a feature whose vertices wrap across a periodic boundary, the correct centroid is the **minimum-image (circular) mean** — unwrap the points at their largest empty gap (equivalently, average the unit vectors of their angular positions around the domain) and map the result back into the domain. Adding `|max - min| / 2` instead lands the centroid in the empty middle of the wrapped feature, offset by an amount that is only coincidentally right for symmetric cases.

Note: unlike the `ComputeFeatureCentroids` `ImageGeom` overload (which also had a cell-units-vs-physical spacing bug), this BoundingBox overload works in physical coordinates, so it does **not** have a spacing bug — the flaw here is purely the constant-offset model.

## Suggested fix

Apply the same treatment used in `ComputeFeatureCentroids`: replace the constant offset with a wrap-aware minimum-image / circular mean for any axis the feature spans, keeping the arithmetic mean for non-spanning axes and an arithmetic fallback for degenerate (domain-filling) features. For triangle geometry the per-feature vertex set is available, so the largest-gap method can be applied directly.

## Reference

- Fixed analog: `ComputeFeatureCentroids` periodic path (reimplemented as circular mean; the wrong `ImageGeom` overload of `AdjustCentroidsForPeriodicFaces` was removed).
- Affected code: `src/simplnx/Utilities/GeometryHelpers.cpp` (`AdjustCentroidsForPeriodicFaces` BoundingBox overload) and its caller `src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeTriangleGeomCentroids.cpp`.
- Affected users: only those enabling the `Is Periodic` option on `ComputeTriangleGeomCentroids`; the default is off.

## Suggested scope

Fold this into a dedicated V&V of `ComputeTriangleGeomCentroids` (independent oracle for the periodic case, e.g. hand-derived minimum-image values on a small wrapping mesh), mirroring the `ComputeFeatureCentroids` V&V.

_Surfaced by the ComputeFeatureCentroids V&V / adversarial review. Filed with Claude Code._

imikejackson:
## Summary

Fixes the periodic (wrap-aware) centroid computation in both `ComputeTriangleGeomCentroids` and `ComputeFeatureCentroids`. Both used a constant half-domain offset added to the naive centroid, which is only correct when a wrapped feature's mass is symmetric about the seam and can place asymmetric wrapped features outside the domain.

Fixes #1665.

* **ComputeTriangleGeomCentroids** — reimplement the `AdjustCentroidsForPeriodicFaces` `BoundingBox` overload as a largest-empty-gap minimum-image mean computed from the feature's vertex coordinates (in-domain wrap-back; arithmetic-mean fallback for domain-filling features).
* **ComputeTriangleGeomCentroids** — fix a latent `std::bad_function_call`: the algorithm invoked `m_MessageHandler.m_Callback()` directly, bypassing the empty-callback guard; it threw whenever the periodic branch fired without an installed handler. Also add cancel checks to both loops and remove a per-feature `std::set` copy.
* **ComputeFeatureCentroids** — reimplement the periodic path as a minimum-image circular mean accumulated during the cell sweep (per-feature unit-vector sums; trig only evaluated when `Is Periodic` is enabled). Remove the range-based `ImageGeom` overload of `AdjustCentroidsForPeriodicFaces`, whose per-feature index ranges are mathematically insufficient to compute a wrapped centroid.
* Replace the prior periodic test fixtures (which validated the constant-offset model against itself — a circular oracle) with independent Class 1 minimum-image oracle values, add a domain-filling fallback fixture, and add V&V deliverables (report, deviations, provenance) for `ComputeTriangleGeomCentroids`.
* The periodic option is SIMPLNX-only in both filters (legacy `FindTriangleGeomCentroids` / `FindFeatureCentroids` have no periodic mode), so no legacy behavior changes. The non-periodic paths are unchanged.

## Test Plan
- [x] Tests pass on in-core build (`SimplnxCore::ComputeTriangleGeomCentroids`, `SimplnxCore::ComputeFeatureCentroids`)
- [x] Tests pass on out-of-core build (`simplnx-ooc-Rel`)
- [x] Adjacent `GeometryHelpers` consumers regression-checked (`SharedFeatureFace`)

New oracle tests include the discriminating asymmetric-wrap case: the old code produced an out-of-bounds centroid (`4.625` on a `[0,4]` domain) where the correct minimum-image value is `3.625`.


src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureCentroids.cpp
```diff
@@ -4,10 +4,21 @@
...
 #include "simplnx/Utilities/ParallelDataAlgorithm.hpp"
 
 #include <algorithm>
+#include <array>
+#include <cmath>
+
+namespace
+{
+// 2*pi, used to map cell-center positions onto the periodic unit circle.
+constexpr double k_TwoPi = 6.283185307179586;
```
JDuffeyBQ: Can be replaced with `nx::core::Constants::k_2Pi`

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureCentroids.cpp
```diff
@@ -169,6 +206,21 @@ Result<> ComputeFeatureCentroids::operator()()
   compensation.fill(0.0);
   count.fill(0.0);
 
+  // Per-feature/per-axis unit-vector sums for the periodic (circular-mean) centroid. Only populated when
+  // Is Periodic is enabled; harmless (all zero) otherwise.
+  auto sumCosPtr = DataStoreUtilities::CreateDataStore<float64>(tupleShape, componentShape, IDataAction::Mode::Execute);
+  auto sumSinPtr = DataStoreUtilities::CreateDataStore<float64>(tupleShape, componentShape, IDataAction::Mode::Execute);
```
JDuffeyBQ:
```suggestion
  auto* sumCosPtr = DataStoreUtilities::CreateDataStore<float64>(tupleShape, componentShape, IDataAction::Mode::Execute);
  auto* sumSinPtr = DataStoreUtilities::CreateDataStore<float64>(tupleShape, componentShape, IDataAction::Mode::Execute);
```
Using `auto*` more clearly indicates that this is a pointer type.

Additionally these arrays are only used when `IsPeriodic` is true, but they are unconditionally created.

JDuffeyBQ (changes requested):

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureCentroids.cpp
imikejackson: Replaced the local constant with `Constants::k_2Pi<double>` in commit 4991ad802.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureCentroids.cpp
imikejackson: `CreateDataStore` returns `std::shared_ptr`, so `auto*` cannot bind to its result. Commit 4991ad802 uses explicit `std::shared_ptr` ownership, passes nullable raw pointers to the worker, and allocates the two stores only for periodic execution.

JDuffeyBQ (changes requested):
Needs rebase

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ComputeFeatureCentroids.cpp
```diff
@@ -146,37 +164,55 @@ Result<> ComputeFeatureCentroids::operator()()
...
+        if(resultant < k_DegenerateResultant)
+        {
+          continue;
+        }
+        double phase = std::atan2(sumSin[axisIdx], sumCos[axisIdx]);
+        if(phase < 0.0)
+        {
+          phase += Constants::k_2Pi<double>;
+        }
+        centroidsBuf[axisIdx] = static_cast<float32>(static_cast<double>(origin[axis]) + (phase / Constants::k_2Pi<double>)*domainLength[axis]);
```
JDuffeyBQ:
```suggestion
        centroidsBuf[axisIdx] = static_cast<float32>(static_cast<double>(origin[axis]) + (phase / Constants::k_2PiD) * domainLength[axis]);
```
I think this will fix the clang-format. From my testing, it seems like the presence of the `<double>` cause clang-format to misinterpret the expression.

Issue #1642, opened 2026-06-26
ReadDREAM3DFilter crashes when a Geometry is nested inside a DataGroup


imikejackson:
## Summary

`ReadDREAM3DFilter` (and the DREAM3D-NX GUI) crashes when reading a `.dream3d` file that
contains a **Geometry nested inside a `DataGroup`**. The same geometry written at the top
level of the `DataStructure` reads back fine. `WriteDREAM3DFilter` writes the nested-geometry
file without any error, so this is a write/read asymmetry.

## Environment

- simplnx Python bindings (conda env), HDF5 1.14.6
- Reproduced both via `ReadDREAM3DFilter` from Python and by opening the file in DREAM3D-NX

## Steps to reproduce

```python
import numpy as np, simplnx as nx

ds = nx.DataStructure()
v = np.array([[0,0,0],[1,0,0],[0,1,0]], dtype=np.float32)
t = np.array([[0,1,2]], dtype=np.uint64)

def make_triangle(path):
    nx.CreateDataArrayFilter.execute(ds, numeric_type_index=nx.NumericType.float32,
        component_count=3, tuple_dimensions=[[3]], output_array_path=nx.DataPath("V"),
        initialization_value_str='0')
    ds[nx.DataPath("V")].npview()[:] = v
    nx.CreateDataArrayFilter.execute(ds, numeric_type_index=nx.NumericType.uint64,
        component_count=3, tuple_dimensions=[[1]], output_array_path=nx.DataPath("T"),
        initialization_value_str='0')
    ds[nx.DataPath("T")].npview()[:] = t
    nx.CreateGeometryFilter.execute(ds, array_handling_index=1, geometry_type_index=4,
        output_geometry_path=nx.DataPath(path),
        face_attribute_matrix_name="Face Data", edge_attribute_matrix_name="Edge Data",
        vertex_attribute_matrix_name="Vertex Data",
        vertex_list_path=nx.DataPath("V"), triangle_list_path=nx.DataPath("T"))

make_triangle("TopTri")                                   # top-level geometry
nx.CreateDataGroupFilter.execute(ds, data_object_path=nx.DataPath("Grp"))
make_triangle("Grp/NestTri")                              # geometry inside a DataGroup

nx.WriteDREAM3DFilter.execute(ds, export_file_path="/tmp/cmp.dream3d", write_xdmf_file=False)
# ^ writes with no errors

# Read it back:
ds2 = nx.DataStructure()
[...]

PR #1683, opened 2026-07-23
VV: WriteDREAM3DFilter Fully V&V'ed

mmarineBlueQuartz:
* Add V&V files for WriteDREAM3DFilter
* Expanded round-trip testing for additional DataObject types.
* New tests for untested code paths.

## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup


nyoungbq (approved):

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -275,11 +534,93 @@ DREAM3D::FileData CreateFileData()
...
+  SIMPLNX_RESULT_REQUIRE_INVALID(writeResult);
+}
+
 TEST_CASE("DREAM3DFileTest:DREAM3D File IO Test", "[WriteDREAM3DFilter]")
 {
   UnitTest::LoadPlugins();
 
   std::lock_guard<std::mutex> lock(m_DataMutex);
+
+  bool writeXdmf = GENERATE(true, false);
```
imikejackson, line 982: **Blocking.** `writeXdmf` is generated but never used — the write below goes through `DREAM3D::WriteFile(fileWriter, fileData)`, which has no xdmf parameter. The entire TEST_CASE body now runs twice identically for zero added coverage. Either route the flag through an overload that accepts it, or delete the GENERATE.

This also invalidates the V&V report's Path 17 row, which cites this test's `writeXdmf=true` as covering the xdmf-rename path. Path 17 is actually covered only via the in-pipeline tests (`CreateExportPipeline` sets `write_xdmf_file: true`).

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -10,9 +10,11 @@
 #include "simplnx/DataStructure/DataGroup.hpp"
 #include "simplnx/DataStructure/DataStore.hpp"
 #include "simplnx/DataStructure/DataStructure.hpp"
+#include "simplnx/DataStructure/DynamicListArray.hpp"
```
imikejackson: **Blocking (dead code).** This include is unused, as are `Constants::k_DynamicListArray` (line 66) and `Constants::k_ListCount` (line 74). This looks like an abandoned DynamicListArray round-trip attempt — and indeed there is no `DynamicListArrayIO` in `src/simplnx/DataStructure/IO/HDF5/`, so the type cannot be serialized at all. Remove the dead include/constants, and record the "DynamicListArray has no HDF5 IO" fact explicitly in the V&V report — that is exactly the kind of gap V&V exists to surface, not to silently drop.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -143,6 +165,188 @@ fs::path GetReMultiExportDataPath()
...
+void CheckTestDataStructure(const DataStructure& dataStructure)
+{
+  DataPath dataGroupPath({Constants::k_DataContainer});
+  REQUIRE(dataStructure.getDataAs<DataGroup>(dataGroupPath) != nullptr);
+
+  const auto* neighborList = dataStructure.getDataAs<Int16NeighborList>(dataGroupPath.createChildPath(Constants::k_NeighborList));
+  REQUIRE(neighborList != nullptr);
+  const auto storePtr = neighborList->getStore();
+  REQUIRE(storePtr != nullptr);
+  REQUIRE(storePtr->getNumberOfTuples() == 6);
```
imikejackson: **Blocking.** Only the tuple count is asserted here — the `{1, 2}` list values set in `CreateTestDataStructure` are never compared after round-trip, so NeighborList data corruption would pass this test. The report claims `CheckTestDataStructure` "asserts exact values against the hand-known fill pattern" for every DataObject kind; that is not true for NeighborList. Add per-list value assertions.

Also: this `6` (and the StringArray count below) should be `Constants::k_ListCount`, which is currently defined and never used.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -143,6 +165,188 @@ fs::path GetReMultiExportDataPath()
...
+  REQUIRE(imageGeom != nullptr);
+  auto dims = imageGeom->getDimensions();
+  REQUIRE(dims[0] == Constants::k_TupleShape[0]);
+  REQUIRE(dims[1] == Constants::k_TupleShape[1]);
+  REQUIRE(dims[2] == Constants::k_TupleShape[2]);
+  const auto* cellData = dataStructure.getDataAs<AttributeMatrix>(imageGeomPath.createChildPath(Constants::k_CellData));
+  REQUIRE(cellData != nullptr);
+  REQUIRE(imageGeom->getCellDataId() == cellData->getId());
+  REQUIRE(cellData->getShape() == Constants::k_TupleShape);
+};
```
imikejackson: Nit: stray semicolon after the function body.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -50,6 +52,26 @@ const fs::path k_ExportFilename2 = "export2.dream3d";
...
+constexpr StringLiteral k_TriangleGeom = "TriangleGeom";
+constexpr StringLiteral k_VertexGeom = "VertexGeom";
+
+constexpr StringLiteral k_DynamicListArray = "DynamicList";
+constexpr StringLiteral k_NeighborList = "NeighborList";
+constexpr StringLiteral k_StringArray = "String Array";
+constexpr StringLiteral k_VertexList = "Vertices";
+constexpr StringLiteral k_Edges = "Edges";
+constexpr StringLiteral k_Faces = "Faces";
+constexpr StringLiteral k_Polyhedra = "Polyhedrals";
```
imikejackson: Nit: "Polyhedrals" is not a word — the plural of polyhedron is "polyhedra". This string becomes the array name in the written file, so worth fixing before it fossilizes.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -156,6 +360,61 @@ DataStructure CreateTestDataStructure()
...
+  auto edges = std::make_shared<UInt64DataStore>(Constants::k_TupleShape, ShapeType{2}, 0);
+  auto* edgesArray = IGeometry::SharedEdgeList::Create(dataStructure, Constants::k_Edges, edges, dataGroup->getId());
+  FillDataStore<uint64>(*edges.get());
+
+  auto faces = std::make_shared<UInt64DataStore>(Constants::k_TupleShape, ShapeType{3}, 0);
+  auto* facesArray = IGeometry::SharedTriList::Create(dataStructure, Constants::k_Faces, faces, dataGroup->getId());
+  FillDataStore<uint64>(*faces.get());
+
+  auto polyhedra = std::make_shared<UInt64DataStore>(Constants::k_TupleShape, ShapeType{4}, 0);
+  auto* polyhedraArray = IGeometry::SharedTriList::Create(dataStructure, Constants::k_Polyhedra, polyhedra, dataGroup->getId());
```
imikejackson: Fixture-validity nits (invisible to the IO round-trip, but anyone reusing this fixture with a geometry-consuming filter inherits garbage):

* The polyhedra array is created via `IGeometry::SharedTriList::Create` — use `SharedHexList`/`SharedFaceList` aliases where they apply, for readability.
* This 4-component array is installed as the cell list of the **Hexahedral** geometry (line 415) — hexahedra need 8 nodes per cell. Similarly the shared 3-component face list is installed on the **Quad** geometry, which needs 4.
* The cell AttributeMatrix shape `{3,2,1}` matches the image dims XYZ instead of the reversed ZYX convention used by `CreateImageGeometryAction`.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -156,6 +360,61 @@ DataStructure CreateTestDataStructure()
...
+  auto* edgeGeom = EdgeGeom::Create(dataStructure, Constants::k_EdgeGeom);
+  edgeGeom->setVertices(*vertexArray);
+  edgeGeom->setEdgeList(*edgesArray);
+
+  // 2D Geometries
+  auto* quadGeom = Create2DGeom<QuadGeom>(dataStructure, Constants::k_QuadGeom, *vertexArray, *edgesArray, *facesArray);
+  auto* triangleGeom = Create2DGeom<TriangleGeom>(dataStructure, Constants::k_TriangleGeom, *vertexArray, *edgesArray, *facesArray);
+
+  // 3D Geometries
+  auto* hexGeom = Create3DGeom<HexahedralGeom>(dataStructure, Constants::k_HexGeom, *vertexArray, *edgesArray, *facesArray, *polyhedraArray);
```
imikejackson: Nit: `quadGeom`, `triangleGeom`, `hexGeom`, `tetrahedralGeom` (and `neighborList`/`stringArray` above) are assigned and never read. Drop the assignments, or use the pointers for immediate sanity REQUIREs.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -275,11 +534,93 @@ DREAM3D::FileData CreateFileData()
 
 } // End Namespace
 
+TEST_CASE("WriteDREAM3DFilter:Invalid Parameters")
```
imikejackson: Nit ×2, applies to all four new TEST_CASEs (lines 537, 571, 593, 605):

* They are the only untagged cases in this file — add `[WriteDREAM3DFilter]` so tag-based filtering doesn't skip them.
* None of them call `UnitTest::CheckArraysInheritTupleDims(dataStructure)` near the end, per project test guidelines.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -275,11 +534,93 @@ DREAM3D::FileData CreateFileData()
...
+  args.insertOrAssign(WriteDREAM3DFilter::k_ExportFilePath, std::make_any<FileSystemPathParameter::ValueType>(GetIODataPath()));
+  args.insertOrAssign(WriteDREAM3DFilter::k_WriteXdmf, std::make_any<bool>(false));
+  args.insertOrAssign(WriteDREAM3DFilter::k_UseCompression, std::make_any<bool>(false));
+  args.insertOrAssign(WriteDREAM3DFilter::k_CompressionLevel, std::make_any<int32>(1));
+
+  // Preflight the filter and check result
+  auto preflightResult = filter.preflight(dataStructure, args);
+  SIMPLNX_RESULT_REQUIRE_VALID(preflightResult.outputActions);
+
+  auto result = filter.execute(dataStructure, args);
```
imikejackson: This test executes the filter but never inspects the output file. The helpers already exist — a read-back plus `CheckTestDataStructure(...)` (or at minimum `REQUIRE(fs::exists(GetIODataPath()))`) would turn this from "filter returned valid" into an actual output assertion, nearly for free.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -275,11 +534,93 @@ DREAM3D::FileData CreateFileData()
...
+  SIMPLNX_RESULT_REQUIRE_VALID(result.result);
+}
+
+TEST_CASE("WriteDREAM3D:Pipeline / WriteXdmf combinations")
+{
+  UnitTest::LoadPlugins();
+  std::lock_guard<std::mutex> lock(m_DataMutex);
+
+  bool writeXdmf = GENERATE(true, false);
```
imikejackson, line 947: Nit: the `writeXdmf=true` iterations leave an `io_test.xdmf` sidecar in the data dir with no cleanup. Also, each of the 4 GENERATE combinations rebuilds the full `CreateTestDataStructure()` fixture — cheap today, but worth knowing this test's cost scales with that fixture.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,121 @@
...
+
+*Material PRs since baseline:* #1606 (added HDF5 compression parameters/behavior), #1544 (moved `executeImpl` logic into the `WriteDREAM3D` algorithm class, no behavior change), #1253 (out-of-core support).
+
+## Oracle
+
+*Class:* **1 (Analytical)**, primary. `Compression_LevelsRoundTrip` also carries a **Class 4 (Invariant)** companion check (file size must be non-increasing as gzip level rises) alongside its per-level content round-trip.
+
+*Applied:* Every test constructs its expected answer directly, without ever running the filter to "produce" the expected value:
+
+- **Content fidelity:** each test builds a `DataStructure` in C++ (`CreateTestDataStructure()`, or an inline array/geometry), writes it, reads it back, and asserts the read-back content equals what was built — by construction, not by comparison to a previously-captured file. `CheckTestDataStructure()` walks every `DataObject` kind the filter must support (nested `DataGroup`s, `AttributeMatrix`, `NeighborList`, `StringArray`, and all seven geometry types: Vertex/Edge/Triangle/Quad/Tetrahedral/Hexahedral/Image) and asserts exact values against the hand-known fill pattern from `FillDataStore()`.
```
imikejackson, line 52: **Blocking (overclaim).** "all seven geometry types" / "every `DataObject` kind" (and "every geometry and DataObject type" in the Summary, line 28) is false:

* simplnx has **8** geometry classes — **RectGridGeom** is missing from the fixture, and its *write* path is never exercised by any test in this file (it appears only in the read-side connectivity test consuming a canned `geoms.dream3d`).
* **ScalarData** and **GridMontage** both have HDF5 IO classes (`ScalarDataIO.hpp`, `GridMontageIO.cpp`) and are never round-tripped.
* **DynamicListArray** has no IO class at all — worth stating explicitly rather than omitting.

Either extend `CreateTestDataStructure()` (RectGrid is cheap — three bounds arrays) or scope this claim honestly and list the gaps.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,121 @@
...
+| 14 | Execute — write    | `DREAM3D::WriteFile(...)` returns valid → proceed to commit                                   | Every passing execute-path test |
+| 15 | Execute — commit   | `atomicFile.commit()` fails (rename onto final destination fails)                             | *Not directly tested.* Would require the destination path to become invalid between `AtomicFile::Create` and `commit()` (e.g., concurrent deletion of the parent directory) — a race not exercised by the suite. |
+| 16 | Execute — commit   | `atomicFile.commit()` succeeds                                                                | Every passing execute-path test (the output file is present and re-readable in every round-trip test) |
+| 17 | Execute — xdmf     | `write_xdmf_file=true` → rename temp `.xdmf` into place, succeeds                              | `"DREAM3DFileTest:DREAM3D File IO Test"` (writeXdmf=true), `CreateExportPipeline()`/`CreateMultiExportFiles()` (`write_xdmf_file=true`) |
```
imikejackson, line 84: **Blocking (false citation).** `"DREAM3DFileTest:DREAM3D File IO Test" (writeXdmf=true)` does not cover this path — that test's `writeXdmf` GENERATE variable is dead (see inline comment in DREAM3DFileTest.cpp:623); it writes via the `WriteFile(fileWriter, fileData)` overload, which never emits xdmf. Only the `CreateExportPipeline()`/`CreateMultiExportFiles()` citations are valid here. Same correction applies to the test-inventory row that says `writeXdmf ∈ {true,false}` (line 96).

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,121 @@
...
+
+*Evidence:* `parametersVersion()` is at 2 (compression parameters added after the filter's initial release); `git log --follow` on the algorithm/filter files shows the write path has been restructured multiple times since inception (out-of-core support #1253, atomic-file rework #900, algorithm-class extraction #1544) without ever tracking legacy DataContainerWriter's implementation.
+
+*Port-time deltas (SIMPL → SIMPLNX argument mapping, `WriteDREAM3DFilter::FromSIMPLJson`):*
+
+1. `OutputFile` → `export_file_path`, `WriteXdmfFile` → `write_xdmf_file`: direct 1:1 mapping, no behavior change.
+2. SIMPL's "Write Time Series" parameter has no SIMPLNX equivalent (dropped) — legacy time-series writing is not part of this filter's scope in NX; not a regression since no NX pipeline concept maps to it.
+3. `use_compression` is force-overridden to `false` for any pipeline converted from SIMPL JSON (`WriteDREAM3DFilter.cpp:130`), even though SIMPLNX's own default is `true`. This is deliberate: SIMPL v6 pipelines never wrote compressed files, so a converted pipeline preserves the exact on-disk encoding it shipped with rather than silently changing file size/behavior on re-run.
```
imikejackson, line 42: This override is real (`WriteDREAM3DFilter.cpp:130`), but nothing tests it — the SIMPL Backwards Compatibility test asserts only `export_file_path` and `write_xdmf_file`. One line closes it:

```cpp
CHECK(args.value<bool>(WriteDREAM3DFilter::k_UseCompression) == false);
```

If FromSIMPLJson paths are deliberately out of the 19-path enumeration, that's fine — but then a delta the report leans on this heavily should still be pinned by a test.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Algorithm Relationship
+
+*Classification:* **Rewrite** ~~| Port | Minor changes | New filter~~
+
+`WriteDREAM3DFilter` keeps the SIMPL UUID mapping (`3fcd4c43-9d75-5b86-aad4-4441bc914f37` → `WriteDREAM3DFilter`, `SimplnxCoreLegacyUUIDMapping.hpp:170`) and the legacy `DataContainerWriter` role, but the algorithm (`Algorithms/WriteDREAM3D.cpp`, 82 lines) was designed from the start for the current v8 `DataStructure` HDF5 layout, `AtomicFile`-based atomic writes, and (as of PR #1606) optional gzip compression — none of which exist in the legacy 6.5.171 writer. This has never been a line-by-line port of the legacy C++; the file format itself is a clean-sheet design (`k_CurrentFileVersion = "8.0"` vs. legacy's `"7.0"`/`DataContainers` group tag, see `Dream3dIO.cpp:31-42`).
```
imikejackson: Citation nit: `k_CurrentFileVersion = "8.0"` is defined in `Dream3dIO.hpp:31`, not `Dream3dIO.cpp:31-42` (the .cpp range holds the group-tag constants).

src/Plugins/SimplnxCore/vv/deviations/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,41 @@
...
+
+**Affected users:** Anyone attempting to open a SIMPLNX-written `.dream3d` file directly in a DREAM3D 6.5.171 install, or vice versa, outside of DREAM3D-NX's own dual-format `ReadDREAM3DFilter`. Users staying entirely within DREAM3D-NX (write with `WriteDREAM3DFilter`, read with `ReadDREAM3DFilter`) never observe this — round-trip fidelity within the new format is verified by the Class 1 tests in the V&V report.
+
+**Recommendation:** Trust SIMPLNX. The new format is a deliberate design improvement (atomicity, compression, JSON pipeline embedding) required for capabilities legacy DREAM3D never had (out-of-core datasets, versioned pipeline metadata). It is not "wrong" relative to 6.5.171; it is a different, and newer, on-disk contract. DREAM3D-NX remains able to *read* legacy 6.5.171 files via `ReadDREAM3DFilter`'s legacy import path, so migration is one-directional by design (import old, export new) rather than bidirectional.
```
imikejackson, line 25: Suggest adding one explicit sentence here: because the two writers target deliberately different on-disk contracts, this filter's V&V intentionally produced **no legacy comparison artifact** — a byte/dataset-level A/B against 6.5.171 `DataContainerWriter` output would be 100% noise by design, and the only tool that reads both formats is simplnx's own `ReadDREAM3DFilter`, which would make a writer comparison circular. Stating that here prevents a future auditor from reading the missing comparison folder as an omission.

imikejackson (changes requested):
Full adversarial V&V review. I independently verified the report's claims against the tree at head `7d24b7d`: all 17 Write-touching TEST_CASEs exist and pass locally (plus CI green on all 5 platforms), `ProbeHdf5Dataset` and the compression sub-suite are real, the error codes (-1/-2/-15), UUIDs, `parametersVersion()==2`, the `FromSIMPLJson` compression override, the 16 KiB doc threshold, and the cited PRs (#1606/#1544/#1253/#900) all check out. The 19-path enumeration is arithmetically consistent with the algorithm source, and the oracle is genuinely Class 1 and non-circular (all fixtures inline, no exemplar archive — so no provenance sidecar is correctly required).

Requesting changes for the items below. The blocking ones:

1. Dead `writeXdmf` GENERATE in the IO test, which also poisons the report's Path 17 citation.
2. The "every geometry and DataObject type" claim is false — RectGridGeom is never written by any test, and ScalarData/GridMontage are uncovered despite having IO classes.
3. Leftover dead code from an abandoned DynamicListArray round-trip (the type has no HDF5 IO at all — worth recording in the report instead of silently dropping).
4. NeighborList values are never compared after round-trip (only the tuple count).
5. The `use_compression=false` FromSIMPLJson override the report leans on is not asserted by the SIMPL BC test.

On the legacy A/B question: I agree no legacy comparison is meaningful here — the two writers target deliberately different on-disk contracts, and the only tool that can read both formats is simplnx's own reader, which would make a writer comparison circular. The Class 1 round-trip + HDF5 layout probing is the right oracle. See the inline suggestion on D1 to say so explicitly.

Remaining inline comments are nits (tags, `CheckArraysInheritTupleDims`, naming, fixture geometry validity, unused locals, cleanup).

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -165,6 +468,115 @@ DataStructure CreateTestDataStructure()
...
+  ShapeType xShape{Constants::k_TupleShape[0]};
+  ShapeType componentBounds{1};
+  auto xBounds = std::make_shared<Float32DataStore>(xShape, componentBounds, 0.0f);
+  auto* xBoundsArray = Float32Array::Create(dataStructure, Constants::k_XBounds, xBounds, rectGridGeom->getId());
+  FillDataStore<float32>(*xBounds.get());
+  ShapeType yShape{Constants::k_TupleShape[1]};
+  auto yBounds = std::make_shared<Float32DataStore>(yShape, componentBounds, 0.0f);
+  auto* yBoundsArray = Float32Array::Create(dataStructure, Constants::k_YBounds, yBounds, rectGridGeom->getId());
+  FillDataStore<float32>(*yBounds.get());
+  ShapeType zShape{Constants::k_TupleShape[1]};
```
imikejackson: Copy-paste bug: `zShape` is built from `Constants::k_TupleShape[1]` instead of `[2]`, and all three RectGrid bounds arrays are sized `dim` instead of `dim + 1`. For `k_TupleShape{3, 2, 1}` the bounds must be x=4, y=3, z=2 — N cells need N+1 bounds, a contract `RectGridGeom` itself relies on (`RectGridGeom.cpp` reads `xBnds->at(x + 1)`). As written, `getCoords`/`getSpacing` on the read-back geometry throws `out_of_range`, and in the `writeXdmf=true` iteration the emitted Topology `Dimensions="2 3 4"` disagrees with the three DataItem dimensions (3/2/2), so the .xdmf is self-inconsistent and ParaView rejects the RectGrid grid.

imikejackson (commented):
Re-review findings. Nine inline comments below; one additional issue lives outside this PR's diff so it goes here:

**Xdmf node-geometry writers transpose `geomName` and `hdf5FilePath`** — `src/simplnx/Utilities/Parsing/DREAM3D/Dream3dIO.cpp` lines 636, 648, 660: `WriteXdmfNodeGeometry1D/2D/3D` each forward to the next-lower writer with the last two arguments swapped (`WriteXdmfNodeGeometry0D(out, nodeGeom1D, hdf5FilePath, geomName)` against a callee signature of `(out, geom, geomName, hdf5FilePath)`). Both are `std::string_view`, so it compiles silently. Any `.xdmf` written for an Edge/Triangle/Quad/Tet/Hex geometry with a vertex attribute matrix emits Attribute DataItems that use the geometry name as the HDF5 file path, so ParaView/VisIt cannot resolve the node-centered arrays. This is pre-existing rather than introduced here, but the PR's V&V report claims Path 17 (xdmf) coverage, and the new `CheckXdmfFile()` oracle (existence-only) cannot catch it — the fix plus a content-level xdmf assertion belong with this V&V effort.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -165,6 +468,115 @@ DataStructure CreateTestDataStructure()
...
+  edgeGeom->setEdgeList(*edgesArray);
+
+  // 2D Geometries
+  auto* quadGeom = Create2DGeom<QuadGeom>(dataStructure, Constants::k_QuadGeom, *vertexArray, *edgesArray, *quadArray);
+  REQUIRE(quadGeom != nullptr);
+  auto* triangleGeom = Create2DGeom<TriangleGeom>(dataStructure, Constants::k_TriangleGeom, *vertexArray, *edgesArray, *trianglesArray);
+  REQUIRE(triangleGeom != nullptr);
+
+  // 3D Geometries
+  auto* hexGeom = Create3DGeom<HexahedralGeom>(dataStructure, Constants::k_HexGeom, *vertexArray, *edgesArray, *hexArray, *polyhedraArray);
```
imikejackson: The face and polyhedra lists are swapped in the `Create3DGeom` calls: the 8-component `hexArray` is installed as `HexahedralGeom`'s FACE list while the 4-component "Polyhedra" array becomes the hex connectivity list — a hex cell needs 8 vertices (`HexahedralGeom::k_NumVerts == 8`), not 4. Same for `TetrahedralGeom`: a 4-component array becomes the face list where triangle faces need 3. `CheckGeom3D` asserts the same swapped IDs, so the round-trip is self-consistently wrong: the flagship content-fidelity test validates malformed hex/tet geometries, and a writer bug corrupting genuine 8-component hex connectivity would pass unnoticed.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -449,18 +861,134 @@ GeometryTestCase MakeGeometryTestCase(std::string typeName, std::function<GeomTy
 
 } // End Namespace
 
+TEST_CASE("WriteDREAM3DFilter:Invalid Parameters", "[ReadDREAM3DFilter][WriteDREAM3DFilter]")
+{
+  UnitTest::LoadPlugins();
+  std::lock_guard<std::mutex> lock(m_DataMutex);
+
+  DataStructure dataStructure = CreateTestDataStructure();
```
imikejackson: Both SECTIONs here only call `filter.preflight()` and validate parameter guards (empty path, bad compression level) that never touch DataStructure contents, yet each SECTION re-entry builds the full `CreateTestDataStructure()` fixture (all 8 geometries, NeighborList, 10 scalars). The pre-existing `Compression_Preflight_RejectsOutOfRangeLevel` test does the same job with an empty `DataStructure ds;` — suggest the same here.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -449,18 +861,134 @@ GeometryTestCase MakeGeometryTestCase(std::string typeName, std::function<GeomTy
...
+    args.insertOrAssign(WriteDREAM3DFilter::k_UseCompression, std::make_any<bool>(true));
+    args.insertOrAssign(WriteDREAM3DFilter::k_CompressionLevel, std::make_any<int32>(0));
+
+    // Preflight the filter and check result
+    auto preflightResult = filter.preflight(dataStructure, args);
+    SIMPLNX_RESULT_REQUIRE_INVALID(preflightResult.outputActions);
+  }
+}
+
+TEST_CASE("WriteDREAM3DFilter:Valid Parameters", "[ReadDREAM3DFilter][WriteDREAM3DFilter]")
```
imikejackson, line 961: None of the four new TEST_CASEs (`Invalid Parameters`, `Valid Parameters`, `Pipeline / WriteXdmf combinations`, `Invalid File`) calls `UnitTest::CheckArraysInheritTupleDims(dataStructure);` near the end, per the project testing convention. `Valid Parameters` is the one that matters most — it reads a full DataStructure back, which is exactly where a tuple-dimension inheritance regression in the reader/writer would go undetected. Other tests in this file do call it, so the omission reads as an inconsistency rather than a file-wide waiver.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -449,18 +861,134 @@ GeometryTestCase MakeGeometryTestCase(std::string typeName, std::function<GeomTy
...
+  }
+}
+
+void CheckXdmfFile()
+{
+  auto filepath = GetXdmfPath();
+  REQUIRE(fs::exists(filepath));
+}
+
+TEST_CASE("WriteDREAM3D:Pipeline / WriteXdmf combinations", "[ReadDREAM3DFilter][WriteDREAM3DFilter]")
```
imikejackson, line 1029: This TEST_CASE, `Invalid File`, `Valid Parameters`, and the rewritten `DREAM3D File IO Test` all write/read the single shared `GetIODataPath()` (`newFile.dream3d`) guarded only by the process-local `m_DataMutex` — but `catch_discover_tests` runs each TEST_CASE as an independent ctest process, so the mutex protects nothing across them. Under `ctest -j N` one process can HDF5-truncate the file while another is mid-read, producing intermittent "truncated file"/HDF5 open errors. The PR triples the contenders on this path; distinct per-TEST_CASE filenames would remove the hazard.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -449,18 +861,134 @@ GeometryTestCase MakeGeometryTestCase(std::string typeName, std::function<GeomTy
...
+
+  bool writeXdmf = GENERATE(true, false);
+  Pipeline exportPipeline = GENERATE(CreateExportPipeline(), Pipeline());
+
+  auto writeResult = DREAM3D::WriteFile(fs::path(), CreateTestDataStructure(), exportPipeline, writeXdmf);
+  SIMPLNX_RESULT_REQUIRE_INVALID(writeResult);
+
+  if(writeXdmf)
+  {
+    CheckXdmfFile();
```
imikejackson: This asserts the .xdmf sidecar EXISTS after a write that is required to fail — a failed write to `fs::path()` must not produce a sidecar, so the assertion is inverted. It only passes because a stale `newFile.xdmf` from the earlier `Pipeline / WriteXdmf combinations` TEST_CASE is still on disk at `GetXdmfPath()`. Run in isolation (test filter, `--order rand`, or a fresh output dir) and `REQUIRE(fs::exists(...))` fails spuriously; run in order and it can never detect a regression where `WriteFile` starts emitting an .xdmf before validating the HDF5 path. Should be `REQUIRE_FALSE(fs::exists(...))` against a path derived from the (empty) target, or dropped.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
```diff
@@ -449,18 +861,134 @@ GeometryTestCase MakeGeometryTestCase(std::string typeName, std::function<GeomTy
...
   std::lock_guard<std::mutex> lock(m_DataMutex);
+
+  bool writeXdmf = GENERATE(true, false);
   // Write .dream3d file
   {
-    auto fileData = CreateFileData();
-    auto fileWriter = HDF5::FileIO::WriteFile(GetIODataPath());
-
-    auto writeResult = DREAM3D::WriteFile(fileWriter, fileData);
+    auto writeResult = DREAM3D::WriteFile(GetIODataPath(), CreateTestDataStructure(), CreateExportPipeline(), writeXdmf);
```
imikejackson, line 1146: Two knock-on effects of switching this to the path-based `DREAM3D::WriteFile` overload: (1) the exported `DREAM3D::WriteFile(HDF5::FileIO&, const FileData&)` forwarder (Dream3dIO.cpp:2426) lost its only caller in the tree, so a future regression in it (e.g. swapping `fileData.first`/`fileData.second`) passes CI silently while breaking downstream callers; (2) `CreateFileData()` at line 692 is now dead code in the anonymous namespace — clang/gcc emit `-Wunused-function`, which fails warnings-as-errors builds. Either keep one call through the FileData overload or delete both the helper and (if truly unwanted) deprecate the overload.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,127 @@
...
+
+*Material PRs since baseline:* #1606 (added HDF5 compression parameters/behavior), #1544 (moved `executeImpl` logic into the `WriteDREAM3D` algorithm class, no behavior change), #1253 (out-of-core support).
+
+## Oracle
+
+*Class:* **1 (Analytical)**, primary. `Compression_LevelsRoundTrip` also carries a **Class 4 (Invariant)** companion check (file size must be non-increasing as gzip level rises) alongside its per-level content round-trip.
+
+*Applied:* Every test constructs its expected answer directly, without ever running the filter to "produce" the expected value:
+
+- **Content fidelity:** each test builds a `DataStructure` in C++ (`CreateTestDataStructure()`, or an inline array/geometry), writes it, reads it back, and asserts the read-back content equals what was built — by construction, not by comparison to a previously-captured file. `CheckTestDataStructure()` walks every `DataObject` kind the filter must support (nested `DataGroup`s, `AttributeMatrix`, `NeighborList`, `StringArray`, and all seven geometry types: Vertex/Edge/Triangle/Quad/Tetrahedral/Hexahedral/Image) and asserts exact values against the hand-known fill pattern from `FillDataStore()`.
```
imikejackson: "all seven geometry types: Vertex/Edge/Triangle/Quad/Tetrahedral/Hexahedral/Image" — `CheckTestDataStructure` actually verifies eight: there is an explicit RectGrid block in the checker. RectGrid is missing from the formal record.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
```diff
@@ -0,0 +1,127 @@
...
+| 17 | Execute — xdmf     | `write_xdmf_file=true` → rename temp `.xdmf` into place, succeeds                              | `"DREAM3DFileTest:DREAM3D File IO Test"` (writeXdmf=true), `CreateExportPipeline()`/`CreateMultiExportFiles()` (`write_xdmf_file=true`) |
+| 18 | Execute — xdmf     | `write_xdmf_file=true`, rename fails → `MakeErrorResult` with system error message             | *Not directly tested.* Would require the `.xdmf` destination to become unwritable between the HDF5 write succeeding and the rename — not portably reproducible in the current suite. |
+| 19 | Execute — xdmf     | `write_xdmf_file=false` → skip rename, return `WriteFile`'s result directly                   | Most `Compression_*` tests, `"DREAM3DFileTest::StringArray"`, `"WriteDREAM3DFilter:Valid Parameters"` |
+
+## Test inventory
+
+| Test case | Status | Notes |
+|-----------|--------|-------|
+| `WriteDREAM3DFilter:Invalid Parameters` (§ Empty FilePath, § Bad Compression Level) | kept | Preflight-only, inline `DataStructure`. Covers Paths 1, 2. |
```
imikejackson: The inventory claims to be the complete list ("17 TEST_CASEs ... See Test inventory below for the full list") but omits `DREAM3DFileTest: Geometry Nested In DataGroup Round Trip` (tagged `[WriteDREAM3DFilter]`, executes the filter directly across all 8 geometry types, added for issue #1642). An auditor using this table as the authoritative coverage record gets a wrong count and a missing test-to-code-path mapping.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d. `zShape` now uses `k_TupleShape[2]`, and all three bounds arrays are sized `dim + 1` (x=4, y=3, z=2). `CheckTestDataStructure` asserts the `+1` sizes on read-back so the N+1 contract is pinned rather than assumed.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d. `HexahedralGeom` now takes `quadArray` as its face list and the 8-component `hexArray` as its polyhedra list; `TetrahedralGeom` takes `trianglesArray` as faces and `tetraArray` as polyhedra. `CheckGeom3D` was updated to assert the corrected IDs.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d. `Invalid Parameters` now uses a bare `DataStructure dataStructure;` with a comment noting both SECTIONs only exercise preflight parameter guards, matching `Compression_Preflight_RejectsOutOfRangeLevel`.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d. `UnitTest::CheckArraysInheritTupleDims(...)` was added to all the new TEST_CASEs, including `Valid Parameters` where it runs against the read-back DataStructure.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d. Each file-writing TEST_CASE now has its own output name (`write_dream3d_valid_params.dream3d`, `write_dream3d_pipeline_combos.dream3d`, `write_dream3d_file_data.dream3d`), with a comment explaining why. Only `DREAM3D File IO Test` still writes `newFile.dream3d`. Verified serially: 23/23 Read/Write DREAM3D tests green.

Correcting my original comment on two counts. First, the severity: I justified this by a `ctest -j N` scenario, but ctest is not run with `-j` in this project — tests share on-disk fixtures and `TestFileSentinel` archives that do not survive parallel invocation, and the CI test presets have no `execution` block, so CI is serial as well. The shared-filename hazard was therefore latent rather than active. The distinct-filename change is still worth keeping as hygiene, but it was not fixing a live failure. Second, the evidence: an earlier version of this reply cited three consecutive `ctest -j 12` runs, which is not a valid way to verify anything in this repo. Re-verified with serial runs instead.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d. The inverted assertion is gone, replaced by a comment noting a write that fails before producing a file has no target sidecar to check.

src/Plugins/SimplnxCore/test/DREAM3DFileTest.cpp
imikejackson: Fixed in d75ba7d by keeping a caller rather than deleting the helper: the new `WriteDREAM3D:FileData Overload` TEST_CASE writes through `DREAM3D::WriteFile(HDF5::FileIO&, const FileData&)` via `CreateFileData()`, reads the file back, and asserts both the pipeline size and full `CheckTestDataStructure` content. That pins the overload argument ordering, so a swap of `fileData.first`/`.second` now fails CI instead of passing silently.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
imikejackson: Fixed in d75ba7d. The report now says eight geometry types and names RectGrid explicitly.

src/Plugins/SimplnxCore/vv/WriteDREAM3DFilter.md
imikejackson: Fixed in d75ba7d. `DREAM3DFileTest: Geometry Nested In DataGroup Round Trip` was added to the inventory with its code-path mapping, and the count corrected. As of 54adc40 the count is 20 TEST_CASEs, with the new `WriteDREAM3DFilter:Unwritable DataObject Type` case also listed.

imikejackson (approved):

PR #708, opened 2023-10-03
DOC: Update documentation files to allow Sphinx to generate html documentation

imikejackson:

PR #673, opened 2023-08-22
ENH: Filter help text filled out. Docs updated with python section

imikejackson:
* Each parameter inside of each filter had it's 'help text' filled out. There are a few exceptions on a few ITK filters that have no description in the ITK code base itself.
* Each filter doc also had a section added that describes the named arguments in the python wrappers.

PR #1693, opened 2026-07-28
VV: ITKImageWriter fully V&V'ed

JDuffeyBQ:
## Naming Conventions

## Filter Checklist

## Unit Testing

## Code Cleanup

### DIFF
--- a/src/Plugins/ITKImageProcessing/docs/ITKImageWriterFilter.md
+++ b/src/Plugins/ITKImageProcessing/docs/ITKImageWriterFilter.md
@@ -6,7 +6,7 @@ ITKImageProcessing (ITKImageProcessing)
 
 ## Description
 
-This **Filter** will save images based on an array that represents grayscale, RGB or ARGB color values. If the input array represents a 3D volume, the **Filter** will output a series of slices along one of the orthogonal axes.  The options are to produce XY slices along the Z axis, XZ slices along the Y axis or YZ slices along the X axis.
+This **Filter** will save images based on scalar, vector, RGB, or RGBA values. Supported component counts are 1, 2, 3, 4, 10, 11, and 36. If the input array represents a 3D volume, the **Filter** will output a series of slices along one of the orthogonal axes. The options are to produce XY slices along the Z axis, XZ slices along the Y axis or YZ slices along the X axis.
 
 The available output formats are determined by the installed ITK image I/O backends and the filename extension. TIFF, BMP, and PNG are common 2D choices; not every pixel type is supported by every format.
 
--- a/src/Plugins/ITKImageProcessing/docs/ITKImageWriterFilter.md
+++ b/src/Plugins/ITKImageProcessing/docs/ITKImageWriterFilter.md
@@ -6,7 +6,11 @@ ITKImageProcessing (ITKImageProcessing)
 
 ## Description
 
-This **Filter** will save images based on an array that represents grayscale, RGB or ARGB color values. If the input array represents a 3D volume, the **Filter** will output a series of slices along one of the orthogonal axes.  The options are to produce XY slices along the Z axis, XZ slices along the Y axis or YZ slices along the X axis. The user has the option to save in one of 3 standard image formats: TIF, BMP, or PNG. The output files will be numbered sequentially starting at zero (0) and ending at the total dimensions for the chosen axis. For example, if the Z axis has 117 dimensions, 117 XY image files will be produced and numbered 0 to 116. Unless the data is a single slice then only a single image will be produced using the name given in the Output File parameter.
+This **Filter** will save images based on an array that represents grayscale, RGB or ARGB color values. If the input array represents a 3D volume, the **Filter** will output a series of slices along one of the orthogonal axes.  The options are to produce XY slices along the Z axis, XZ slices along the Y axis or YZ slices along the X axis.
+
+The available output formats are determined by the installed ITK image I/O backends and the filename extension. TIFF, BMP, and PNG are common 2D choices; not every pixel type is supported by every format.
+
+For a series, the *Output File* stem is followed by an underscore and the slice index. *Index Offset*, *Total Number of Index Digits*, and *Fill Character* control the first index and its formatting. For example, with 117 Z cells, offset *0*, width *3*, and fill character `0`, the XY output files are named `slice_000.tif` through `slice_116.tif`. A single-slice output keeps the exact name supplied in *Output File*.
 
 An example of a **Filter** that produces color data that can be used as input to this **Filter**
 is the {ref}`Generate IFP Colors <OrientationAnalysis/ComputeIPFColorsFilter:Description>` **Filter**, which will generate RGB values for each voxel in the volume.
--- a/src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/ITKImageWriterFilter.cpp
+++ b/src/Plugins/ITKImageProcessing/src/ITKImageProcessing/Filters/ITKImageWriterFilter.cpp
@@ -39,13 +39,6 @@ namespace cxITKImageWriterFilter
 using ArrayOptionsType = ITK::ScalarVectorPixelIdTypeList;
 using RgbRgbaArrayOptionsType = ITK::ArrayOptions<ITK::ArrayComponentOptions<true, false, true>, ITK::ArrayUseAllTypes>;
 
-constexpr std::array<usize, 7> k_AllowedComponentSizes = {1, 2, 3, 4, 10, 11, 36};
-
-bool IsValidComponentSize(usize componentSize)
-{
-  return std::find(k_AllowedComponentSizes.begin(), k_AllowedComponentSizes.end(), componentSize) != k_AllowedComponentSizes.end();
-}
-
 // Rejects 0 - 31 ASCII control characters
 bool IsValidFillCharacter(char fillCharacter)
 {
@@ -269,7 +262,7 @@ Parameters ITKImageWriterFilter::parameters() const
   params.insert(std::make_unique<GeometrySelectionParameter>(k_ImageGeomPath_Key, "Image Geometry", "Select the Image Geometry Group from the DataStructure.", DataPath{},
                                                              GeometrySelectionParameter::AllowedTypes{IGeometry::Type::Image}));
   params.insert(std::make_unique<ArraySelectionParameter>(k_ImageArrayPath_Key, "Input Image Data Array", "The image data that will be processed by this filter.", DataPath{},
-                                                          nx::core::ITK::GetScalarPixelAllowedTypes()));
+                                                          nx::core::ITK::GetScalarPixelAllowedTypes(), ArraySelectionParameter::AllowedComponentShapes{{1}, {2}, {3}, {4}, {10}, {11}, {36}}));
 
   return params;
 }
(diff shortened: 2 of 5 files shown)


imikejackson:
# V&V Review: Write Image (ITK)

Reviewed as a second-engineer V&V review: every claim in `vv/ITKImageWriterFilter.md` and `vv/deviations/ITKImageWriterFilter.md` was re-derived, the A/B claim was re-run with an independent methodology, and adversarial / CPU / memory / out-of-core passes were performed. Findings are grouped by severity.

## 1. Claims that verify — no action needed

- [x] **uint32/uint64 dispatch fix is real and is covered.** Independent A/B: reverted `CopyTupleTyped<uint32>`→`<int32>` and `<uint64>`→`<int64>`, rebuilt, ran `ctest -R "Analytical Pixel Order"` → tests **957 (uint32)** and **959 (uint64)** fail; the other 8 pass. Restored the fix → 10/10 pass. The pre-fix failure mode is an uncaught `std::bad_cast` escaping `filter.execute()` (Catch2 reports `{Unknown expression after the reported line}` at `ITKImageWriterTest.cpp:159`), not a wrong-value result.
- [x] **Oracle values are correct.** All seven expected slice matrices re-derived from `value(x,y,z)=x+10y+100z` on the 3×2×2 fixture (XY 3×2 ×2 slices, XZ 3×2 ×2 slices, YZ 2×2 ×3 slices) match the test source exactly. Max value 112 fits every type including `int8`, so no silent saturation.
- [x] **Port fidelity of the slicing arithmetic.** The XY/XZ/YZ index expressions and `indexNew` in `ITKImageWriterFilter.cpp:419-490` are byte-for-byte equivalent to 6.5.171 `ITKImageWriter.cpp:396-449` + `copyTuple()`.
- [x] **Deviation D1 verified against legacy source.** 6.5.171 `ITKImageWriter::saveImageData()` does `out << "_" << slice;` with no padding, guarded by the same `if(maxSlice != 1)`. The stated remediation (Total Number of Index Digits = 1) reproduces the legacy filename shape.
- [x] **15/15 ctest entries pass on PR head** (`ctest -R ITKImageWriter`, 1.40 s).
- [x] **No circular oracle, no new exemplar archive** — inline expected values only, consistent with policy.

## 2. Blocking — undocumented behavioral deviations from 6.5.171

- [ ] **B1. RGBA (4-component) input is a functional regression from 6.5.171, and the docs still advertise ARGB.** Legacy compiled with `DREAM3D_USE_RGB_RGBA 1` / `DREAM3D_USE_Vector 0`, i.e. `itk::RGBPixel`/`itk::RGBAPixel`. NX uses `ITK::ScalarVectorPixelIdTypeList` → scalar + `itk::Vector`, which has no `nComp == 4` arm. Measured on a 4×3×2 uint8 geometry:
  | nComp | preflight | execute |
  |---|---|---|
  | 1 | valid | valid |
  | 3 | valid | valid (written as `itk::Vector<T,3>`, **not** `itk::RGBPixel<T>`) |
  | 4 | **valid** | **fails, `-21010` "Vector dimension not supported. cDims[0] = 4"** |

  `docs/ITKImageWriterFilter.md:9` still says "grayscale, RGB or **ARGB** color values" and lines 15-16 direct users to Compute IPF Colors as valid input. Needs: a `-D2` deviation entry, a component-count check in `preflightImpl` so this fails at preflight rather than mid-execute, and a docs correction.
- [ ] **B2. XZ and YZ output files carry the wrong pixel spacing and origin.** `ITK::ImageGeomData(imageGeom)` captures `(sx,sy,sz)`/`(ox,oy,oz)`; `executeImpl` overwrites only `.dims`, so `WrapDataStoreInImage` always emits axis0 = `sx`/`ox` and axis1 = `sy`/`oy` regardless of plane. Measured with spacing `(1,2,4)` and origin `(10,20,40)`, reading the `.mha` headers:
  | Plane | `DimSize` | `ElementSpacing` | should be | `Offset` | should be |
  |---|---|---|---|---|---|
  | XY | `3 2` | `1 2` | `1 2` ✅ | `10 20` | `10 20` ✅ |
  | XZ | `3 2` | `1 2` | **`1 4`** | `10 20` | **`10 40`** |
  | YZ | `2 2` | `1 2` | **`2 4`** | `10 20` | **`20 40`** |

  This is also a deviation from 6.5.171, which built a fresh `ImageGeom` and set only dimensions (so it wrote spacing `1 1` / origin `0 0`). Needs a fix plus a deviation entry. The oracle asserts pixels only, so it cannot catch this.

## 3. Correctness / robustness the V&V did not reach

- [ ] **C1. Out-of-core input array → uncaught `std::bad_cast`.** This filter hand-rolls its execute and never calls `ITK::DataCheck`/`ITK::Execute`, so it never reaches the `k_OutOfCoreDataNotSupported (-2002)` guard added by #1555 (`ITKArrayHelper.hpp:843,865`). With an OOC array, `currentData.createNewInstance()` returns an `H5Store<T>` (`FileStore.hpp:138` → `AbstractDataStore<T>`), and `WriteImage()` does `dynamic_cast<DataStore<T>&>(dataStore)` → throws. Identical failure mode to the uint32 bug reproduced above. The report has no out-of-core section; #1555 is cited in the PR list but its guard was never verified to apply here.
- [ ] **C2. A one-character Fill Character can throw out of `preflightImpl`.** `CreateIndexString` builds `fmt::format("{{:{}>{}}}", fillCharacter, totalDigits)`. Reproduced:
  - `fillChar = "{"` → `fmt::format_error: invalid fill character '{'` escapes preflight (in the GUI this is a keystroke away).
  - `fillChar = "/"` → accepted; preview becomes `s_//3.tif`, and execute writes to a mangled path.

  The `size() != 1` fix validates length only. Validate that the character is a legal fmt fill **and** a legal filename character. Note `executeImpl` uses `std::setfill(fillChar[0])` (iostreams) while preflight uses fmt — two different formatting paths for the same parameter.
- [ ] **C3. The `Example Output File` preflight preview is wrong.** `CreateIndexString(maxSlice, ...)` formats the slice **count** as the index and ignores `indexOffset`. Reproduced on a 3-slice XY export: preview `s_003.tif`, actual files `s_000.tif`, `s_001.tif`, `s_002.tif` — index 003 is never written. For a single-slice output it previews `name_001.ext` while the file is `name.ext`, contradicting the sentence this PR added to the docs. Should be `CreateIndexString(indexOffset, ...)` with the suffix suppressed when `maxSlice == 1`. Code path #1 ("Valid preflight") is claimed covered, but no test asserts this preflight value. (Same defect exists in `SimplnxCore/WriteImageFilter.cpp:290` — worth a follow-up issue.)
- [ ] **C4. `copyFrom()`'s `Result<>` is discarded.** `CopyTupleTyped` (`ITKImageWriterFilter.cpp:178`) drops the return value, so a `-14600`/`-14601` would silently produce wrong output instead of an error.
- [ ] **C5. Dead code containing a live off-by-one.** `executeImpl` always sets `newImageGeom.dims` Z = 1, so `ArraySwitchFuncDimsImpl` always selects `Dimensions == 2`, making `if(Is2DFormat(filePath) && Dimensions == 3)` unreachable. `Is2DFormat()`, `WriteAs2DStack()` and error `-21012` are therefore all dead. `WriteAs2DStack` also contains `for(uint64 index = indexOffset; index < (z_size - 1); index++)` (writes one file too few, and starts numbering at `indexOffset`) plus a hardcoded `{:03d}` that ignores `totalDigits`/`fillChar`. Delete it, or state the unreachability explicitly in the report.
- [ ] **C6. No cancel check and no progress messages.** `shouldCancel` and `messageHandler` are unused in `executeImpl`. `IFilter::execute` only samples `shouldCancel` *after* `executeImpl` returns (`IFilter.cpp:225`), so a large multi-slice export is uninterruptible and silent. VV_Notes exempts *testing* the cancel path, not omitting the check.

## 4. CPU and memory review

- [ ] **P1. Per-tuple type dispatch dominates runtime (~6.7× avoidable).** `CopyTuple` performs a `switch` on `DataType` + two `dynamic_cast`s + a virtual `copyFrom` **per tuple**. Micro-benchmark, 4,444,713 tuples (Small IN100 189×201×117), float32, this machine:
  ```
  per-tuple dispatch = 208 ms    hoisted dispatch = 31 ms
  ```
  Whole-filter times on the same volume: XY **268 ms**, XZ **286 ms**, YZ **339 ms** — so the avoidable dispatch is roughly 60–75% of total runtime. Hoist the cast out of the slice loops. Better: XY slices are fully contiguous in the source store (one bulk `copyFrom(0, src, slice*dA*dB, dA*dB)` per slice replaces `dA*dB` calls) and XZ rows are contiguous per Z.
- [ ] **P2. Full-volume temp buffer where legacy allocated one slice.** `currentData.createNewInstance()` (`DataStore.hpp:521`) allocates `getTupleShape()` — the entire volume — zero-filled, then only one slice is ever populated. Legacy's `ITKIW_PREP_DC` allocated `numElements = dA * dB`. Small IN100 float32: **17.8 MB** temp vs legacy's **152 KB** (~117×), doubling peak footprint for the array. A 1000³ uint16 volume goes from ~2 GB + 2 MB to ~4 GB, plus a full-volume `memset` on every run. Allocate the slice shape instead. Neither the report nor the PR mentions this port-time regression.
- [ ] **P3. `nComp` is dead.** Passed to `CopyTupleTyped` and used only inside the `#if 0` block. Remove both the parameter and the `#if 0` block (`ITKImageWriterFilter.cpp:172-176`) per the PR's own "no commented out code" checklist.

## 5. Report and documentation accuracy

- [ ] **R1. Test name mismatch.** The report references `Analytical TIFF Pixel Order` in the Oracle "Encoded" line, in all six code-path rows, and in the test inventory. The actual test is `Analytical Pixel Order`. "TIFF" also misleads, since 9 of the 10 types write MetaImage.
- [ ] **R2. "17 of 18 branches exercised" is understated.** Not enumerated: component-count dispatch (`nComp` 1/2/3/10/11/36) and `-21010`; `WriteAs2DStack`/`-21012` (dead, see C5); `create_directories` and `-19000`; `AtomicFile::Create` failure; `CopyTuple`'s `default: throw std::runtime_error`; the preflight plane `default:`; and the OOC path (C1). Please recount, or state the denominator's scope explicitly.
- [ ] **R3. Code path 17 and the test name misstate the mechanism.** `3D Image Single-File Output` produces one file because `maxSlice == 1` in `SaveImageData` (`dims.getY() == 1` for the XZ plane) — not because the format is "non-2D". `Is2DFormat` is never consulted (C5). Suggest renaming to e.g. `Single Slice Keeps Exact Output Name` and rewording the row.
- [ ] **R4. "Minor changes … the pixel comparisons show no output change from these substitutions" is too strong.** Pixels match, but the ITK-API substitution changed supported pixel types (B1) and the file metadata written (B2). Please reword the Algorithm Relationship section.
- [ ] **R5. `vv/provenance/ITKImageWriterFilter.md` is missing.** Every other V&V'd filter in the tree has one (`SimplnxCore/vv/provenance/`, `OrientationAnalysis/vv/provenance/`). The report's legacy-comparison claim — "Small IN100 has 117 XY, 201 XZ, and 189 YZ float32 TIFF slices", "each slice matches exactly" — ships with no pipelines, no comparison script, no hashes, no stated tolerance, and no OneDrive pointer, so it is not independently verifiable. Please add the provenance deliverable.
- [ ] **R6. No Algorithm review section.** `OrientationAnalysis/vv/ReadAngDataFilter.md` has one; this report has no algorithm review, CPU, memory, or out-of-core discussion — which is where P1, P2, and C1 would have surfaced.
- [ ] **R7. PR history omissions.** `git log --follow` on the `.cpp`/`.hpp` also lists **#1088 "Added versioning to filter parameters and json"** (omitted, and relevant since the report says nothing about `parametersVersion()`), plus docs-only PRs #1571, #956, #975, #914, #847, #708, #673.
- [ ] **R8. `UnitTest::CheckArraysInheritTupleDims(dataStructure)` missing** from `Fill Character Validation`, `Dimension Mismatch Validation`, and `3D Image Single-File Output` (present only in the templated test).
- [ ] **R9. Strengthen the oracle.** Add expected file **count** per plane (2 XY / 2 XZ / 3 YZ, and no extra files — `Write Stack` already does this via `validateOutputFiles`) and, once B2 is fixed, a spacing/origin round-trip assertion.
- [ ] **R10. Report header placeholders.** `Verified commit` and `Sign-off` are still `<...>`.
- [ ] **R11. State the `parametersVersion()` decision.** The fill-character rule changed from "≤ 1 character accepted" to "exactly 1", so a saved pipeline with an empty Fill Character now errors on load. The parameter *set* is unchanged, so leaving it at 2 is defensible, but per policy the decision should be recorded in the report with a one-line rationale.

## Reproduction notes

All measurements above were taken from `NX-Com-Qt69-Vtk96-Rel` at PR head `4488f29`, using temporary probe test cases that were removed afterward (working tree verified clean, 15/15 ITKImageWriter tests passing).

nyoungbq (approved):

imikejackson:
@JDuffeyBQ Are each of the checklist items addressed?

JDuffeyBQ:
> @JDuffeyBQ Are each of the checklist items addressed?

I believe so.

PR #1697, opened 2026-08-06
ENH: Overhaul filter messaging API and remove the spdlog dependency

imikejackson:
## Summary

Overhauls the filter messaging API. Three problems motivated this: `MessageHandler` sent everything
through `operator()` overloads that gave no hint what kind of message was being sent, spdlog was used
nowhere else in the codebase yet brought an async logger and a second throttling layer, and 27 files
gated progress reporting on a `steady_clock::now()` call in the loop body.

* `MessageHandler` gains named send methods and loses all four `operator()` overloads.
  `IFilter::ProgressMessage` folds into `Message` as an `int32 progress` field, which removes the
  `static_cast` downcast in `PipelineFilter`.
* Progress is reported through two intent-bearing calls rather than one generic one, so a filter
  records what it meant to display. `sendProgressCount()` renders `"Processing tuples: 35/100"` for
  loops whose counts mean something to a user; `sendProgressPercent()` renders
  `"Analyzing voxels: 33.33%"` where the counts are too large to read. Overloading on numeric type was
  rejected because an integer literal would silently pick a form.
* New `ThrottledMessageHandler` replaces `MessageHelper`, `Messenger`, `ThrottledMessenger`,
  `ProgressMessenger`, `ProgressMessageHelper` and `ProgressMessageData`. A per-instance thread opens
  an `atomic_bool` gate once per interval, so the loop body only reads a flag. Measured on Apple
  Silicon, clang `-O2`, 200M iterations: **14.7 ns/iter for `steady_clock::now()` versus 0.25 ns/iter
  for the flag**, against 1.0 ns/iter for a nontrivial loop body. Messages are formatted only when one
  is due, so a throttled loop does not allocate.
* **spdlog is removed** from `CMakeLists.txt` and `vcpkg.json`.
* 227 call sites across 99 files migrated. Deleting `operator()` turned every missed site into a
  compile error, which is how the exceptions were found rather than shipped.

### Why the API is shaped this way

**Named methods instead of `operator()`.** `m_MessageHandler(x)` gave no indication at the call site
whether `x` was status, a warning, or an error. Four overloads meant the *type* of the argument silently
selected the behaviour. `sendInfoMessage` / `sendWarningMessage` / `sendErrorMessage` say it outright.

**Two progress calls instead of one.** There are two distinct things a filter wants a user to read.
Sometimes the counts are meaningful — `"35/100 tuples complete"`. Sometimes the loop bound is so large
that counts are noise and a percentage with a decimal place or two is the only readable form. That is a
judgement the filter author makes, and encoding it in the *method name* records it for whoever maintains
the filter next:

```cpp
throttle.updateCount("Processing tuples", i, total);    // "Processing tuples: 35/100"
throttle.updatePercent("Analyzing voxels", i, total);   // "Analyzing voxels: 33.33%"
```

We deliberately rejected overloading on numeric type (`int32` vs `float32`) for this. It reads as the
obvious C++ answer, but `sendProgressMessage(label, 35)` versus `sendProgressMessage(label, 35.0f)`
would let an integer literal silently choose the display form — hiding the very intent the API exists
to capture. Distinct names cannot be selected by accident.

**The sender renders the text.** `"35/100"` contains no percentage, so a `(label, percent)` pair cannot
express it. `Message` therefore carries the fully rendered text plus an `int32 progress` used only to
drive a progress bar. That pair is exactly what `PipelineFilter` and DREAM3D-NX's
`onFilterProgressMessage(index, progress, message)` already consume, so nothing downstream changed.

**Same vocabulary at both levels.** `MessageHandler` has `sendProgressCount` / `sendProgressPercent`
for one-off reports; `ThrottledMessageHandler` mirrors them as `updateCount` / `updatePercent` for
loops, with `incrementCount` / `incrementPercent` when you only know how many items you just finished.
Non-progress status text goes through `queueMessage()`, which keeps status and progress visibly
separate.

### How the throttle works

The problem: `steady_clock::now()` costs ~15x a nontrivial loop body, so gating progress on the clock
on every iteration dominates cheap loops. The fix is to move all timekeeping off the hot path.

Each `ThrottledMessageHandler` owns one thread that waits on a condition variable for the interval,
sets a flag, and waits again:

```cpp
while(!m_ConditionVariable.wait_for(lock, m_Interval, [this] { return m_Stop; }))
{
  m_Ready.store(true, std::memory_order_relaxed);
}
```

The loop body only reads that flag. Everything past the gate — the percent maths, the `fmt::format`,
the handler call — happens at most once per interval:

```cpp
bool ThrottledMessageHandler::isReady()
{
  if(!m_Ready.load(std::memory_order_relaxed))   // ~0.25 ns, read-only
  {
    return false;
  }
  return m_Ready.exchange(false, std::memory_order_relaxed);
}
```

Four details worth knowing while reviewing:

* **Load-then-exchange, not a bare `exchange`.** An unconditional read-modify-write on a shared cache
  line measures 0.62 ns/iter against 0.25 ns for the guarded form. The hot path stays read-only until a
  send actually happens.
* **`exchange(false)` gives single-winner election for free.** When N threads race the gate, exactly one
  observes `true`, so a parallel algorithm emits one message per interval rather than N.
* **The flag starts `true`**, so the first message fires immediately and then throttles. This reproduces
  the old spdlog `ThrottleSink` behaviour and avoids a dead first second on long filters.
* **One flag per instance, not a global.** Because observing the flag consumes it, a single shared flag
  would let the first throttle to look starve every other one.

Thread lifetime is per instance: created in the constructor, signalled and joined in the destructor so
shutdown is immediate rather than waiting out the interval. Measured create + signal + join is 15.3 µs,
so a 47-filter pipeline spends about 0.72 ms on this in total. A shared refcounted ticker was considered
and rejected — it would have needed a singleton, a static `weak_ptr`, a registry, its own mutex,
per-entry interval arithmetic and static-destruction-order reasoning, all to amortise well under a
millisecond, given only 1–3 instances exist concurrently in practice.

The class owns a thread capturing `this`, so it is neither copyable nor movable, and it is deliberately
**not** internally thread-safe — see the parallel section below for how that is handled.

### Bugs fixed along the way

* `PipelineFilter` discarded warning and error message *text*, sending only the fault state, so the
  reason for a fault never reached observers.
* `WriteSPParksSites` computed a percent and tagged it `Type::Info`, so the value never reached the
  progress bar.
* `ReadDeformKeyFileV12` reported seven parse faults as informational progress text.
* `ConvertOrientations` and `ComputeAvgOrientations` called `steady_clock::now()` while holding the
  progress mutex, on every worker call.
* Removed `PipelineFilterMessage` (never constructed), three `PipelineObserver` connections to
  Pipeline-level signals that have no emitter, and three `updateProgress()` seams with zero callers.

### Parallel algorithms

`ThrottledMessageHandler` is not internally thread-safe and owns a thread, so it is never created per
worker. Each parallel algorithm owns one throttle behind a mutex and exposes a thread-safe seam;
workers call the seam and hold nothing that can reach an `IFilter::MessageHandler`. 25 parallel
algorithms were converted this way. `ErodeDilateBadData` additionally pre-gates on a count so the
shared mutex is taken about a hundred times per array rather than once per voxel.

### Behaviour changes visible to users

* Progress message text standardises. Bespoke forms such as `" - Calculating || 45.00% Complete"`
  become `" - Calculating: 45.00%"`. No test asserts on message text.
* Roughly 20 loops now drive the DREAM3D-NX progress bar and the `nxrunner` percent column for the
  first time. Previously only two call sites in the entire repository fed that channel, because
  `MessageHelper` structurally could not.
* `AlignSections` and `SampleSurfaceMesh` report aggregate progress instead of naming a per-worker
  array or feature. In both cases the counter was already shared, so only the per-worker prefix is
  gone, and `AlignSections` already announces each array separately before dispatch.

### Documentation

`docs/PortingFilters.md` Section 6 taught the deleted classes with compilable examples, so anyone
following it would have written code that no longer builds. Rewritten around the current API, including
the thread-safe seam pattern that replaces the out-of-date `steady_clock`-under-mutex example. Six V&V
reports that named the deleted classes when describing progress feedback as a deviation from legacy
DREAM3D 6.5.171 now state simply that SIMPLNX adds throttled progress feedback; the deviations
themselves are unchanged and remain classified as UX-only.

## Test Plan

- [x] 30 new tests covering the send methods, both progress display forms and their decimal handling,
      percent clamping, zero denominators, the throttle gate, cross-thread sender election, counter
      accumulation, prompt destruction, and message routing
- [x] Tests pass on in-core build (`NX-Com-Qt69-Vtk96-Rel`): 1765/1766
- [ ] Tests pass on out-of-core build — **not yet run**
- [x] `nxrunner` output verified unchanged by running pipelines before and after the observer cleanup

The single in-core failure is `SIMPL Map Inventory: Validate Coverage`, which is pre-existing and
unrelated: it reports `EMsoftSO3SamplerFilter (SimplnxReview)` missing a SIMPL backwards-compatibility
fixture. It fails identically with and without this branch.

**Companion change required:** `SimplnxReview` has 14 files that need the same call-site migration, or
that plugin will not compile against this branch.


imikejackson:
@jduffeybq — good question, and worth writing up properly because the answer is "no, that specific thing can't happen, but you're right that the class isn't safe for direct worker use, for three reasons that aren't the one you named."

## The double-send can't happen

`isReady()` in `ThrottledMessageHandler.cpp`:

```cpp
if(!m_Ready.load(std::memory_order_relaxed))
{
  return false;
}
return m_Ready.exchange(false, std::memory_order_relaxed);   // was relaxed, see below
```

The **load is not the gate** — it's a pre-filter that keeps the hot path read-only, which is where the perf claim comes from. The gate is the **exchange**, an atomic read-modify-write. If N threads all pass the load, all N reach the exchange, and the atomic's modification order guarantees exactly one of them reads back `true`. The others get `false` and return. So at most one message per gate-opening no matter how many threads race it. `ThrottledMessageHandlerTest.cpp` pins this as "Elects a single sender across threads".

## What the gate does *not* cover

**1. The accumulator is a plain `usize`, updated before the gate.**

```cpp
void ThrottledMessageHandler::incrementCount(usize delta)
{
  m_CurrentProgress += delta;   // non-atomic RMW, every call, every thread
  if(!isReady()) { return; }
```

Data race, UB, lost counts. It runs on every iteration of every thread, not just the winners, and the gate does nothing for it. This is the actual hazard if a worker calls the throttle directly.

**2. `reset()` writes `m_Label` and `m_MaxProgress`** while other threads read them in the update methods. Per-phase resets overlapping live workers is a race on a `std::string`.

**3. Relaxed ordering gave no happens-before between successive winners.** Thread A wins interval N and invokes the callback; thread B wins interval N+1 and invokes it. Relaxed atomics establish no synchronizes-with edge, so B isn't guaranteed to see A's writes. Real `MessageHandler::Callback`s are stateful — appending to a pipeline log, updating a progress widget — so the callback's captured state races with itself even though the two sends never overlap in wall-clock time.

## Fixed #3 in fe6a9afa

That third one was a genuine defect rather than a documentation gap, so it's now fixed. The winning exchange goes to `acq_rel`, and the ticker thread opens the gate with an `exchange(true, acq_rel)` instead of a plain store. The RMW matters: a plain store from the ticker thread would break the release sequence headed by the previous winner's exchange, so the chain A → ticker → B wouldn't form. As a read-modify-write it stays in that sequence, and winner N is properly ordered before winner N+1.

Hot path is unchanged — the pre-filter load stays relaxed, since a stale read there only costs a dropped message, and the exchange only executes on the rare iteration that passes.

## The seam is still required

1 and 2 are unfixed by design, so `sendThreadSafeProgressMessage()` stays mandatory for anything calling `incrementCount`/`incrementPercent`. Worth being explicit that **that mutex is doing three jobs, not one**: guarding the counter, guarding label/denominator, and — until this commit — supplying the ordering. Anyone who reasons "the atomic already handles it, drop the lock" reintroduces all three. I've written that into the class doc comment so the next reader doesn't have to re-derive it.

`AlignSections::sendThreadSafeProgressMessage()` and `ConvertOrientations::sendThreadSafeProgressMessage()` are the reference implementations.

JDuffeyBQ (approved):
So long as `ThrottledMessageHandler` is called by a single thread or under a mutex lock, everything should be good.

Issue #1707, opened 2026-08-12
SurfaceNets: MMCellMap::nodeType re-derives per-vertex data that init() already computed


imikejackson:
## Summary

`MMCellMap::nodeType()` re-derives data that `MMCellMap`'s own initialisation already computed and threw away. It costs 8 scattered, bounds-checked `label()` reads plus a small distinct-count loop **per output vertex**, replacing what used to be a single cached bit-field read.

This is not a correctness problem and may well be lost in the noise — filing it with the measurement so someone can decide with a profile rather than a guess.

## Background

SurfaceNets used to write `MMCellFlag::numJunctions()` directly as its Node Types — a count of junction face-crossings in the range 0-6. That collided with `NodeType::Unused == 0` and did not match the convention QuickSurfaceMesh and M3CSurfaceMeshing use, so it was changed to compute the shared convention (`min(distinct corner labels, 4)`, `+10` when any corner is exterior padding) from the cell's 8 corner labels.

`numJunctions()` was a bit-field read off `MMCellFlag::m_BitFlag`. The replacement gathers the 8 corner labels afresh.

## The cost

Per **output vertex** (call it `V`, the surface-vertex count — not the voxel count `N`):

- 8 calls to `MMCellMap::label()`, each doing an index computation, a range check, and an array read;
- an O(64) worst-case distinct-count loop over those 8 values.

The access pattern is unfriendly: the 4 "near" corners are stride-1 and stride-`dimX` apart, but the 4 "far" corners sit a full XY-plane away — roughly 1 MB at 512³ — so most of the 8 reads are likely L2 misses.

**The same 8-corner gather already happens once per cell** during initialisation, to build `pCell->flag`, and is then discarded. So this is a genuine re-derivation of data the code already had in hand.

Mitigating context, for honesty: `V` is much smaller than `N`, and this is the same order of work as the `getEdgeQuad()` calls already performed for that vertex. It may not be measurable.

## Suggested fix

Compute the node type during initialisation, where the 8 labels are already loaded, and store it.

`MMCellFlag::m_BitFlag` is a `uint32` with `numJunctions` occupying bits 29-31, so there may be spare bits to hold the node type (it needs 4 bits to represent 2/3/4/12/13/14) at **zero** memory cost. That would be the ideal shape.

**Audit the spare bits first.** If they are not available, the fallback — a byte on each `Cell` — costs one byte per voxel, which is ~134 MB at 512³ and clearly not worth it for this.

[...]

Issue #1706, opened 2026-08-12
M3CSurfaceMeshing emits orphan vertices that no triangle references


imikejackson:
## Summary

`M3CSurfaceMeshing` emits vertices that no triangle in the output references. They are present in ordinary output, with no filter option enabled. On a 12³ test volume there are about 144 of them; the count presumably scales with the volume's boundary area.

## How it surfaced

While adding the "Bounding Box Skin" option to the surface meshers, the option's contract required that enabling it on data with no background voxels be a bit-for-bit no-op. A test of exactly that caught a discrepancy: faces matched (3528 == 3528, nothing pruned, correct) but vertex counts did not (1791 vs 1647).

The cause was that the option's orphan cleanup was clearing `nodeType` for **every** node unreferenced by a surviving triangle — including nodes that no triangle ever referenced. That has since been narrowed to clear only nodes orphaned *by* the prune, which fixed the no-op violation and left these pre-existing orphans in place, where they belong until this issue is addressed.

So this issue is the remainder: M3C's candidate-node generation produces `nodeType` entries near the volume boundary that survive compaction and become output vertices, yet are referenced by nothing.

## Why it matters

- Vertex counts are inflated relative to the actual mesh, which is misleading to anyone reasoning about mesh size or comparing meshers.
- Downstream consumers that iterate vertices rather than faces do wasted work, and any per-vertex array carries entries for geometry that is not part of the surface.
- It makes "does this mesh have orphan vertices?" a useless health check for M3C output, because the answer is always yes.
- Most concretely: with the Bounding Box Skin option enabled and every face pruned (an all-background volume), the output is now 0 faces with ~144 orphan vertices. That is honest — it reflects this wart rather than hiding it — but it looks strange, and the filter's warning text has to say "zero faces" rather than "zero vertices" because of it.

## Where to look

`src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/M3CSurfaceMeshing.cpp`

The node compaction is the `nodeType > 0` prefix-scan in `finalizeMesh`: every candidate with a nonzero `nodeType` becomes an output vertex. The question is why the marching-cubes candidate sweep sets a nonzero `nodeType` on nodes that no generated triangle ends up using — most likely boundary/ghost-shell interaction, given where they cluster.

## Suggested approach

Decide which is correct:

[...]

Issue #1705, opened 2026-08-12
Surface meshers: legitimate Feature Ids can alias exterior/ghost sentinels, silently corrupting output


imikejackson:
## Summary

All three surface meshers encode "not a real feature" using values drawn from the same `int32` space as legitimate Feature Ids. A Feature Ids array containing one of those values is silently misinterpreted — in some cases producing a corrupt mesh with no error, and in one case invoking signed-integer overflow (undefined behaviour).

Four distinct collisions, all pre-existing. Found during an adversarial review of surface-meshing work; a targeted mitigation ships alongside that work (see below), but the underlying design is unchanged.

## The collisions

**1. `MMSurfaceNet::Padding == std::numeric_limits<int32_t>::max()` — SurfaceNets**

`MMSurfaceNet.h` uses `INT32_MAX` as the label for the synthetic ghost shell around the volume. `MMCellMap::label()` returns the raw Feature Id with no range check, so a legitimate Feature Id of `2147483647` is indistinguishable from "outside the volume". That voxel's cell flags, vertex placement and Node Types are all computed as if it were exterior. No warning.

**2. `maxGrainId = maxGrainId + 1` overflows — M3CSurfaceMeshing**

In `initialize_micro`, `maxGrainId` is derived from the data and then incremented, to serve as the internal renumbering of Feature Id 0. With a Feature Id of `INT32_MAX` present this is signed overflow — undefined behaviour, in practice wrapping negative. The wrapped value is then written into every Feature-0 voxel as "the renumbered background", where it satisfies `nSpin < 0`, which is M3C's test for "ghost/exterior". Feature 0 then folds into Face Label `-1` instead of `0`.

**3. Negative Feature Ids collide with M3C's ghost sign convention**

M3C marks ghost cells by the *sign* of `nSpin`. A legitimate Feature Id of, say, `-5` is therefore treated as exterior wherever that convention is consulted. `M3CSurfaceMeshingFilter` places no range restriction on the Feature Ids parameter.

**4. Feature Id `-1` collides with QuickSurfaceMesh's own exterior marker**

`QuickSurfaceMesh` hard-codes `-1` as the exterior Face Label and inserts `-1` into its per-node owner sets. A legitimate Feature Id of `-1` is deduplicated against the exterior marker inside those `std::set`s, undercounting distinct owners and so producing a wrong Node Type; and a wall face backed by that feature gets Face Labels `{-1, -1}`.

[...]

Issue #1704, opened 2026-08-12
SIMPLNX_RESULT_REQUIRE_VALID evaluates its argument 4+ times, silently re-running filter calls


imikejackson:
## Summary

`SIMPLNX_RESULT_REQUIRE_VALID` and `SIMPLNX_RESULT_REQUIRE_INVALID` expand their argument **four or more times**. Passing them a function-call expression therefore re-executes that call, which for a filter means re-running it against an already-mutated `DataStructure`.

## The macros

`test/UnitTestCommon/include/simplnx/UnitTest/UnitTestCommon.hpp:43-62`

```cpp
#define SIMPLNX_RESULT_CATCH_PRINT(result)                        \
  for(const auto& warning : (result).warnings())      /* 1 */     \
  { WARN(...); }                                                  \
  if((result).invalid())                              /* 2 */     \
  {                                                               \
    for(const auto& error : (result).errors())        /* 3 */     \
    { ... }                                                       \
  }

#define SIMPLNX_RESULT_REQUIRE_VALID(result)                      \
  SIMPLNX_RESULT_CATCH_PRINT(result);                             \
  REQUIRE((result).valid());                          /* 4 */
```

So `result` is evaluated at least 4 times. `SIMPLNX_RESULT_REQUIRE_INVALID` has the same shape.

## Why it matters

The intended usage is a stored local, which is safe:

```cpp
auto executeResult = filter.execute(dataStructure, args);
SIMPLNX_RESULT_REQUIRE_VALID(executeResult.result);   // fine
```

But the macro invites the terser form, which is not:

```cpp
SIMPLNX_RESULT_REQUIRE_VALID(filter.execute(dataStructure, args).result);   // runs execute() 4x
```

That second form compiles, usually passes, and silently executes the filter four times against the same `DataStructure`. Failure modes include:

- filters that append to or resize an existing array producing 4x the data;
- filters that create a `DataObject` failing on the second run because the object already exists — reported as a confusing error from a test that "just calls execute once";
- any filter whose output depends on prior state producing results no one can reproduce by hand;
- masked non-determinism, since the assertion inspects the *last* run while later code inspects a `DataStructure` mutated four times.

It is also easy to introduce accidentally when factoring a test into a helper that returns a `Result<>`.

## Current status

[...]

PR #1708, opened 2026-08-13
ENH: Add Bounding Box Skin option to the surface meshing filters

### LINKED ISSUE #1684: ENH: Add option to SurfaceMesh filters to only mesh internal structure.
This would save a few steps for the user who only cares about internal structures such as porosity. Essentially, if the triangle has a FaceLabel that would be 0, then that triangle is not created.

imikejackson:
## Summary

Adds a **Bounding Box Skin** `ChoicesParameter` to Create Surface Mesh (QuickMesh), Create Surface Mesh (Surface Nets) and M3C Surface Meshing, with modes **Off** (default) and **Background-Backed Walls Only**.

Mode 1 suppresses a bounding box wall face only where the voxel behind it is background (Feature Id 0). Wall faces that cap a real Feature are still generated — that cut plane is the only possible closure for a Feature flush with the box — so such Features stay closed and watertight. Because the test is per *face* rather than per *vertex*, no triangles are lost along the rim where an internal boundary meets the wall, which is the erosion the existing Node Type Range criterion produces.

* Where an internal boundary meets the wall with the skin left on, three faces share one edge; dropping the background-backed face leaves two, so in the tested configurations the option yields manifold output where the default is non-manifold.
* When no wall face is backed by background (fully-indexed volume, or background enclosed as interior porosity) the option removes nothing and reports a warning saying so. On an all-background volume it removes everything and reports a warning, not an error.
* Adds a **Face Labels** criterion mode to Extract Internal Surfaces From Triangle Geometry so existing meshes can be corrected without re-meshing; its output matches meshing directly with the new option.
* Feature Ids that collide with internal sentinels are now rejected with a clear error instead of being silently misread: QuickMesh rejects negative values; Surface Nets and M3C reject negative values and `INT32_MAX`.

### Changed behavior for existing pipelines (not gated by the new option)

* **Surface Nets Face Labels** no longer conflate Feature Id 0 with the exterior: exterior-vs-Feature-0 faces are now `{-1, 0}` (was `{-1, -1}`), and Feature-0-vs-Feature-k faces are `{0, k}` (was `{-1, k}`).
* **Surface Nets Node Types** now follow the shared 2/3/4 + 12/13/14 convention (previously a 0-6 junction count). Anyone with a hand-tuned Node Type Range or stored Surface Nets output should re-check it.
* As a consequence, **Laplacian Smoothing and Hierarchical Smooth now actually work on Surface Nets meshes**: ~65% of vertices previously carried Node Type 0 (`Unused`), pinning their smoothing weight to 0.0. Neither filter's tests use a Surface Nets mesh, so neither could detect this.
* **Surface Nets vertex Z** previously used the Y spacing for its half-voxel offset; wrong on anisotropic volumes, now fixed (a QuickMesh wall-corner write was also fixed; its error was masked and output is unchanged).

Surface Nets exemplars are regenerated as `SurfaceNetsTest_v4.tar.gz` (uploaded to the Data_Archive release); only the four Node Type arrays differ from v3. Note the exemplar dataset contains no Feature Id 0 voxels, so exemplars cannot exercise the new option — the synthetic tests added here are its only coverage.

Related follow-up issues: #1704, #1705, #1706, #1707.

## Test Plan

- [x] `ctest -R "SimplnxCore::"` fully green, including new synthetic tests: no-op on fully-indexed volumes (bit-identical output per mesher), enclosed-background no-op, all-background degenerate case, flush-cylinder and box-corner watertightness, cross-mesher Face-Label conformance, QuickMesh↔Surface Nets Node Type corner-for-corner equivalence, anisotropic-spacing vertex Z, Feature Id sentinel rejection
- [x] `nx::core::FilterValidation` suite green (parameter key conventions)
- [x] Extract Internal Surfaces (Face Labels mode) output matches direct meshing with the option on
- [x] M3C sweep-path parity (entire-volume / windowed-serial / windowed-parallel) with the option on
- [x] Bundled example pipelines refreshed; load without `-5432` warnings
- [x] Docs build clean; documentation fact-checked against the implementation


src/Plugins/SimplnxCore/docs/M3CSurfaceMeshingFilter.md
```diff
@@ -64,6 +67,67 @@ heuristic, which does not guarantee globally consistent normals across the whole
 consistent across connected faces. This is recommended for meshes that will be used for normal- or
 curvature-dependent analysis.
 
+### Omitting the Bounding Box Skin
```
nyoungbq: This section feels like an information/brain dump, especially towards the latter half. I would try to make this section a little more concise and organize it by options exposed to the user. A small explainer paragraph at the top then have a subsection for X option and a small sentence/paragraph for each of the states available ("on" "off" etc). Be sure to update its copied section in the other surface meshing filters too.

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/ExtractInternalSurfacesFromTriangleGeometry.cpp
```diff
@@ -73,7 +73,34 @@ Result<> ExtractInternalSurfacesFromTriangleGeometry::operator()()
...
+  const Int8AbstractDataStore* nodeTypesPtr = nullptr;
+  const Int32AbstractDataStore* faceLabelsPtr = nullptr;
+  if(m_InputValues->CriterionMode == 0)
+  {
+    nodeTypesPtr = &m_DataStructure.getDataRefAs<Int8Array>(m_InputValues->NodeTypesPath).getDataStoreRef();
+  }
+  else
+  {
+    faceLabelsPtr = &m_DataStructure.getDataRefAs<Int32Array>(m_InputValues->FaceLabelsPath).getDataStoreRef();
+  }
```
nyoungbq, line 88: if they are stored as pointers theres no reason to get them as ref and deref them, just pull the datastore pointers via API directly

nyoungbq (commented):
code looks good small notes. make the changes and ill get an approval in before I leave today

nyoungbq (approved):

PR #1709, opened 2026-08-14
ENH: Changed test macros to require lvalues

### LINKED ISSUE #1704: SIMPLNX_RESULT_REQUIRE_VALID evaluates its argument 4+ times, silently re-running filter calls

JDuffeyBQ:
Fixes #1704

`SIMPLNX_RESULT_REQUIRE_VALID` and `SIMPLNX_RESULT_REQUIRE_INVALID` now assign the input `Result<T>` to avoid macro duplication when the argument is an expression (may copy or move depending on input type).

For tests, I think the following typical form should still be preferred since it allows for easy access to the `Result` in the debugger. This PR mainly serves as a guard against unintended behavior. 
```cpp
  auto executeResult = filter.execute(dataStructure, args);
  SIMPLNX_RESULT_REQUIRE_VALID(executeResult.result);
```

### DIFF
--- a/src/Plugins/ITKImageProcessing/test/ITKImageWriterTest.cpp
+++ b/src/Plugins/ITKImageProcessing/test/ITKImageWriterTest.cpp
@@ -385,8 +385,10 @@ TEST_CASE("ITKImageProcessing::ITKImageWriterFilter: RGBA Image Output", "[ITKIm
   args.insertOrAssign(ITKImageWriterFilter::k_TotalIndexDigits_Key, std::make_any<Int32Parameter::ValueType>(3));
   args.insertOrAssign(ITKImageWriterFilter::k_LeadingDigitCharacter_Key, std::make_any<StringParameter::ValueType>("0"));
 
-  SIMPLNX_RESULT_REQUIRE_VALID(filter.preflight(dataStructure, args).outputActions);
-  SIMPLNX_RESULT_REQUIRE_VALID(filter.execute(dataStructure, args).result);
+  auto preflightResult = filter.preflight(dataStructure, args).outputActions;
+  SIMPLNX_RESULT_REQUIRE_VALID(preflightResult);
+  auto executeResult = filter.execute(dataStructure, args).result;
+  SIMPLNX_RESULT_REQUIRE_VALID(executeResult);
 
   using ImageType = itk::Image<itk::RGBAPixel<uint8>, 2>;
   auto reader = itk::ImageFileReader<ImageType>::New();
--- a/src/Plugins/SimplnxCore/test/KeepRemoveRankedFeaturesTest.cpp
+++ b/src/Plugins/SimplnxCore/test/KeepRemoveRankedFeaturesTest.cpp
@@ -497,9 +497,11 @@ TEST_CASE("SimplnxCore::KeepRemoveRankedFeaturesFilter: Invariants", "[SimplnxCo
     // Keeping 2 gives {1, 4}; keeping 3 gives {1, 4, 3}. Compare cell-wise: every cell surviving
     // under k=2 must also survive under k=3.
     DataStructure smallDs = BuildFiveFeatureData();
-    SIMPLNX_RESULT_REQUIRE_VALID(filter.execute(smallDs, MakeArgs(0ULL, 0ULL, 2ULL)).result);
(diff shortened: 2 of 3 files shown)


JDuffeyBQ:
After thinking about this change, one issue with this method is that it makes the error message when a `REQUIRE` statement fails is less helpful. There will still be the line number but for quick diagnosing of errors the message helped.

Because of that it might be better to just add a compile time check that only allows lvalues to prevent the macro expansion from running the expression multiple times. I've tested a version that does that locally now and there are currently only 3 locations that fail. All of which are quick fixes.

imikejackson (approved):

PR #1710, opened 2026-08-18
ENH: Updated clang-format in ci to 18

JDuffeyBQ:
Updated clang-format to `18.1.3` on the GitHub actions runner using Ubuntu 24.04. Local clang-format should also be updated to be *exactly* `18.1.3`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -17,6 +17,13 @@ AllowShortFunctionsOnASingleLine: false
 # AlwaysBreakAfterReturnType: None
 # AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: true
+# This is necessary for clang-format 19. See https://github.com/llvm/llvm-project/issues/94184
+AttributeMacros:
+  - SIMPLNX_EXPORT
+  - SIMPLNXCORE_EXPORT
+  - ORIENTATIONANALYSIS_EXPORT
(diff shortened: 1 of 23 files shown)


imikejackson:
@JDuffeyBQ Why do we need exactly that version of Clang? That is not available for MacOS. I can get 18.1.8.

JDuffeyBQ:
The issue is less about the specific version and more that the CI and local versions must match. And 18.1.3 is what is available on the runner. The reason I mentioned that is because I had tested locally with 18.1.8 and it formats some files differently. So either we have to get 18.1.3 on all local machines or get 18.1.8 on the CI to avoid formatting mismatch problems.

imikejackson:
@JDuffeyBQ What about version 19? Looks like I can get 19.1.1, 19.1.7, 19.1.6, 19.1.5, 19.1.4

JDuffeyBQ:
@imikejackson 19 and 20 are available on 24.04 but they aren't installed by default on the runner. Looking at my local ubuntu it looks like they are specifically `19.1.1` and `20.1.2`.

imikejackson:
@JDuffeyBQ 19.1.1 is available for MacOS ARM. No MacOS x64 but I'm not really worried about that platform.

imikejackson:
Try and get clang-format 19.1.1 working.

imikejackson (approved):

PR #1712, opened 2026-08-18
COMP: Fix clang implicit conversion warnings in parameter conversion

imikejackson:
## Summary

Newer clang versions emit `-Wimplicit-const-int-float-conversion` in `NumberParameter.cpp` / `VectorParameter.cpp` and `-Wconstant-conversion` in `WriteBinaryDataTest.cpp`. The 64 bit integer cases were real defects, not just diagnostics.

`std::numeric_limits<int64>::max()` (2^63-1) is not representable as a `float64`. Converting it rounds **up** to 2^63, so `if(tmp > std::numeric_limits<T>::max())` was really `if(tmp > 2^63)` — a JSON value of exactly 2^63 passed the range check and then hit undefined behavior in `static_cast<int64>`. Same for `uint64` at 2^64.

* Add `Common/NumericBounds.hpp` with `ExceedsMaxOf<T>()` and `ExceedsLowestOf<T>()`. When the destination is integral and the value is floating point they compare against 2^`digits` — exactly representable, and one past the largest value the type can hold — with an inclusive test. Otherwise they fall back to the plain `numeric_limits` comparison, so behavior for the non-64-bit and floating point destination types is unchanged.
* Use the helpers in `IntFilterParameterConverter<T>::convert()` and `ConversionChecks<ExpectedT, TempT>()`.
* `WriteBinaryDataTest::setMemb()` used a runtime `if(std::is_same<T, ...>::value)` chain, so every branch was compiled for every `T` and the `uint64` fill values were narrowed to `int16` in the `int16` instantiation. Converted to `if constexpr` so the non-matching branches are discarded.

## Test Plan

- [x] New `TEST_CASE "nx::core::SIMPL Json Conversion 64 Bit Integer Bounds"` in `test/SimplJsonConversionTest.cpp` — `static_assert`s on the helpers plus runtime checks that 2^63 / 2^64 are rejected and the largest representable doubles below them still convert, through both `IntFilterParameterConverter` and `Vec3FilterParameterConverter`. Confirmed it fails against the unpatched sources and passes with the fix.
- [x] `cmake --build . --target all` clean, no compiler warnings
- [x] `ctest -R "WriteBinaryData|Test SIMPL Json Conversion|64 Bit Integer Bounds"` — 4/4 pass

### DIFF
--- a/src/simplnx/Common/NumericBounds.hpp
+++ b/src/simplnx/Common/NumericBounds.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <concepts>
 #include <limits>
 #include <type_traits>
 
@@ -16,10 +17,9 @@ namespace detail
  * @param exponent The power of two to compute. Must be non-negative and within the exponent range of FloatT.
  * @return 2^exponent
  */
-template <class FloatT>
+template <std::floating_point FloatT>
 constexpr FloatT PowerOfTwo(int exponent)
 {
-  static_assert(std::is_floating_point_v<FloatT>, "PowerOfTwo requires a floating point type");
   FloatT result = 1;
   for(int i = 0; i < exponent; i++)
   {
--- a/src/simplnx/Common/NumericBounds.hpp
+++ b/src/simplnx/Common/NumericBounds.hpp
@@ -1,0 +1,92 @@
+#pragma once
+
+#include <limits>
+#include <type_traits>
+
+namespace nx::core
+{
+namespace detail
+{
+/**
+ * @brief Computes 2^exponent as a floating point value at compile time.
+ *
+ * Every power of two within the exponent range of FloatT is exactly representable, so repeated
+ * doubling introduces no rounding error.
+ * @tparam FloatT The floating point type to compute the value in
+ * @param exponent The power of two to compute. Must be non-negative and within the exponent range of FloatT.
+ * @return 2^exponent
+ */
+template <class FloatT>
+constexpr FloatT PowerOfTwo(int exponent)
+{
+  static_assert(std::is_floating_point_v<FloatT>, "PowerOfTwo requires a floating point type");
+  FloatT result = 1;
+  for(int i = 0; i < exponent; i++)
+  {
+    result *= 2;
+  }
+  return result;
+}
+} // namespace detail
+
+/**
+ * @brief Returns true if `value` is larger than the largest value that T can represent.
+ *
+ * Comparing a floating point value directly against std::numeric_limits<T>::max() is incorrect for
+ * the 64 bit integer types. Their maximum is not exactly representable as a float64, so the
+ * implicit conversion rounds the bound *up* to 2^63 / 2^64. That both widens the accepted range by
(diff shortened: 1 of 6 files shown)


src/simplnx/Common/NumericBounds.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * Every power of two within the exponent range of FloatT is exactly representable, so repeated
+ * doubling introduces no rounding error.
+ * @tparam FloatT The floating point type to compute the value in
+ * @param exponent The power of two to compute. Must be non-negative and within the exponent range of FloatT.
+ * @return 2^exponent
+ */
+template <class FloatT>
+constexpr FloatT PowerOfTwo(int exponent)
+{
+  static_assert(std::is_floating_point_v<FloatT>, "PowerOfTwo requires a floating point type");
```
JDuffeyBQ:
```suggestion
template <std::floating_point FloatT>
constexpr FloatT PowerOfTwo(int exponent)
{
```
We can use a concept to replace the static_assert to more directly express intent. Will also require the `<concepts>` header.

test/SimplJsonConversionTest.cpp
```diff
@@ -44,3 +50,70 @@ TEST_CASE("nx::core::Test SIMPL Json Conversion", "[simplnx][Filter]")
...
+// std::numeric_limits<int64>::max() rounds up to exactly 2^63 when converted to a float64, a naive
+// `value > max()` range check accepts these and then invokes undefined behavior in the cast.
+constexpr float64 k_TwoToThe63 = 9223372036854775808.0;
+constexpr float64 k_TwoToThe64 = 18446744073709551616.0;
+} // namespace
+
+TEST_CASE("nx::core::SIMPL Json Conversion 64 Bit Integer Bounds", "[simplnx][Parameters]")
+{
+  // The bound helpers themselves
+  static_assert(ExceedsMaxOf<int64>(k_TwoToThe63));
```
JDuffeyBQ: Not strictly required but Catch2 does have a `STATIC_REQUIRE` macro.

https://github.com/catchorg/Catch2/blob/devel/docs/other-macros.md

It behaves the same as `static_assert` but adds a nicer error message. It also reports that it passed when the test runs.

JDuffeyBQ (changes requested):

src/simplnx/Common/NumericBounds.hpp
imikejackson: Applied in aeb5e5c8 — `template <std::floating_point FloatT>` with `<concepts>` added. Matches the existing spelling in `src/simplnx/Utilities/Math/GeometryMath.hpp`.

test/SimplJsonConversionTest.cpp
imikejackson: Switched to `STATIC_REQUIRE` / `STATIC_REQUIRE_FALSE` in aeb5e5c8. Since the expansion appends a runtime `SUCCEED`, and code ahead of a `SECTION` re-runs once per section, I moved the nine checks into their own `SECTION` so they are reported once rather than four times. Test case now reports 19 assertions (10 runtime + 9 static).

PR #1728, opened 2026-08-24
VV: Read H5OINA Data Fully V&V'ed

imikejackson:
## Summary
Full oracle-first V&V of `ReadH5OinaDataFilter` per the MTR V&V policy (`docs/vv_templates/`). This is the program's first new-filter V&V — no DREAM3D 6.5.171 equivalent exists — so the legacy binary A/B is replaced by a **Class 2 independent readback**: an h5py script re-reads every fixture and the real AZtec file and independently derives the expected arrays. Report, deviations, and provenance at `src/Plugins/OrientationAnalysis/vv/`.

⚠️ **MERGE DEPENDENCY:** the final commit pins `vcpkg.json` to EbsdLib `>= 3.1.2`. Five of the corrections live in EbsdLib's `H5OinaReader` on `topic/3_1_1_staging` and reach users only through that release (which also gates #1723). Until 3.1.2 is published, this branch's HEAD does not configure; all test evidence is recorded at HEAD−1 against the local staging EbsdLib.

**Eleven defects corrected (6 filter, 5 EbsdLib), each pinned by a regression test:**
* ⚠️ **Release-note item — hexagonal Euler corruption (D1):** the "Convert Hexagonal X-Axis" option (on by default) added the literal `30.0` — thirty *radians* — to radian-valued φ2, producing garbage orientations for every hexagonal point in every release that shipped this filter. The alignment now adds the intended 30° (π/6). Invisible until now because the only test file was cubic.
* **Multi-scan stacking (D2, D3):** the Euler slab offset was off by 3× (later scans overwrote earlier scans' tails and left their own slabs zeroed), and hex alignment was applied repeatedly to scan 0 and never to later scans.
* **Multi-scan crash (D8):** ensemble arrays were sized from the first scan only; a well-formed file whose later scans declare different phase groups crashed the process (reproduced; both probe files are now committed guard fixtures). Preflight now validates every selected scan (`-9587`, `-9589`).
* **EbsdLib `H5OinaReader` (D5, D6, D7, D10, D11):** the lattice-constant γ slot echoed β; lattice angles were imported in radians where every other importer reports degrees — ⚠️ **breaking change to a published output**, with a release note and migration section in the deviations file; a phase group missing `Lattice Angles` crashed the process; error-path hygiene.
* **Stacking Order implemented (D9):** previously accepted, serialized, and silently ignored.
* **Pattern import descoped:** it could never succeed (EbsdLib stub); the toggle now fails with an honest "not yet supported" error, documented as a known limitation.

**Tests:** 2 → 17 test cases (9,374 assertions): the circular NX-generated exemplar comparison was replaced by readback-derived assertions against the genuine AZtec file; hand-built h5py fixtures cover hex alignment (pinned to numerically distinguish the correct π/6 from the former literal), multi-scan stacking, guard cases, and the real-file import. Mutation-verified. Also fixes the invalid-input TEST_CASE that referenced a file present in no archive (it passed vacuously).

**Docs:** the format-version claim, the CTF copy-paste alignment paragraph, the created-objects table, and a masking recipe that selected nothing on real data are corrected.

**Follow-ups (in the report):** `ReadH5OimData`/`ReadH5EspritData` share several of these defect shapes via `IEbsdOemReader` and are enumerated for the engineer track; closing the preflight-vs-execute gap at the shared write needs a cross-filter change.

## Test Plan
- [x] `ctest -R "OrientationAnalysis::ReadH5Oina"` — 17/17 (HEAD−1, local EbsdLib 3.1.1 staging)
- [x] Full `OrientationAnalysis::` — 308 tests, 1 pre-existing failure (`ComputeSchmidsFilter`, the known CubicOps drift corrected by #1723); `SimplnxCore::` — 985/985
- [x] EbsdLib suite — 408/408 with the staging corrections
- [x] Independent h5py readback agrees with every imported array (evidence archived per V&V policy)


imikejackson:
Internal V&V review — actionable findings

- [x] `ReadH5OinaDataFilter-D1`: correct the hexagonal alignment from 30 radians to 30 degrees expressed in radians.
- [x] `ReadH5OinaDataFilter-D2`: use the three-component Euler element offset for every scan slab.
- [x] `ReadH5OinaDataFilter-D3`: apply hexagonal alignment to the current scan slab.
- [x] `ReadH5OinaDataFilter-D4`: reject unsupported H5OINA pattern import in preflight and provide an alternative.
- [x] `ReadH5OinaDataFilter-D5`: copy the gamma lattice angle into the gamma slot in EbsdLib.
- [x] `ReadH5OinaDataFilter-D6`: convert H5OINA lattice angles from radians to degrees in EbsdLib.
- [x] `ReadH5OinaDataFilter-D7`: validate required phase datasets before reading their values in EbsdLib.
- [x] `ReadH5OinaDataFilter-D8`: require safe phase indices, equal phase counts, and identical phase definitions for scans that form one 3D microstructure.
- [x] `ReadH5OinaDataFilter-D9`: honor the selected stacking order.
- [x] `ReadH5OinaDataFilter-D10`: preserve complete H5OINA reader error messages in EbsdLib.
- [x] `ReadH5OinaDataFilter-D11`: validate both cell counts and return the stored EbsdLib error code.
- [x] `ReadH5OinaDataFilter-D12`: reject noncanonical phase-group names without an exception or HDF5 handle leak.
- [x] `ReadH5OinaDataFilter-D13`: reject non-positive or non-finite X, Y, and Z spacing.
- [x] Replace the generated DREAM3D circular oracle with inline Class 1 analytical data, Class 4 invariants, and an independent h5py readback.
- [x] Rewrite the V&V report and deviation records around the differences from DREAM3D-NX 7.0.0 through 7.4.1; DREAM.3D 6.5.171 has no H5OINA importer.
- [x] Complete the CPU, memory, naming, error-message, documentation, and adversarial review. Runtime is linear in the imported point count. Working storage is limited to the reader data for one scan plus the required output arrays; no extra full-volume working copy is allocated.
- [x] Verify the focused suite through ctest: 20 of 20 tests pass.
- [x] Verify EbsdLib PR 56: macOS, Linux, Windows, and clang-format checks pass.
- [ ] After EbsdLib 3.1.2 and its registry update merge, rerun simplnx CI with the final `vcpkg.json` requirement. Current simplnx CI cannot resolve the unpublished dependency version.

Implemented in simplnx commit `80b5fff07` and EbsdLib commit `635b129`.

PR #1738, opened 2026-09-10
FILT: Add WriteAbaqusCrystalPlasticityFilter ported from SimulationIO

imikejackson:
## Summary
Ports the legacy DREAM.3D 6.6 SimulationIO `CreateAbaqusFile` filter into SimplnxCore as `WriteAbaqusCrystalPlasticityFilter`. The filter writes five Abaqus input files (master, nodes, C3D8R elements, per-grain element sets, and solid sections) for an Image Geometry, with one `*User Material` card per grain that carries the grain id, phase id, three Euler angles in degrees, and user-supplied material constants.

* Registers the legacy UUID and `CreateAbaqusFile` name so SIMPL 6.4 and 6.5 pipelines convert
* Fixes the legacy `_set`/`_mat` material-name mismatch so the deck references a defined material
* Buckets element ids per grain in one contiguous pass instead of rescanning all cells per grain
* Keeps the legacy last-cell-wins phase and orientation assignment and documents it
* Warns on feature ids with no cells and errors when there are no positive feature ids
* Reports throttled progress per Z slice and per grain with cancel checks at the same boundaries

## Test Plan
- [x] `WriteAbaqusCrystalPlasticityFilter` unit tests: byte-exact five-file synthetic case, wrapping rules (16 per line element sets, 8 per line constants), gap in feature ids, no positive ids, feature id 0 skipped, tuple-count mismatch, missing output directory, negative counts, SIMPL 6.4 and 6.5 conversion
- [x] `WriteAbaqusHexahedronFilter` regression tests
- [x] SIMPL Map Inventory coverage test
- [x] clang-format 19.1.1 dry run clean


src/Plugins/SimplnxCore/vv/deviations/WriteAbaqusHexahedronFilter.md
```diff
@@ -0,0 +1,59 @@
+# Deviations from DREAM3D 6.5.171: WriteAbaqusHexahedronFilter
```
JDuffeyBQ, line 1: Should there also be the rest of the V&V here?

src/Plugins/SimplnxCore/src/SimplnxCore/Filters/Algorithms/WriteAbaqusInputDeck.cpp
```diff
@@ -0,0 +1,585 @@
...
+  std::vector<int32> Phases;
+  std::vector<std::array<float32, 3>> Orientations;
+  std::vector<usize> CellIndices;
+  std::vector<usize> Offsets;
+};
+
+constexpr usize ToIndex(FileIndex fileIndex)
+{
+  return static_cast<usize>(fileIndex);
+}
```
JDuffeyBQ, line 52: Could be replaced by `to_underlying`

JDuffeyBQ (changes requested):
Also I think this has a stray file commited.

